PR #7, opened 2017-03-06
Bump LTS to latest and add upper bounds for dependencies

adelbertc:


snowleopard:
Thanks @adelbertc!

At some point I might reconsider the bounds on package dependencies, but this should be fine for now.

ndmitchell:
I would caution against too many upper bounds - you end up doing a lot of busy work with package version tweaks. However, opinions vary...

snowleopard:
@ndmitchell I haven't yet experienced this myself, but I did manage to find your Shake commit removing all upper bounds, so it seems that you got fed up with them at some point :-)

adelbertc:
Out of curiosity what is the rule of thumb for controlling these bounds? Or is it a case, as you said, that opinions vary?

ndmitchell:
http://neilmitchell.blogspot.co.uk/2014/11/upper-bounds-or-not.html is my explanation.

snowleopard:
Thanks @ndmitchell!

In the case of Alga, we currently only depend on core GHC libraries `array`, `base` and `containers`, which do not change very often and when they do, it's not a surprise. So, omitting upper bounds will most probably minimise the number of times we need to edit the `.cabal` file.

adelbertc:
Gotcha, makes sense - thank you for the explanation!

I'm just getting back into Haskell after a... 4 year hiatus so please excuse my newbie questions :-)

PR #18, opened 2017-07-12
Add klVertex, dfsForestFrom

feuerbach:
Hi,

These are some functions that I needed and didn't find.

This PR is not ready to be merged — there's no corresponding addition to IntAdjacencyMap, and the documentation is probably not up to your standards, but at least it can start a conversation.

### DIFF
--- a/src/Algebra/Graph/AdjacencyMap.hs
+++ b/src/Algebra/Graph/AdjacencyMap.hs
@@ -38,16 +38,17 @@ module Algebra.Graph.AdjacencyMap (
     removeVertex, removeEdge, replaceVertex, mergeVertices, transpose, gmap, induce,
 
     -- * Algorithms
-    dfsForest, topSort, isTopSort, scc,
+    dfsForest, dfsForestFrom, topSort, isTopSort, scc,
 
     -- * Interoperability with King-Launchbury graphs
-    GraphKL, getGraph, getVertex, graphKL, fromGraphKL
+    GraphKL, getGraph, getVertex, klVertex, graphKL, fromGraphKL
   ) where
 
 import Data.Array
 import Data.Foldable (toList)
 import Data.Set (Set)
 import Data.Tree
+import Data.Maybe
 
 import Algebra.Graph.AdjacencyMap.Internal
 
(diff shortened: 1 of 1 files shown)


snowleopard:
Awesome, glad you're using the library! I agree that a lot of useful things are currently missing.

I'm travelling until the weekend, so I'll only be able to give a proper review on Saturday, but please keep commits going. I'd like to see some examples to make sure I understand how these functions are going to be used. Thanks!

feuerbach:
Oh yeah, I'm a big fan of the library. We had 30 lines of code doing topological sort with fgl, I replaced them with 5 lines using algebraic graphs. So thank you for it!

`dfsForestFrom` finds all vertices below the given set of vertices. You could use it to find the set of transitive dependencies or reverse dependencies of a package, for example.

And `klVertex` is basically the inverse of `getVertex`. One example of its usage is `dfsForestFrom` itself, but I'm sure there are other cases too.

snowleopard:
> Oh yeah, I'm a big fan of the library. We had 30 lines of code doing topological sort with fgl, I replaced them with 5 lines using algebraic graphs. So thank you for it!

Cool! I'm very curious to see the before/after comparison, assuming the code is in public domain.


What do you think about hiding `GraphKL`? I've just realised that the `getVertex` function is partial, and I'd like to avoid having partial functions in the API where possible.

The purpose of `GraphKL` is to provide a way to reuse the algorithms from `Data.Graph`, but there are not too many of them left (`dfs`, which this PR adds, `bcc`, plus a few reachability tests) and I think we can reasonably aim to provide clean interfaces to all of them from `(Int)AdjacencyMap`.

feuerbach:
> Cool! I'm very curious to see the before/after comparison, assuming the code is in public domain.

Unfortunately, it's not, it's part of @nstack's proprietary platform. But basically, many lines were spent to establish a mapping between Haskell values and integers and convert back and forth, which alga basically does for you.

I might write a blog post with a comparison, but honestly don't know if I can find the time.

> What do you think about hiding GraphKL?

I think it makes sense, but I am too early a user to have an informed opinion.

snowleopard:
> it's part of @nstack's proprietary platform

Cool :) I know of several closed-source projects using Alga, but no open-source ones (apart from my own)!

> many lines were spent to establish a mapping between Haskell values and integers and convert back and forth, which alga basically does for you.

I see, thanks for clarifying.

> I think it makes sense, but I am too early a user to have an informed opinion.

Right, I think I'll remove `GraphKL` after we close this PR. See #19. 

How would you like to proceed? If you don't have time to write docs/tests for this PR, we can merge it as is and I'll do the rest as part of tackling #19.

feuerbach:
> If you don't have time to write docs/tests for this PR, we can merge it as is and I'll do the rest as part of tackling #19.

That would be great! Thank you.

snowleopard:
@feuerbach OK, done. I'll try to close #19 during this week.

locallycompactCT:
Hi, I was trying to use this MR for a [build system](https://gitlab.com/locallycompact/earthquake) I'm writing. It seems dfsForestFrom is only implemented for AdjacencyMap and not the standard Graph type? Is there a technical reason for this? What's the difference between these two things? Can I convert from one to the other?

locallycompact:
Hi, I was trying to use this MR for a [build system](https://gitlab.com/locallycompact/earthquake) I'm writing. It seems dfsForestFrom is only implemented for AdjacencyMap and not the standard Graph type? Is there a technical reason for this? What's the difference between these two things? Can I convert from one to the other?

snowleopard:
@locallycompact Hi there!

> What's the difference between these two things?

The difference is in the internal graph representation that `Graph` and `AdjacencyMap` use. `Graph` is an algebraic data type that represents a graph by a tree-like expression. `AdjacencyMap` represents a graph as a map from a vertex to a set of its adjacent vertices, which is more convenient for performing a depth-first search using conventional algorithms available in the `containers` library.

> Can I convert from one to the other?

Yes. `Graph` is an instance of the type class `ToGraph` ([docs](http://hackage.haskell.org/package/algebraic-graphs-0.0.4/docs/Algebra-Graph-Class.html#t:ToGraph)), so if you have a `g :: Graph a` and need to convert it to `h :: AdjacencyMap a`, you can do this by `h = toGraph g`.

Hope this helps! Don't hesitate to create an issue describing your problem if you need more help.

PR #22, opened 2017-10-15
Catch a few typos

ggreif:
Don't merge just yet, if you let it soak, more commits will appear :-)

### DIFF
--- a/src/Algebra/Graph/AdjacencyMap.hs
+++ b/src/Algebra/Graph/AdjacencyMap.hs
@@ -348,7 +348,7 @@ vertexSet = Map.keysSet . adjacencyMap
 edgeSet :: Ord a => AdjacencyMap a -> Set (a, a)
 edgeSet = Map.foldrWithKey (\v es -> Set.union (Set.mapMonotonic (v,) es)) Set.empty . adjacencyMap
 
--- | The /postSet/ of a vertex is the set of its /direct successors/.
+-- | The /postset/ (here 'postSet') of a vertex is the set of its /direct successors/.
 --
 -- @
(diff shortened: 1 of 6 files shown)


snowleopard:
@ggreif Thank you! :-) Sure, I'll wait -- ping me when the PR is ready.

P.S.: Great to meet you in person at Haskell eXchange. Sorry I had to leave early and couldn't attend your talk, hopefully it was recorded!

src/Algebra/Graph/AdjacencyMap.hs
```diff
@@ -348,7 +348,7 @@ vertexSet = Map.keysSet . adjacencyMap
 edgeSet :: Ord a => AdjacencyMap a -> Set (a, a)
 edgeSet = Map.foldrWithKey (\v es -> Set.union (Set.mapMonotonic (v,) es)) Set.empty . adjacencyMap
 
--- | The /postset/ of a vertex is the set of its /direct successors/.
+-- | The /postSet/ of a vertex is the set of its /direct successors/.
```
snowleopard: I was actually using 'preset' and 'postset' as terms (not very common ones, but accepted in the Petri Net community), hence no capitalisation.

Inconsistent capitalisation does seem a bit confusing, so if we change to using 'postSet' instead of 'postset' then I suggest that we also drop the slashes around, replacing them with either quotes `'postSet'` or @ sign `@postSet@`.

ggreif: Yep. Made some more commits so that this should be clear now. Feel free to squash-and-merge.

snowleopard:
@ggreif Thank you! Merged.

PR #26, opened 2017-11-11
Add NFData Graph instance

phadej:


algebraic-graphs.cabal
```diff
@@ -69,7 +69,8 @@ library
                         Algebra.Graph.Relation.Transitive
     build-depends:      array      >= 0.5 && < 0.8,
                         base       >= 4.9 && < 5,
-                        containers >= 0.5 && < 0.8
+                        containers >= 0.5 && < 0.8,
+                        deepseq    >= 1.4.2.0 && <1.5
```
phadej: that's the version distributed with GHC-8.0

src/Algebra/Graph.hs
```diff
@@ -47,6 +47,7 @@ module Algebra.Graph (
   ) where
 
 import Control.Applicative (Alternative, (<|>))
+import Control.Deepseq (NFData (..))
```
snowleopard: Looks like this should be `DeepSeq` instead -- see AppVeyor build log:

https://travis-ci.org/snowleopard/alga/jobs/300622428

algebraic-graphs.cabal
snowleopard: I'll see if Alga works with earlier versions of GHC and might move this lower bound to allow earlier GHC versions.

snowleopard:
@phadej Thanks you, merged!

Will do the release soon.

Issue #27, opened 2017-11-11
Support older GHC.


phadej:
I can make a PR for this is well.

It would be great to support GHC >= 7.6 (7.4 would need older containers-0.4)
Then I could ask people to add dependency on alga.

with https://github.com/hvr/multi-ghc-travis it would be easy to test them all on Travis (even OSX, if you really insist, though for `alga` it's IMHO waste of scarse OSX travis resources)

snowleopard:
@phadej Thank you! I've never worked on Haskell projects with long-term GHC compatibility requirements, so let me ask a couple of naive questions:
* Your PR #28 gives a working solution for GHC 7.8, which means supporting the current version of GHC 8.2.1, plus three previous versions. Is this a common compatibility target? Would supporting even older versions be useful, e.g. are there a lot of users of GHC 7.6?
* What is the long term plan for Alga with new versions of GHC coming out? Say, GHC 8.4 is out -- do we then discontinue support for GHC 7.8, or will we keep accumulating CPP and other compatibility hacks? 

> Then I could ask people to add dependency on alga.

Are you referring to a particular project you work on? I'm curious! (And a bit worried, since Alga's API is still not very polished and stable.)

> even OSX, if you really insist, though for alga it's IMHO waste of scarse OSX travis resources

I think it would be nice to run Travis OSX with the latest GHC version. Just in case.

phadej:
There are three strategies:
- Support as many GHC as is convenient. This is subjective measure, but it means, that if you would like to use say `TypeInType` or `TypeApplications`, then you'll require at least GHC-8.0.
    - This is common with libraries which heavily rely on some GHC features.
    - type-leval `Nat` and `Symbol` didn't exist before GHC-7.8
    - When GHC-8.4 is released, you won't do anything, except if "it's time to clean-up stuff a little"
- Support three latest GHC (three because of https://prime.haskell.org/wiki/Libraries/3-Release-Policy)
    - that means that supporting GHC-7.8 isn't required anymore.
    - When GHC-8.4 is released, you can drop GHC-7.10 support, etc.
- Support as many as you can. This is hard-core, and I don't propose it for `alga`. That's semi required e.g. for testing / benchmarking libs etc.
    - This is a variation of the first strategy.
        - there "I want `TypeApplications`", I use them
        - here, "I want `TypeApplications`", I rethink, do I really need them.
    - Though there is little reason to support older compilers than GHC-7.4 atm.
[...]

PR #28, opened 2017-11-11
Ghc 7.8

phadej:
PoC for #27 

For GHC-7.6, one would need to have few CPP
- in `Relation` (`Set.map` has (Ord a, Ord b)  constraint in old containers)
- in tests, type of `box` makes GHC-7.6 fail to compile. (It tries to unify it for some reason)


algebraic-graphs.cabal
```diff
@@ -103,13 +113,17 @@ test-suite test-alga
                         Algebra.Graph.Test.IntAdjacencyMap,
                         Algebra.Graph.Test.Relation
     build-depends:      algebraic-graphs,
-                        base       >= 4.9,
-                        containers >= 0.5,
+                        base,
+                        base-compat,
+                        base-orphans >= 0.6 && <0.7,
```
snowleopard: I presume `base-compat` and `base-orphans` are small, standard and stable solutions, so depending on them is not a big maintenance burden?

phadej: @RyanGlScott is one of maintainers of both libraries, they are well maintained.

Wouldn't say they are *standard* (there a bunch of alternative preludes out there), but they have quite well known packages as revdeps:
- http://packdeps.haskellers.com/reverse/base-orphans
- http://packdeps.haskellers.com/reverse/base-compat

snowleopard: I see, thanks.

algebraic-graphs.cabal
```diff
@@ -67,10 +71,14 @@ library
...
-                        containers >= 0.5 && < 0.8,
-                        deepseq    >= 1.4.2.0 && <1.5
+    build-depends:      array       >= 0.4     && < 0.6,
+                        base        >= 4.7     && < 5,
+                        base-compat >= 0.9.1   && < 0.10,
+                        containers  >= 0.5.5.1 && < 0.8,
+                        deepseq     >= 1.3.0.1 && < 1.5
+    if !impl(ghc >= 8.0)
+      build-depends:
+                        semigroups >=0.18.3 && <0.18.4
```
snowleopard: I don't see where the `semigroups` package is used. Is it really needed?

phadej: For `Data.Semigroup` with old `base`.

snowleopard: Aha, I see. I will merge this as is for now, but I will try to get rid of the dependency on `semigroups` via CPP. Essentially, I would like to provide the `Semigroup` instance only if it is required by the new `Semigroup => Monoid` hierarchy. For older versions of `base` I think it's OK to provide only `Monoid`.

snowleopard:
Merged, thank you!

algebraic-graphs.cabal
phadej: I'd advice against that. Currently `semigroups` overhead is
- three lines in cabal file, and 
- few additional transitive dependencies for GHC < 8

However if you try to avoid that depedency at all cost:
- The API will be different across different GHC, which might cause *wtf* moments for users, as e.g. Hackage (and Stackage) haddocks will show `Semigroup` instance (as they are generated with recent GHC)
- there will be CPP clutter (guards for imports and instance itself at least).

---

Note: this is the case where we see 3-release-policy in practice. If you were to support only three latest GHC, with GHC-8.4 you could finally add `Semigroup` instance across whole support window`, without CPP *or* addtional dependencies. And in GHC-8.4 you have to, as `Semigroup` is a superclass of `Monoid` in `base-4.11` (GHC-8.4). FWIW, `algebraic-graphs` (this branch to be precise) compiles cleanly with quite recent GHC HEAD too, so we don't need to worry about this.

snowleopard: I see, thanks! I agree that the API should be uniform across different GHC. The main obstacle in this case is that it is impossible to hide the `Semigroup` instance from the API.

src/Algebra/Graph/HigherKinded/Class.hs
```diff
@@ -500,7 +508,7 @@ deBruijn :: Graph g => Int -> [a] -> g [a]
 deBruijn 0   _        = edge [] []
 deBruijn len alphabet = skeleton >>= expand
   where
-    overlaps = mapM (const alphabet) [2..len]
+    overlaps = traverse (const alphabet) [2..len]
```
snowleopard: @phadej I'm confused about this change. This doesn't seem to be for the sake of compatibility -- I just tried to undo it and all CI instances passed. Could you clarify? Happy to do it, but I'd like to understand the reason.

snowleopard: Furthermore, there are few more instances of `mapM` in the code, so if we change this one, we should probably changes others too.

PR #30, opened 2017-11-12
Clean up

snowleopard:


src/Algebra/Graph/HigherKinded/Class.hs
```diff
@@ -508,7 +508,7 @@ deBruijn :: Graph g => Int -> [a] -> g [a]
 deBruijn 0   _        = edge [] []
 deBruijn len alphabet = skeleton >>= expand
   where
-    overlaps = traverse (const alphabet) [2..len]
+    overlaps = mapM (const alphabet) [2..len]
```
phadej: my bad. I could change it back to `mapM`. My personal preference is to use `traverse` all over the place though.

snowleopard: I like to use `traverse` too because it is more general. But on the other hand `mapM` may be easier to understand, exactly because it is less general. So, I'm undecided what to use in code which is supposed to be read by many people, including novices.

Issue #31, opened 2017-11-13
Non empty graphs


snowleopard:
@phadej makes a very good observation in #24: Alga currently does not provide any support for type-safe handling of non-empty graphs.

I also wanted `NonEmptyGraph` a few times. Many graph transformations become much simpler when graph expressions cannot contain the `Empty` leaf. (And conversely, many graph transformations become simpler when `Empty` leafs are allowed, so general graphs are also useful -- we need both.)

The algebraic data type for non-empty graphs is straightforward:
```haskell
data NonEmptyGraph a = Vertex a
                     | Overlay (NonEmptyGraph a) (NonEmptyGraph a)
                     | Connect (NonEmptyGraph a) (NonEmptyGraph a)
```

However, there are a couple of design questions that need to be answered:
* Where should non-empty graphs live? I suggest a separate module `Algebra.Graph.NonEmpty`. This will lead to quite a bit of code/docs/tests duplication, but I don't see a better way. Incidentally, this is what happens with the standard `Data.List.NonEmpty`, which has quite a bit of overlap with `Data.List`.
* Do we also create a type class for non-empty graphs (or plexes, or whatever else we are going to name the general concept of things with overlay and connect operations, see #24)? If yes, and if we stick to the current naming where we have `class Graph g`, we'll then add `class NonEmptyGraph g` that will be the superclass of `class Graph g`.
* What do we do with `Relation`, `AdjacencyMap`, etc.? Do we need their non-empty versions? I think the answer is No, but I'm not sure.

Thoughts?

snowleopard:
Perhaps, a slightly better approach would be to stick to the name `Graph`, but use module names to distinguish between general and non-empty graphs?

More specifically:
```haskell
module Algebra.Graph.NonEmpty (Graph (..)) where

data Graph a = Vertex a
                     | Overlay (Graph a) (Graph a)
                     | Connect (Graph a) (Graph a)
```

Now, if one would like to be explicit about the data type, it's possible to write `NonEmpty.Graph`. And if non-empty graphs are assumed throughout the codebase, then a simple `Graph` will work.

This is similar to `Data.Map.Strict` and `Data.Map.Lazy` that both definine data type `Map`.

Following this approach we can also have `Undirected.Graph`, `Reflexive.Graph`, `Undirected.NonEmpty.Graph` etc. without the need to invent longer and longer names for various graph data-types.

phadej:
[...]

Issue #15, opened 2017-03-28
Reduce the size complexity of removeEdge


snowleopard:
As suspected by @ndmitchell, the current implementation of `removeEdge` is not linear w.r.t. the size of the resulting expression. An example where the result explodes by the factor of 10 found by QuickCheck:
```haskell
g = clique [2,2,0,1,0,0,2,1,2,1,0,1,2,0,2,0,0,0,0,2,1,1,0,2,2,0,2,0,0,2,1,2,0
           ,2,0,1,1,2,0,0,0,1,2,1,0,1,1,2,2,0,2,2,2,1,0,0,1,2,0,1,0,0,2,0,2]

> size g
66

> size (removeEdge 0 1 g)
681
```
Although this example is contrived (it's just a fully connected graph on vertices {0,1,2} with a lot of redundancy), it still demonstrates the problem.

snowleopard:
After #37, we have the following improvement:
```haskell
g = clique [2,2,0,1,0,0,2,1,2,1,0,1,2,0,2,0,0,0,0,2,1,1,0,2,2,0,2,0,0,2,1,2,0
           ,2,0,1,1,2,0,0,0,1,2,1,0,1,1,2,2,0,2,2,2,1,0,0,1,2,0,1,0,0,2,0,2]

> size g
66

> size (removeEdge 0 1 g)
155
```

We also have a guarantee that the size of the resulting expression is at most 3x of the original. This is a very pessimistic worst case which only applies to strange cliques like in the above example. A couple of more realistic examples:
```haskell
> p = path [1..10]
> size p
19
> size (removeEdge 5 6 p)
25

> q = clique [1..10]
> size q
11
> size (removeEdge 5 6 q)
24
```

snowleopard:
After merging #49, the results got better, but the worst case still remains at 3x, and I don't know how to improve it -- the last example below demonstrates the problem. I believe this is a rather rare case.

```haskell
> g = clique [2,2,0,1,0,0,2,1,2,1,0,1,2,0,2,0,0,0,0,2,1,1,0,2,2,0,2,0,0,2,1,2,0
           ,2,0,1,1,2,0,0,0,1,2,1,0,1,1,2,2,0,2,2,2,1,0,0,1,2,0,1,0,0,2,0,2]

> size g
65

> size (removeEdge 0 1 g)
125

> p = path [1..10]
> size p
18
> size (removeEdge 5 6 p)
19

> q = path [1..100]
> size q
198
[...]

PR #33, opened 2017-11-25
WIP: Add non-empty graphs

snowleopard:
This is work-in-progress to address #31. I still need to fix Haddock comments and add tests.

@phadej Could you have a brief look? 

I don't particularly like the `foo1` naming convention used in `Data.List.NonEmpty`, so I'm tempted to find an alternative naming. Any suggestions?

### DIFF
--- a/src/Algebra/Graph/NonEmpty.hs
+++ b/src/Algebra/Graph/NonEmpty.hs
@@ -1,0 +1,747 @@
+{-# LANGUAGE DeriveFunctor, DeriveFoldable, DeriveTraversable #-}
+-----------------------------------------------------------------------------
+-- |
+-- Module     : Algebra.Graph.NonEmpty
+-- Copyright  : (c) Andrey Mokhov 2016-2017
+-- License    : MIT (see the file LICENSE)
+-- Maintainer : andrey.mokhov@gmail.com
+-- Stability  : experimental
+--
+-- __Alga__ is a library for algebraic construction and manipulation of graphs
+-- in Haskell. See <https://github.com/snowleopard/alga-paper this paper> for the
+-- motivation behind the library, the underlying theory, and implementation details.
+--
+-- This module defines the data type 'NonEmptyGraph' and associated algorithms.
+--
+-----------------------------------------------------------------------------
+module Algebra.Graph (
+    -- * Algebraic data type for graphs
+    NonEmptyGraph (..),
+
+    -- * Basic graph construction primitives
+    vertex, edge, overlay, connect, vertices, edges, overlays, connects,
+    graph,
+
+    -- * Graph folding
+    foldg,
+
+    -- * Relations on graphs
+    isSubgraphOf, (===),
+
+    -- * Graph properties
+    isEmpty, size, hasVertex, hasEdge, vertexCount, edgeCount, vertexList,
+    edgeList, vertexSet, vertexIntSet, edgeSet,
+
+    -- * Standard families of graphs
+    path, circuit, clique, biclique, star, tree, forest, mesh, torus, deBruijn,
+
+    -- * Graph transformation
+    removeVertex, removeEdge, replaceVertex, mergeVertices, splitVertex,
+    transpose, induce, simplify,
+
+    -- * Graph composition
+    box
+  ) where
+
+import Prelude ()
+import Prelude.Compat
+
+import Control.Applicative (Alternative, (<|>))
+import Control.DeepSeq (NFData (..))
+import Control.Monad.Compat
+
+import qualified Algebra.Graph.AdjacencyMap       as AM
+import qualified Algebra.Graph.Class              as C
+import qualified Algebra.Graph.HigherKinded.Class as H
+import qualified Algebra.Graph.Relation           as R
+import qualified Data.IntSet                      as IntSet
+import qualified Data.Set                         as Set
+import qualified Data.Tree                        as Tree
+
+data NonEmptyGraph a = Vertex a
+                     | Overlay (NonEmptyGraph a) (NonEmptyGraph a)
+                     | Connect (NonEmptyGraph a) (NonEmptyGraph a)
+                     deriving (Foldable, Functor, Show, Traversable)
+
+instance NFData a => NFData (NonEmptyGraph a) where
+    rnf (Vertex a)    = rnf a
+    rnf (Overlay x y) = rnf x `seq` rnf y
+    rnf (Connect x y) = rnf x `seq` rnf y
+
+-- instance C.Graph (Graph a) where
+--     type Vertex (Graph a) = a
+--     empty   = empty
+--     vertex  = vertex
+--     overlay = overlay
+--     connect = connect
+
+-- instance C.ToGraph (NonEmptyGraph a) where
+--     type ToVertex (NonEmptyGraph a) = a
+--     toGraph = foldg C.empty C.vertex C.overlay C.connect
+
+-- instance H.ToGraph NonEmptyGraph where
+--     toGraph = foldg H.empty H.vertex H.overlay H.connect
+
+-- instance H.Graph NonEmptyGraph where
+--     connect = connect
+
+instance Ord a => Eq (Graph a) where
+    x == y = C.toGraph x == (C.toGraph y :: AM.AdjacencyMap a)
+
+instance Applicative Graph where
+    pure  = Vertex
+    (<*>) = ap
+
+instance Monad Graph where
+    return  = pure
+    g >>= f = foldg Empty f Overlay Connect g
+
+-- | Construct the graph comprising /a single isolated vertex/. An alias for the
+-- constructor 'Vertex'.
+-- Complexity: /O(1)/ time, memory and size.
+--
+-- @
+-- 'isEmpty'     (vertex x) == False
+-- 'hasVertex' x (vertex x) == True
+-- 'hasVertex' 1 (vertex 2) == False
+-- 'vertexCount' (vertex x) == 1
+-- 'edgeCount'   (vertex x) == 0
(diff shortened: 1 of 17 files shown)


phadej:
If you plan to add e.g. `edges1` to `Algebra.Graph` (and other) modules, than I'd keep the 1-suffix. At least Edward Kmett has `foldMap1` in [`Data.Semigroup.Foldable`](http://hackage.haskell.org/package/semigroupoids-5.2.1/docs/Data-Semigroup-Foldable.html) and plenty of `something1` combinators in `lens`, e.g.

src/Algebra/Graph/NonEmpty.hs
```diff
@@ -0,0 +1,563 @@
...
+-- box x ('overlay' y z)   == 'overlay' (box x y) (box x z)
+-- box x ('vertex' ())     ~~ x
+-- 'transpose'   (box x y) == box ('transpose' x) ('transpose' y)
+-- 'vertexCount' (box x y) == 'vertexCount' x * 'vertexCount' y
+-- 'edgeCount'   (box x y) <= 'vertexCount' x * 'edgeCount' y + 'edgeCount' x * 'vertexCount' y
+-- @
+box :: NonEmptyGraph a -> NonEmptyGraph b -> NonEmptyGraph (a, b)
+box x y = overlays1 xs `overlay` overlays1 ys
+  where
+    xs = fmap (\b -> fmap (,b) x) $ fromJust $ NonEmpty.nonEmpty $ toList y
```
phadej: should there be `toNonEmpty`?

src/Algebra/Graph/NonEmpty.hs
```diff
@@ -0,0 +1,563 @@
...
+-- @
+clique1 :: NonEmpty a -> NonEmptyGraph a
+clique1 = connects1 . fmap vertex
+
+-- | The /biclique/ on a list of vertices.
+-- Complexity: /O(L1 + L2)/ time, memory and size, where /L1/ and /L2/ are the
+-- lengths of the given lists.
+--
+-- @
+-- biclique (x1 :| [x2]) (y1 :| [y2]) == 'edges' [(x1,y1), (x1,y2), (x2,y1), (x2,y2)]
```
phadej: biclique1

src/Algebra/Graph/NonEmpty.hs
```diff
@@ -0,0 +1,563 @@
...
+-- @
+edges1 :: NonEmpty (a, a) -> NonEmptyGraph a
+edges1 = overlays1 . fmap (uncurry edge)
+
+-- | Overlay a given list of graphs.
+-- Complexity: /O(L)/ time and memory, and /O(S)/ size, where /L/ is the length
+-- of the given list, and /S/ is the sum of sizes of the graphs in the list.
+--
+-- @
+-- overlays1 [x]   == x
```
phadej: not nonempty aragument

src/Algebra/Graph/NonEmpty.hs
```diff
@@ -0,0 +1,563 @@
...
+import qualified Algebra.Graph.AdjacencyMap       as AM
+import qualified Algebra.Graph.Class              as C
+import qualified Algebra.Graph.HigherKinded.Class as H
+import qualified Algebra.Graph.IntAdjacencyMap    as IAM
+import qualified Algebra.Graph.Relation           as R
+import qualified Data.IntSet                      as IntSet
+import qualified Data.List.NonEmpty               as NonEmpty
+import qualified Data.Set                         as Set
+
+data NonEmptyGraph a = Vertex a
```
phadej: worth documenting that non-empty means that there's always at least one vertex, but there might be zero edges!

src/Algebra/Graph/NonEmpty.hs
```diff
@@ -0,0 +1,563 @@
...
+-- 'hasVertex' z (overlay x y) == 'hasVertex' z x || 'hasVertex' z y
+-- 'vertexCount' (overlay x y) >= 'vertexCount' x
+-- 'vertexCount' (overlay x y) <= 'vertexCount' x + 'vertexCount' y
+-- 'edgeCount'   (overlay x y) >= 'edgeCount' x
+-- 'edgeCount'   (overlay x y) <= 'edgeCount' x   + 'edgeCount' y
+-- 'size'        (overlay x y) == 'size' x        + 'size' y
+-- 'vertexCount' (overlay 1 2) == 2
+-- 'edgeCount'   (overlay 1 2) == 0
+-- @
+overlay :: NonEmptyGraph a -> NonEmptyGraph a -> NonEmptyGraph a
```
phadej: `Data.List.NonEmpty` also misses appends like:

```haskell
appendl :: [a] -> NonEmpty a -> NonEmpty a
appendr :: NonEmpty a -> [a] -> NonEmpty a
```

good to keep in mind, but coming up with good names is difficult.

src/Algebra/Graph/NonEmpty.hs
```diff
@@ -0,0 +1,563 @@
...
+overlays1 :: NonEmpty (NonEmptyGraph a) -> NonEmptyGraph a
+overlays1 = foldr1 overlay
+
+-- | Connect a given list of graphs.
+-- Complexity: /O(L)/ time and memory, and /O(S)/ size, where /L/ is the length
+-- of the given list, and /S/ is the sum of sizes of the graphs in the list.
+--
+-- @
+-- connects1 [x]   == x
+-- connects1 [x,y] == 'connect' x y
```
phadej: also arg

src/Algebra/Graph/NonEmpty.hs
```diff
@@ -0,0 +1,563 @@
...
+
+-- | The /clique/ on a list of vertices.
+-- Complexity: /O(L)/ time, memory and size, where /L/ is the length of the
+-- given list.
+--
+-- @
+-- clique1 (x :| []   ) == 'vertex' x
+-- clique1 (x :| [y]  ) == 'edge' x y
+-- clique1 (x :| [y,z]) == 'edges1' ((x,y) :| [(x,z), (y,z)])
+-- clique1 (xs ++ ys) == 'connect' (clique1 xs) (clique1 ys)
```
phadej: `++ -> <>`

src/Algebra/Graph/NonEmpty.hs
```diff
@@ -0,0 +1,563 @@
...
+
+-- | Construct the graph from a list of edges.
+-- Complexity: /O(L)/ time, memory and size, where /L/ is the length of the
+-- given list.
+--
+-- @
+-- edges1 [(x,y)]     == 'edge' x y
+-- 'edgeCount' . edges1 == 'Data.List.NonEmpty.length' . 'Data.List.NonEmpty.nub'
+-- @
+edges1 :: NonEmpty (a, a) -> NonEmptyGraph a
```
phadej: side comment: should non-empty variants be generalised to `Foldable f => f (a, a) -> Graph a` ?

snowleopard:
@phadej Many thanks for reviewing! I'll take care of the suggestions over the weekend.

> If you plan to add e.g. `edges1` to `Algebra.Graph`

Hmm, why would we want `Algebra.Graph.edge1`? It's type signature can't match the one defined in this module. And having something like `edge1 :: NonEmpty (a, a) -> Graph a` seems not very useful.

> At least Edward Kmett has `foldMap1` in `Data.Semigroup.Foldable` and plenty of `something1` combinators in `lens`, e.g.

Yes, this seems to be a wide-spread convention. I don't like it, but I couldn't come up with anything better (e.g. `nonEmptyEdges` is just awful), so I'm going to adopt it too.

src/Algebra/Graph/NonEmpty.hs
snowleopard: Having plain types like `[a] -> Graph a` seems friendlier for newcomers.

I was thinking to make type signatures fully polymorphic in modules `Algebra.Graph.Class` and `Algebra.Graph.HigherKinded.Class` instead: these modules are a bit more advanced anyway, so their users are probably comfortable with types like `(Foldable f, Graph g) => f a -> g a`.

This allows you to `import Algebra.Graph.HigherKinded.Class (vertices)` and get a fully polymorphic version of `vertices` in scope. Does this sound like good idea?

phadej:
@snowleopard true, `edges1` is a little weird for `Graph`, that's why generalising 
```
edges :: Foldable f => f (a, a) -> Graph a
```

would make sense, as `NonEmpty` and many other things are `Foldable`

It would be nice to use `Foldable1` too, but `semigroupoids` is quite heavy dependency unfortunately.

snowleopard:
Yes, using `Foldable1` would be nice for consistency, but I'll wait for it to propagate to core libraries :)

src/Algebra/Graph/NonEmpty.hs
snowleopard: Good point. For `overlay` we could probably provide only one `overlay1` function due to commutativity, but for `connect` we'll need both.

I decided not to add these functions for now, because I don't expect users to mix non-empty and general graphs in a single project. We'll see if there will be any demand.

In general, this seems to be a perfect use-case for subtyping. Just for the sake of curiousity I'll give it a try in my draft Scala implementation.

src/Algebra/Graph/NonEmpty.hs
snowleopard: I've added `toNonEmpty` but not sure if we need to expose it. Shall I?

At some point it will become available via `Foldable1`.

snowleopard:
I think this is done.

The only missing bit is `removeEdge`, but while working on #15 I came up with an approach that will also lead to much easier implementation of `removeEdge` for non-empty graphs, so I'll add as part of #15.

Issue #13, opened 2017-03-26
Reduce code duplication in testsuite


snowleopard:
There is a lot of code duplication in the testsuite currently. For example, see this commit: https://github.com/snowleopard/alga/commit/fc2854de76e7e05be5fd09c362750d6db9a05502.

It should be possible to get rid of most duplication by defining generic testsuites like [this one](https://github.com/snowleopard/alga/blob/master/test/Algebra/Graph/Test.hs#L52-L61).

There is also duplication in the documentation (see the same commit example above), but I'm not sure if it's possible to do anything about it.

anfelor:
Have you looked into [doctest](https://github.com/sol/doctest)? I think it would go a long way towards reducing code duplication.

snowleopard:
@anfelor Thanks for the pointer! I haven't tried it but it looks promising.

It will likely not be able to handle all examples, as some of them use function equality. We need a proof-of-concept implementation.

nobrakal:
So I tried with `doctest`. Took some time to set up (the need to hide some of Prelude, etc), but I came to a partial implementation of tests of `Algebra.Graph`.

You can see the (very very partial, in fact just for the data definition and the first function) results here: https://github.com/nobrakal/alga/commit/6a65eb19ba4a5fb7934f16e7e66a0bfa4b429693

With a "clever" setup (where I define the (+) and (*) operators, and copied/pasted the `Arbitrary` instance of `Graph Int`), doctest detect and test everything like:
```Haskell
-- prop>       (x + y) == (y + x)
-- prop>   x + (y + z) == (x + y) + z
```
and
```Haskell
-- >>> isEmpty empty
-- True
```
So the pros:
* It definitely reduce code duplication :)

Cons:
* The `$setup` part is not beautiful at all, but I think it could be improved
* About haddock, the doc will be less pretty, because haddock render `prop>` line-by-line. I also think you cannot "reference" a function in a `prop>` (make a link form its name to its definition)

Question:
* Not sure how to handle things like 
```Haskell
isEmpty empty     == True
```
`prop>` make a pretty output but is not very made for it, but the syntax
```Haskell
-- >>> isEmpty empty
-- True
```
render not really great in haddock.

What do you think ?

snowleopard:
[...]

PR #47, opened 2018-01-27
escape strings to satisfy Dot syntax rules

jmtd:
GraphViz's "dot" language has several options for encoding IDs, as
described here: https://www.graphviz.org/doc/info/lang.html

Alga makes use of the quoted-string syntax. This permits any character
within double quotes, except the double-quote character itself, which
must be prefixed by a backslash character. (the backslash itself does
not need to be escaped).

In order to implement this requirement, introduce an "Escape" typeclass
and "escape" function.

Provide an instance definition for regular Haskell Strings. This
requires the FlexibleInstances Syntax extension.


jmtd:
Things *I* know need doing (at least)

 - [ ] tests
 - [ ] GHC warns about redundant constraints for some functions, to investigate
 - [ ] docs
 - [ ] do the non-show export methods still work as expected

jmtd:
Simple test program

```
import Algebra.Graph
import Algebra.Graph.Export.Dot
main = do
    putStrLn test
    putStrLn $ show $ test == correct
    putStrLn (export defaultStyleViaShow (path [1, 2]) :: String)

test = (export defaultStyleViaShow (path ["a", "b"]) :: String)
correct = "digraph\n\
\{\n\
\  \"\\\"a\\\"\"\n\
\  \"\\\"b\\\"\"\n\
\  \"\\\"a\\\"\" -> \"\\\"b\\\"\"\n\
\}\n"
```

src/Algebra/Graph/Class.hs
```diff
@@ -395,3 +399,8 @@ forest = overlays . map tree
 class ToGraph t where
     type ToVertex t
     toGraph :: (Graph g, Vertex g ~ ToVertex t) => t -> g
+
+class Escape s where
+    escape :: s -> s
+instance Escape String where
+    escape = concatMap (\x-> if x == '"' then "\\\"" else [x])
```
snowleopard: Since this is DOT-specific escaping, I suggest to move this class to `Export.Dot`.

src/Algebra/Graph/Export/Dot.hs
```diff
@@ -71,8 +71,8 @@ defaultStyle v = Style mempty mempty [] [] [] v (\_ -> []) (\_ _ -> [])
 -- @
 -- defaultStyleViaShow = 'defaultStyle' ('fromString' . 'show')
 -- @
-defaultStyleViaShow :: (Show a, IsString s, Monoid s) => Style a s
-defaultStyleViaShow = defaultStyle (fromString . show)
+defaultStyleViaShow :: (Show a, IsString s, Escape s, Monoid s) => Style a s
+defaultStyleViaShow = defaultStyle (fromString . escape . show)
```
snowleopard: Ahh, I just realised that since we can escape **after** converting to `String` we don't need the constraint, and we might not even need the typeclass!

snowleopard:
@jmtd Thank you! I just realised that I might have overcomplicated things by suggesting to add the type class, since we can escape after we have the `String`, just like you do in your implementation.

As for generic `export`, I don't know if we want to give the user full control and escaping themselves, or shall we provide built-in escaping?

src/Algebra/Graph/Export/Dot.hs
jmtd: I've just realised this too... at least in the context of defaultStyleViaShow.  I don't know whether you'd agree Alga should sanitize some of the other paths e.g. exportAsIs and whether we'd need the typeclass for that.

jmtd:
> As for generic export, I don't know if we want to give the user full control and escaping themselves, or shall we provide built-in escaping?

I suppose the question is, since Alga.Graph.Dot generates the output for graphviz, do you want to promise it will output something valid, no matter construction the user has thrown at you, and whether that is feasible.

snowleopard:
@jmtd What if the user has already provided escaped strings, e.g. we have `\"x\"` in the input? Do we want to mess this up by changing it to `\\"x\\"`. Or are we going to try to be even more clever and leave `\"` as is? I think this might become a maintenance hazard. It may be better to simply warn the users in the comments that `export` and `exportAsIs` do not perform escaping. (And test that they indeed do not.)

Note that in `exportViaShow` we know for sure that input strings are not escaped, because `Show` instances are expected to produce valid Haskell. Therefore, we can safely apply your escaping implementation.

jmtd:
I've just pushed a simplified patch that only implements escape for defaultStyleViaShow. I've gone for a simpler escape implementation than earlier which I think performs much better and is debatably simpler to understand too. I still need to write tests and docs.

snowleopard:
@jmtd Thanks! I have a suspicion that it might be too simple and fail to handle this case correctly:

```haskell
-- A string containing a single character "
test1 :: String
test1 = "\""
```

Will the current implementation turn it into `\"\\"\"`

Perhaps, we need to escape the character `\` too? Since DOT uses it for escaping, presumably it should be handled with care, producing `\"\\\"\"`, which I guess is the correct output for `test1`.

We might also add a test specifically for `\`:

```haskell
-- A string containing a single character \
test2 :: String
test2 = "\\"
```

jmtd:
Yes, the current implementation will generate that, and dot will not accept it:

```
> putStrLn $ (exportViaShow $ vertex "\"" :: String)
digraph
{
  "\"\\"\""
}
```
Relatedly, a strict reading of the Dot language description would lead one to believe `"\"` was a valid identifer, but it is not: there's a gap between the description and actual practice. For my graphviz, here, `"\\"` is accepted, and renders as a single `\`. I suppose this case could be handled in the existing pattern-match implementation with something inserted between the existing terms like (untested) `escape x:[] = if x = '\\' then x:x:[] else x:[])` (I've started contorting myself to avoid writing any more backslashes...)

(all these backslashes! even GH's markdown parser is having trouble :))

snowleopard:
My 'test framework' is not something I'm proud about and could definitely be improved (#13). However, for this PR it should be fairly straightforward to add tests somewhere around here:

https://github.com/snowleopard/alga/blob/master/test/Algebra/Graph/Test/Export.hs#L151-L162

snowleopard:
By the way, it may be useful to expose the escaping functionality for two reasons:
* It would be much easier to document and test the escaping functionality in isolation.
* Users can take advantage of it when creating their custom export `Style`s.

The cost is that we'll need to maintain it, but it doesn't seem to high.

PR #59, opened 2018-04-13
Allow base-compat-0.10

phadej:
A revision, or new release would be great! 

Note: PVP says next version can be `0.1.1.1`, as first three digits stays the same we "communicate" that there are no publicly visible changes in the API - which makes update a no-brainer. FWIW, 0.1.1 could been `0.1.0.1` - I was a little disappointed as there weren't "new features" in 0.1.1 :( - That said, it's not wrong, it's only not as semantically precise as it could be. I'm confused by those things myself.

### DIFF
--- a/algebraic-graphs.cabal
+++ b/algebraic-graphs.cabal
@@ -77,7 +77,7 @@ library
                         Algebra.Graph.Relation.Transitive
     build-depends:      array       >= 0.4     && < 0.6,
                         base        >= 4.7     && < 5,
-                        base-compat >= 0.9.1   && < 0.10,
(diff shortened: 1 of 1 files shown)


snowleopard:
Thanks @phadej! Merged. Indeed, I should have used `0.1.0.1` last time.

I'll make a new release `v0.1.1.1` today.

phadej:
@snowleopard thanks for prompt response!

snowleopard:
@phadej Done: https://hackage.haskell.org/package/algebraic-graphs-0.1.1.1

phadej:
Thanks!

Issue #54, opened 2018-03-29
Add NFData instances


anfelor:
While writing my [benchmark experiment](https://github.com/anfelor/graph-benchmarks) I noticed that no graph implementation but `data Graph` has a `NFData` instance. Is there a reason for this?

Adding them would for example provide a way to fully evaluate a `Relation a` without [relying on equality with itself](https://github.com/anfelor/graph-benchmarks/blob/12a0a46b1891fb402c663563c5874ea9a105e970/Common.hs#L28). The biggest design decision here seems to be the question of whether to evalute the `graphKL` field on `AdjacencyMap a`. On the one hand the docs clearly state that `rnf` should "fully evaluate all sub-components", but then the field is only meant to be cached for some computations that might not be needed. What do you think @snowleopard @nobrakal?

Related: #14

nobrakal:
I agree that having a `NFData` is a good idea. The problem is even more philosophical, what does it mean to "fully evaluate" a graph ?
Equality as you wrote? Or visiting each edge ? Each vertex ? Or maybe map or fold something on it ?
As a mathematical object, it is not clear at all to me...

anfelor:
> what does it mean to "fully evaluate" a graph?

It means to eliminate all lazy references from a data structure:
```haskell
> let x = Algebra.Graph.Relation.star 'a' "abcdefghij"
> :sprint x -- x is just a lazy reference at this moment
x = _
> Algebra.Graph.Relation.isEmpty  x
False
> :sprint x -- now the first field has been evaluated to normal form, but ...
x = Relation
      (containers-0.5.7.1:Data.Set.Base.Bin
         10 'g'
         (containers-0.5.7.1:Data.Set.Base.Bin
            6 'c'
            (containers-0.5.7.1:Data.Set.Base.Bin
               2 'b'
               (containers-0.5.7.1:Data.Set.Base.Bin
                  1 'a' containers-0.5.7.1:Data.Set.Base.Tip
                  containers-0.5.7.1:Data.Set.Base.Tip)
               containers-0.5.7.1:Data.Set.Base.Tip)
            (containers-0.5.7.1:Data.Set.Base.Bin
               3 'e'
               (containers-0.5.7.1:Data.Set.Base.Bin
                  1 'd' containers-0.5.7.1:Data.Set.Base.Tip
                  containers-0.5.7.1:Data.Set.Base.Tip)
               (containers-0.5.7.1:Data.Set.Base.Bin
[...]

Issue #24, opened 2017-11-11
rename Graph type.


phadej:
It's quite inconvenient to use `class` and `data` at the same time.

As `data Graph` directly encodes methods of the class, i'd name it `Syntax`, but that's a little weird name.

snowleopard:
@phadej I agree that it's very inconvenient, so thanks for raising this issue!

My plan was to eventually change the name of the type class, because the algebra works not only for graphs, but, for example, equivalence relations (combination of `Reflexive`, `Undirected` and `Transitive` subclasses), and all sorts of other strange instances.

On the other hand, a concrete data type `Graph` has a fixed `Eq` instance which currently corresponds exactly to (directed) graphs.

Having `data Graph` also seems to correspond nicely with standard libraries where we have build-in lists, `data Tree`, `data Forest` and other concrete data types. Naming this data type `Expr` (as I had before) or `Syntax` (as you suggest) feels a bit awkward...

phadej:
I see, renaming the class will work for me too.

snowleopard:
In a recent discussion with colleagues, `Plex` came up as a possible name for the type class (there seem to be a nice correspondence between the algebra and topological concepts like simplexes and complexes). I will write a blog post about this soon.

If you (or anyone else) have any suggestions for the name of the type class, please let me know!

glaebhoerl:
A simple and boring option is `IsGraph`.

snowleopard:
@glaebhoerl Yes, something like `IsGraph` is an option, but I believe we are looking for an abstract term. 

As an example, consider rings as a common abstraction for numbers. Then `instance Ring Integer` and `instance Ring Matrix` make more sense than `instance IsInteger Integer` and `instance IsInteger Matrix`.

So I want to be able to define `instance X Graph` and `instance X EquivalenceRelation` for some `X`. But I fear no standard `X` exists so we need to come up with a new term (e.g. `Plex`).

phadej:
One note: as there's `Graph` class, there could be `NonEmptyGraph`, (think `Monoid` and `Semigroup`).

one could have e.g.

```haskell
vertices1 :: NonEmptyGraph g => NonEmpty (Vertex g) -> g
vertices1 = foldr1 overlay . map vertex
```

In the paper 3.1 section could have

- (G,+) is a idempotent commutative semigroup
- (G,->) is a semigroup
- -> distributes over +

I'm quickly skimming thru the paper, and `empty` isn't used that much there!

---

[...]

Issue #19, opened 2017-07-15
Remove GraphKL from the API


snowleopard:
`GraphKL` records are defined in `AdjacencyMap` and `IntAdjacencyMap` modules for interoperability with King-Launchbury graphs. Specifically, they provide a way to reuse the algorithms from `Data.Graph`.

While reviewing #18, I realised that GraphKL's `getVertex` function is partial, and I'd like to avoid having partial functions in the API where possible.

I propose to remove `GraphKL`, because there are not too many algorithms left in `Data.Graph` that we haven't yet wrapped in a clean interface: `dfs` (which is added in #18), `bcc`, plus a few reachability tests.

My plan is therefore to provide clean interfaces to all of these functions instead of exposing `GraphKL`.

Please shout if you are using `GraphKL`!

snowleopard:
There is one performance aspect to consider: constructing `GraphKL` from `AdjacencyMap` takes time, so if the user would like to run multiple `Data.Graph` algorithms, it would be better to construct `GraphKL` once and then reuse it. If we drop `GraphKL`, we'll need to somehow cache it within `AdjacencyMap`.

feuerbach:
> If we drop GraphKL, we'll need to somehow cache it within AdjacencyMap.

Good point. Probably the easiest way is to have a lazy pair `(AdjacencyMap, GraphKL)` (or an equivalent data type) where one field is lazily constructed in terms of the other.

snowleopard:
@feuerbach Is this what you had in mind?

https://github.com/snowleopard/alga/blob/master/src/Algebra/Graph/AdjacencyMap/Internal.hs#L90-L102

I hope I'll finish the refactoring by the end of this week and will release the new version.

feuerbach:
Yes, it is what I had mind. I think you may want to make the `adjacencyMap` field strict.

snowleopard:
> I think you may want to make the `adjacencyMap` field strict.

Good point! I'll do this.

This brings me to the question I've been thinking about for some time (which probably deserves a separate issue, but I'll start here). 

Another interesting implementation to consider is:
```haskell
data AdjacencyMap a = Empty
                    | Vertex a
                    | Overlay (AdjacencyMap a) (AdjacencyMap a)
                    | Connect (AdjacencyMap a) (AdjacencyMap a)
                    | Cached a

data Cached a = Cached
    { expression   :: AdjacencyMap a   -- subexpression
[...]

PR #62, opened 2018-04-20
Draft for extracting King-Launchbury graphs from AdjacencyMap

anfelor:
This removes the `graphKL` field from `AdjacencyMap` as suggested in #54 and makes the `GraphKL` data type publicly available in `Algebra.Graph.KingLaunchbury` instead (similar to the proposal in #24). Changes:

 - Remove `graphKL` and `mkAM` from `AdjacencyMap`
 - Replace `dfs`, `topSort` and `scc` with slightly slower versions operating on pure `AdjacencyMap`s
 - Move the old ones into `Algebra.Graph.KingLaunchbury` and make them (largely) independent of `AdjacencyMap`

Not changed (yet):

 - The `graphKL` field and algorithms of `IntAdjacencyMap`
 - No `Graph a -> GraphKL a` conversion yet.
 - Comments are a mess.

The benefit of this is that the original behaviour of the API can still be recovered by just manually converting to the `GraphKL` type, while the unexpected behaviour of `dfs`, ... is fixed. What do you think of this approach?

### DIFF
--- a/src/Algebra/Graph/AdjacencyMap.hs
+++ b/src/Algebra/Graph/AdjacencyMap.hs
@@ -49,7 +49,7 @@ import Algebra.Graph.AdjacencyMap.Internal
 
 import qualified Algebra.Graph.Class as C
 import qualified Data.Graph.Typed    as Typed
-import qualified Data.Graph          as KL (scc)
+import qualified Data.Graph          as KL
 import qualified Data.Map.Strict     as Map
 import qualified Data.Set            as Set
 
@@ -620,7 +620,11 @@ dfs vs = concatMap flatten . dfsForestFrom vs
 -- fmap (flip 'isTopSort' x) (topSort x) /= Just False
 -- @
 topSort :: Ord a => AdjacencyMap a -> Maybe [a]
-topSort = Typed.topSort . Typed.fromAdjacencyMap
+topSort m =
+    if isTopSort result m then Just result else Nothing
+  where
+    (Typed.GraphKL g r _) = Typed.fromAdjacencyMap m
+    result = map r (KL.topSort g)
 
 -- | Check if a given list of vertices is a valid /topological sort/ of a graph.
 --
@@ -633,7 +637,11 @@ topSort = Typed.topSort . Typed.fromAdjacencyMap
 -- isTopSort [x]       ('edge' x x)      == False
 -- @
 isTopSort :: Ord a => [a] -> AdjacencyMap a -> Bool
-isTopSort xs = Typed.isTopSort xs . Typed.fromAdjacencyMap
+isTopSort xs m = go Set.empty xs
+  where
+    go seen []     = seen == Map.keysSet (adjacencyMap m)
+    go seen (v:vs) = let newSeen = seen `seq` Set.insert v seen
+        in postSet v m `Set.intersection` newSeen == Set.empty && go newSeen vs
 
 -- | Compute the /condensation/ of a graph, where each vertex corresponds to a
 -- /strongly-connected component/ of the original graph.
--- a/src/Algebra/Graph/AdjacencyMap.hs
+++ b/src/Algebra/Graph/AdjacencyMap.hs
@@ -623,8 +623,7 @@ topSort :: Ord a => AdjacencyMap a -> Maybe [a]
 topSort m =
     if isTopSort result m then Just result else Nothing
   where
-    (Typed.GraphKL g r _) = Typed.fromAdjacencyMap m
-    result = map r (KL.topSort g)
+    result = Typed.topSort (Typed.fromAdjacencyMap m)
 
 -- | Check if a given list of vertices is a valid /topological sort/ of a graph.
 --
--- a/src/Algebra/Graph/AdjacencyMap.hs
+++ b/src/Algebra/Graph/AdjacencyMap.hs
@@ -50,7 +50,6 @@ import Algebra.Graph.AdjacencyMap.Internal
 import qualified Algebra.Graph.Class as C
 import qualified Data.Map.Strict     as Map
 import qualified Data.Set            as Set
-import qualified Data.List           as List
 
 -- | Construct the /empty graph/.
 -- Complexity: /O(1)/ time and memory.
@@ -593,7 +592,9 @@ dfsForest g = dfsForestFrom (vertexList g) g
 dfsForestFrom :: Ord a => [a] -> AdjacencyMap a -> Forest a
 dfsForestFrom vs g = snd (chop Set.empty (map (go g) valid))
   where
-    valid = List.intersect vs (vertexList g)
+    -- We need to keep the order of vs
+    valid = filter (flip Set.member valid') vs
+    valid' = Set.intersection (Set.fromList vs) (Map.keysSet (adjacencyMap g))
 
     go :: Ord a => AdjacencyMap a -> a -> Tree a
     go g v = Node v (map (go g) (maybe [] Set.toList (Map.lookup v (adjacencyMap g))))
--- a/src/Algebra/Graph/AdjacencyMap.hs
+++ b/src/Algebra/Graph/AdjacencyMap.hs
@@ -48,6 +48,8 @@ import Data.Tree
 import Algebra.Graph.AdjacencyMap.Internal
 
 import qualified Algebra.Graph.Class as C
+import qualified Data.Graph.Typed    as Typed
+import qualified Data.Graph          as KL (scc)
 import qualified Data.Map.Strict     as Map
 import qualified Data.Set            as Set
 
@@ -590,21 +592,7 @@ dfsForest g = dfsForestFrom (vertexList g) g
 --                                                        , subForest = [] }]
 -- @
 dfsForestFrom :: Ord a => [a] -> AdjacencyMap a -> Forest a
-dfsForestFrom vs g = snd (chop Set.empty (map (go g) valid))
-  where
-    -- We need to keep the order of vs
-    valid = filter (flip Set.member valid') vs
-    valid' = Set.intersection (Set.fromList vs) (Map.keysSet (adjacencyMap g))
-
-    go :: Ord a => AdjacencyMap a -> a -> Tree a
-    go g v = Node v (map (go g) (maybe [] Set.toList (Map.lookup v (adjacencyMap g))))
-
-    chop :: Ord a => Set a -> Forest a -> (Set a, Forest a)
-    chop known [] = (known, [])
-    chop known (Node v ts:xs) = if Set.member v known then chop known xs
-      else let (known' , ts') = chop (Set.insert v known) ts
-               (known'', xs') = chop known' xs
-           in (known'', Node v ts' : xs')
+dfsForestFrom vs = Typed.dfsForestFrom vs . Typed.fromAdjacencyMap
 
 -- | Compute the list of vertices visited by the /depth-first search/ in a graph,
 -- when searching from each of the given vertices in order.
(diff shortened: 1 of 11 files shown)


anfelor:
One thing I forgot: at the moment there are two `toAdjacenyMap` functions and I am not sure which will be faster.

src/Algebra/Graph/AdjacencyMap.hs
```diff
@@ -591,8 +590,20 @@ dfsForest (AM _ (GraphKL g r _)) = fmap (fmap r) (KL.dff g)
 --                                                 , Node { rootLabel = 4
 --                                                        , subForest = [] }]
 -- @
-dfsForestFrom :: [a] -> AdjacencyMap a -> Forest a
-dfsForestFrom vs (AM _ (GraphKL g r t)) = fmap (fmap r) (KL.dfs g (mapMaybe t vs))
+dfsForestFrom :: Ord a => [a] -> AdjacencyMap a -> Forest a
```
snowleopard: What is the time and memory complexity of this implementation?

anfelor:
The complexity of `dfsForestFrom` is:
```haskell
dfsForestFrom :: k -> (n=|V|, m=|E|) -> (p=|V|)
valid = O(k * n)
validBasedOnSets = O(n + k*log(k)) -- going to push this in a minute
chop = O(p * log(p)) -- traverse the whole forest and push elements into a set.
go = O(p * (log(n) + m/n)) -- for every reachable element get the edges to other elements (log(n)) 
 --and turn the set into a list (m/n).
dfsForestFrom = O(n + p * (log(n) + m/n)) -- we know that k <= p and p <= n
```

`dfs` and `dfsForest` only add linear factors to that.

anfelor:
What do you think about adding `postOrder` to the API by the way? I just noticed that I didn't make it public in my first commit. It would look like this: https://github.com/anfelor/alga/commit/a56591c62e7cc011cd5c40c82f8372c28bdcca96

snowleopard:
@anfelor My apologies for taking so long with this PR.

> maybe there are so many graph instances and so few algorithms operating on pure `Graph`s because Alga is still incomplete? 

Yes, algebraic graphs are in their infancy and lack many algorithms. Although it's fine to rely on `containers` for working around the lack of purely algebraic algorithms, I don't think we should be exposing these workarounds as part of Alga's API.

> I still think that `scc` could be useful as it is non-trivial and probably yields a speed-up compared to the pure `AdjacencyMap` implementation.

Then perhaps it should be hidden in the `AdjacencyMap` module, that is, the user of the library should not be able to call it directly, but only when operating on `Graph`s or `AdjacencyMap`s. This will make it easier to maintain Alga in the long term.

> Except for `isTopSort` and `scc` the functions are all one-liners, so I doubt that they would have any benefit for the containers team.

If you look at the source of `Data.Graph`, you'll see plenty of one-liners:

https://hackage.haskell.org/package/containers-0.5.11.0/docs/src/Data.Graph.html

The interface of `Data.Graph` is very old. It can benefit from a more modern, typed layer. It is definitely worth exploring it. Perhaps, this could be done in a separate library first, but the end goal should be to help *all* users of `Data.Graph`, not just those that happen to depend on Alga.

In principle, I'm OK with doing this exploration as part of Alga, but then it should live in a different namespace -- not in `Algebra.Graph`. What about `Data.Graph.Typed`? We could develop this module as part of Alga (and depend on it internally) and at some point, when it becomes mature enough, we could propose to merge it to `containers`. Does this sound like a good way forward?

> What do you think about adding `postOrder` to the API by the way?

Postorder DFS is indeed a useful component of some graph algorithms, but again I'm hesitant to extend the API just because it might be useful. I would prefer to wait for actual users of Alga to request this feature first.

anfelor:
> Then perhaps it should be hidden in the AdjacencyMap module, that is, the user of the library should not be able to call it directly, but only when operating on Graphs or AdjacencyMaps. 

I don't understand how to do this without using a hidden `graphKL` field? That was the central idea behind this experiment after all.

> What about Data.Graph.Typed?

If you really think it is worth doing, I could publish a `graph-extras` package to hackage and open a issue on the containers repository about adding some of it to the API. That way we keep the API of alga focused solely on algebraic graphs and the haskell community can decide whether it is worth adding.

> I'm hesitant to extend the API just because it might be useful. 

Sure. 

So, how do you feel about merging this code? Yes, maybe, not at all? It has the benefit of removing the surprising behavior of `graphKL` and making it explicit (in a very natural way, I believe) with the drawback of increasing the API surface area. Maybe we should wait for @nobrakals benchmarks to see whether it is worth it?

In any case, I see the following issues that yet need to be resolved, feel free to add yours to the list:

- [ ] Remove `postOrder` from the api
- [ ] Add the time complexities to the comments
- [ ] Decide whether to keep `scc` in the API

snowleopard:
> I don't understand how to do this without using a hidden `graphKL` field?

Like this, for example:
* Remove the `graphKL` field from `AdjacencyMap`.
* Implement methods like `dfs` by first converting an `AdjacencyMap` to a KL representation, and then passing the result to KL algorithms from `Data.Graph`.

It will no longer be possible to cache the construction of intermediate KL representation. Maybe this is not a big deal.

> If you really think it is worth doing, I could publish a graph-extras package to hackage and open a issue on the containers repository

What I meant is: rename the module in this PR to `Data.Graph.Typed` and add a note at the top that it is considered an internal module of Alga and the API can therefore change at any time (including removing the whole module).

If you do this and the other TODOs, then I think we can merge.

anfelor:
Sorry for taking so long to respond.

I am afraid I still don't see the way forward from here. As I have understood so far, the roadmap is as follows:
 - Rename `Algebra.Graph.KingLaunchbury` to `Data.Graph.Typed`
 - Remove the new functions operating on pure `AdjacencyMap`s and replace them with calls to `Data.Graph.Typed`, without caching the `graphKL` field.

However I feel like this kind of defeats the purpose of this PR, since in effect there will be just a decrease in performance (no cached representation) and maybe a slightly better organized code base, but not the choice between either using `Data.Graph` internally or not I had in mind.

If you are sure that that is what you want I can implement it though, but I wanted to cross-check first.

snowleopard:
> Remove the new functions operating on pure `AdjacencyMap`s 

Custom implementation of some graph algorithms developed specifically for `AdjacancyMap` is absolutely fine. But I think it may be best to leave this for a separate PR. Let's focus this PR on factoring out KL graphs into a separate module and removing the caching.

> there will be just a decrease in performance (no cached representation)

@anfelor The lack of caching doesn't mean necessarily reduced performance. It is merely an inconvenience.

Let's say you have `x :: AdjacencyMap a`. You can currently run `dfs [1, 2] x` and then `dfs [1, 3] x` and the second invocation will be faster thanks to caching. But with your PR users will still be able to achieve the same effect but without magic caching -- instead they will have to explicitly convert the graph into a new representation, e.g. `y = toGraphKL x` and then apply `dfs` to `y`, so both `dfs [1, 2] y` and `dfs [1, 3] y` will be fast. This is more explicit and a bit less convenient, but this is conceptually simpler and avoids surprises. As a side benefit this allows us to forget about the `NFData` instance dilemma #54.

anfelor:
> Let's focus this PR on factoring out KL graphs into a separate module and removing the caching.

Sure!

>  instead they will have to explicitly convert the graph into a new representation, e.g. y = toGraphKL x and then apply dfs to y, so both dfs [1, 2] y and dfs [1, 3] y will be fast.

Ah okay, then I misunderstood you before. So basically the final structure will be:

 - A module `Data.Graph.Typed` with the `GraphKL` type.
 - In `AdjacencyMap` functions like `dfs :: [a] -> GraphKL a -> [a]`

My only complaint would be that a stable module of the API (`AdjacencyMap`) will depend on a unstable one (`Data.Graph.Typed`) and you would need to create a `GraphKL` to use essential functions on `AdjacencyMap`s. But I think it is the right way to solve this.

anfelor:
@snowleopard I think this is ready to merge!
A possible area for improvement might be to add `Data.Graph.Typed` to the test suite, especially `toAdjacencyMap . fromAdjacencyMap = id` and vice versa. Feel free to write me if you want something else.

anfelor:
Ah, and there are two `toAdjacenyMap` functions in `Data.Graph.Typed`, because I didn't know which one would be fastest. We could:

 1. Leave both in there as the module is internal
 2. Delete both as they are not used in the AdjacencyMap code
 3. Benchmark them and remove the slower one

Which way do you prefer?

src/Data/Graph/Typed.hs
```diff
@@ -0,0 +1,156 @@
+-----------------------------------------------------------------------------
+-- |
+-- Module     : Algebra.Graph.KingLaunchbury
+-- Copyright  : (c) Anton Lorenzen 2016-2018
+-- License    : MIT (see the file LICENSE)
+-- Maintainer : andrey.mokhov@gmail.com
```
snowleopard: @anfelor This looks a bit odd: you hold the copyright, but I'm the maintainer. I guess it makes sense to list both of us in both fields to avoid confusion. (For example, if someone finds a bug in this module, I hope you're OK to fix it!)

test/Algebra/Graph/Test/AdjacencyMap.hs
```diff
@@ -70,12 +69,3 @@ testAdjacencyMap = do
                                            , (Set.fromList [1,4], Set.fromList [5]  )
                                            , (Set.fromList [3]  , Set.fromList [1,4])
                                            , (Set.fromList [3]  , Set.fromList [5 :: Int])]
-
-    putStrLn "\n============ AdjacencyMap.Internal.GraphKL ============"
```
snowleopard: I think this shouldn't be just deleted -- you should add a test module for testing all examples in `Data.Graph.Typed`.

snowleopard:
> A possible area for improvement might be to add `Data.Graph.Typed` to the test suite

@anfelor Yes, please test all laws and examples from `Data.Graph.Typed`.

Regarding two implementations of `toAdjacenyMap`: it would be best to pick one if there is a clear winner. Perhaps, we can use the benchmark suite that @nobrakal has been working on to pick the best one?

snowleopard:
> Delete both as they are not used in the AdjacencyMap code

On a second thought, if they are not used, I think we should just delete them. I think we only use King-Launchbury graphs for consuming graphs, so we probably don't need a `toAdjacencyMap` function.

src/Data/Graph/Typed.hs
anfelor: Of course I am! I have also added you to the copyright, since most of the code was originally written by you.

snowleopard:
@anfelor Ping me when this is ready.

anfelor:
@snowleopard, it is ready! For my part this is ready to be merged.

test/Data/Graph/Test/Typed.hs
```diff
@@ -0,0 +1,167 @@
...
+(%) :: (GraphKL Int -> a) -> AM.AdjacencyMap Int -> a
+a % g = a $ fromAdjacencyMap g
+
+testTyped :: IO ()
+testTyped = do
+    putStrLn "\n============ Typed ============"
+
+    putStrLn "\n============ Typed.fromAdjacencyMap ============"
+
+    test "toGraphKL (fromAdjacencyMap (1 * 2 + 3 * 1)) == array (0,2) [(0,[1]),(1,[]),(2,[0])]" $
```
snowleopard: @anfelor These are useful examples, please add them to Haddock.

src/Data/Graph/Typed.hs
```diff
@@ -0,0 +1,153 @@
+-----------------------------------------------------------------------------
+-- |
+-- Module     : Algebra.Graph.KingLaunchbury
```
snowleopard: Fix the module name.

test/Data/Graph/Test/Typed.hs
```diff
@@ -0,0 +1,167 @@
+-----------------------------------------------------------------------------
+-- |
+-- Module     : Algebra.Graph.Test.IntAdjacencyMap
```
snowleopard: Fix the module name.

snowleopard:
@anfelor Thanks! Spotted a couple of more issues -- please see the comments above.

snowleopard:
@feuerbach I'd like to let you know that this PR removes the caching behaviour in `AdjacencyMap` that we previously introduced (back in #19). The reason is: it turned out to be a bit too magic; for example, it's unclear what should be the right `NFData` instance in presence of cached fields.

I hope this will not be an issue for you, but if it is, please let us know and we'll think what we can do.

feuerbach:
Fine with me!

snowleopard:
@feuerbach Thanks!

snowleopard:
@anfelor Oh, it looks like I've introduced conflicts by my latest commits.

Should have merged this first. sorry!

anfelor:
@snowleopard No problem, the necessary changes were all fairly mechanical.

anfelor:
@snowleopard The linux ghc-8.4.1 build failed due to a travis install error, maybe the build script is wrong?

snowleopard:
@anfelor It was just a glitch -- fixed after restarting. 

I've merged the PR, many thanks!

Issue #12, opened 2017-03-25
Overview of functionality/performance of different graph instances


snowleopard:
It would be great to have a table (say, in the `doc` folder) comparing different graph instances in terms of provided functionality (e.g., does it have `removeVertex`?) and performance (e.g. how fast is `removeVertex`?).

mateoatr:
By performance you mean benchmarking these functions (`removeVertex`, `removeEdge`, `mergeVertices`,...)?

PR #64, opened 2018-05-04
Add overview of functionality/performance for different graph transfo…

mateoatr:
…rmations in doc

See issue https://github.com/snowleopard/alga/issues/12

### DIFF
--- a/doc/sketch.md
+++ b/doc/sketch.md
@@ -1,0 +1,83 @@
+This documents intends to give an overview of the time complexity of the different graph transformations implemented by the various data types existent in Alga.
+
+When specifying the time and memory complexity of graph algorithms, _n_ will denote the number of vertices in the graph, _m_ will denote the number of edges in the graph, and _s_ will denote the _size_ of the corresponding `Graph` expression.
+
+**`removeVertex`**: Removes a vertex from a given graph. In `NonEmpty` graphs, returns `Nothing` if the resulting graph is empty.
+
+| removeVertex       | Time          | Memory |
+|--------------------|---------------|--------|
+| AdjacencyMap       | `O(n * log(n))` | `   ` |
+| Fold               | `O(s)` | `O(s)` |
+| HigherKinded.Class | `O(s)` | `O(s)` |
+| IntAdjacencyMap    | `O(log(n))` | `O(s)` |
+| NonEmpty           | `O(s)` | `   ` |
+| Relation           | `O(n+m)`| `   ` |
+
+**`removeEdge`**: Removes an edge from a given graph.
+
+|    removeEdge      | Time      | Memory |
+|--------------------|-----------|--------|
+| AdjacencyMap       | `O(log(n))` | `   ` |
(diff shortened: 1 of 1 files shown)


doc/sketch.md
```diff
@@ -0,0 +1,83 @@
...
+**`removeVertex`**: Removes a vertex from a given graph. In `NonEmpty` graphs, returns `Nothing` if the resulting graph is empty.
+
+| removeVertex       | Time          | Memory |
+|--------------------|---------------|--------|
+| AdjacencyMap       | `O(n * log(n))` | `   ` |
+| Fold               | `O(s)` | `O(s)` |
+| HigherKinded.Class | `O(s)` | `O(s)` |
+| IntAdjacencyMap    | `O(log(n))` | `O(s)` |
+| NonEmpty           | `O(s)` | `   ` |
+| Relation           | `O(n+m)`| `   ` |
```
snowleopard: You didn't cover `Graph` in this table -- is this because the implementation simply reuses `HigherKinded.Class.removeVertex`?

I think it would be nice to have an entry for `Graph` as well, since at some point we might change the implementation and it will no longer rely on `HigherKinded.Class`.

doc/sketch.md
```diff
@@ -0,0 +1,83 @@
+This documents intends to give an overview of the time complexity of the different graph transformations implemented by the various data types existent in Alga.
+
+When specifying the time and memory complexity of graph algorithms, _n_ will denote the number of vertices in the graph, _m_ will denote the number of edges in the graph, and _s_ will denote the _size_ of the corresponding `Graph` expression.
+
+**`removeVertex`**: Removes a vertex from a given graph. In `NonEmpty` graphs, returns `Nothing` if the resulting graph is empty.
+
+| removeVertex       | Time          | Memory |
```
snowleopard: The first header here should probably be `Data type` instead of `removeVertex` (same for other tables).

doc/sketch.md
```diff
@@ -0,0 +1,83 @@
...
+
+|    removeEdge      | Time      | Memory |
+|--------------------|-----------|--------|
+| AdjacencyMap       | `O(log(n))` | `   ` |
+| Fold               | `O(s)` | `O(s)` |
+| IntAdjacencyMap    | `O(log(n))` | `   ` |
+| NonEmpty           | `O(s)` | `O(s)` |
+| Relation           | `O(log(m))` | `   ` |
+
+**`replaceVertex`**: Replaces a vertex `x` with a vertex `y` in a given `AdjacencyMap`. If `y` already exists, `x` and `y` will be merged.
```
snowleopard: `AdjacencyMap` -> `graph`.

snowleopard:
@mateoatr Many thanks for the PR! I've added a couple of comments above.

Also two general comments:
* For some reason you do not cover the most important data type `Graph` -- please add it to the tables.
* Could you please make sure the lines fit into around 80 characters, so it'll be easier to check diffs in future?

I haven't yet checked the complexity bounds themselves -- will do that soon.

PR #69, opened 2018-05-30
Use custom induce when possible in HigherKinded

nobrakal:
I found the times for `hasEdge` a bit big when I checked the results of my benchmarks (https://github.com/haskell-perf/graphs/blob/master/RESULTS_TIME.md#hasedge ).

I tried some improvements and they was too fine to be true, so I investigated a bit more, and found that `HigherKinded` use its proper `induce` ( `= mfilter` ), and the one in `Algebra.Graph` is faster.

Then, adding `induce` to the class-definition in `HigherKinded` with a default implementation allow the change to be minimal while improving performance:

For example with `hasEdge` of `Algebra.Graph` with the current implementation (`Alga`) and with mine (`Alga'`):
### Mesh

<TABLE>
   <TR>
      <TH>
      </TH>
      <TH CLASS = "thinright">
         1
      </TH>
      <TH CLASS = "thinright">
         10
      </TH>
      <TH>
         100
      </TH>
   </TR>
   <TR>
      <TH>
         Alga'
      </TH>
      <TD CLASS = "thinright">
         2.029 μs
      </TD>
      <TD CLASS = "thinright">
         4.421 μs
      </TD>
      <TD>
         21.49 μs
      </TD>
   </TR>
   <TR>
      <TH>
         Alga
      </TH>
      <TD CLASS = "thinright">
         2.102 μs
      </TD>
      <TD CLASS = "thinright">
         9.406 μs
      </TD>
      <TD>
         85.68 μs
      </TD>
   </TR>
</TABLE>

### Complete

<TABLE>
   <TR>
      <TH>
      </TH>
      <TH CLASS = "thinright">
         1
      </TH>
      <TH CLASS = "thinright">
         10
      </TH>
      <TH>
         100
      </TH>
   </TR>
   <TR>
      <TH>
         Alga'
      </TH>
      <TD CLASS = "thinright">
         4.815 μs
      </TD>
      <TD CLASS = "thinright">
         21.82 μs
      </TD>
      <TD>
         1.175 ms
      </TD>
   </TR>
   <TR>
      <TH>
         Alga
      </TH>
      <TD CLASS = "thinright">
         5.483 μs
      </TD>
      <TD CLASS = "thinright">
         67.54 μs
      </TD>
      <TD>
         7.573 ms
      </TD>
   </TR>
</TABLE>

SUMMARY:

 * Alga' was the fastest 13 times
There was 3 ex-aequo

ABSTRACT:

 * Alga' was 2.26 times faster than Alga

### DIFF
--- a/src/Algebra/Graph.hs
+++ b/src/Algebra/Graph.hs
@@ -172,6 +172,7 @@ instance H.ToGraph Graph where
 
 instance H.Graph Graph where
     connect = connect
+    induce = induce
 
 instance Num a => Num (Graph a) where
     fromInteger = Vertex . fromInteger
--- a/src/Algebra/Graph.hs
+++ b/src/Algebra/Graph.hs
@@ -437,7 +437,7 @@ hasVertex = H.hasVertex
 -- hasEdge x y                  == 'elem' (x,y) . 'edgeList'
 -- @
 hasEdge :: Ord a => a -> a -> Graph a -> Bool
-hasEdge = H.hasEdge
+hasEdge u v = (edge u v `isSubgraphOf`) . induce (`elem` [u, v])
 
 -- | The number of vertices in a graph.
 -- Complexity: /O(s * log(n))/ time.
@@ -694,7 +694,7 @@ deBruijn = H.deBruijn
 -- removeVertex x . removeVertex x == removeVertex x
 -- @
 removeVertex :: Eq a => a -> Graph a -> Graph a
-removeVertex = H.removeVertex
+removeVertex v = induce (/= v)
 
 -- | Remove an edge from a given graph.
 -- Complexity: /O(s)/ time, memory and size.
--- a/src/Algebra/Graph.hs
+++ b/src/Algebra/Graph.hs
@@ -172,7 +172,6 @@ instance H.ToGraph Graph where
 
 instance H.Graph Graph where
     connect = connect
-    induce = induce
 
 instance Num a => Num (Graph a) where
     fromInteger = Vertex . fromInteger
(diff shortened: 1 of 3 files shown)


nobrakal:
> In general, the main priority right now is Algebra.Graph module

Sorry if it wasn't clear: The `hasEdge` function of `Algebra.Graph` is implemented using the `hasEdge` of `Algebra.Graph.HigherKinded.Class`, himself using `induce`. But, it is use `induce` __from__ `HigherKinded.Class`, so `mfilter`.

This PR allow it to use the `induce` from `Algebra.Graph` and thus speedup the `hasEdge` of `Algebra.Graph` (and also the `removeVertex`, since it is implemented the same way).

> However, I don't want to extend the type class yet. I'd prefer to keep it to just 4 graph constructors.

For now, the class of `Graph` from `HigherKinded.Class` is just `connect`, this PR add only `induce`, with a default definition using `mfilter`


Otherwise, if you really don't want to change the `HigherKinded.Class` definition, I can copy paste the code from `HigherKinded.Class.hasEdge` to `Algebra.Graph.hasEdge`, and it will do the trick, but it is not so pretty ^^

snowleopard:
@nobrakal Ah, I see, thanks for clarifying! Please simply copy the implementation

``hasEdge u v = (edge u v `isSubgraphOf`) . induce (`elem` [u, v])``

Generic implementation is pretty much never the fastest, so I'm pretty sure we'll be providing more and more custom implementations, tailored for `Algebra.Graph`. 

By the way, I plan to invert the dependency between the modules at some point: `HigherKinded.Class` will actually be importing `Algebra.Graph` at some point, since I want `Algebra.Graph` to be fully standalone.

nobrakal:
Ok, done :) (also for `removeVertex`)

snowleopard:
Thanks, merged!

What is the performance impact on `removeVertex`?

nobrakal:
Just tested, about `1.42` times faster ;)

PR #70, opened 2018-06-01
Remove Graph dependency on Relation

nobrakal:
Hi,

This does not at all change the the complexity of `edgeSet` (the time ratio between the old implementation and the new is `1.01`).

But it remove the dependency of `Algebra.Graph` on `Algebra.Graph.Relation`, so I think it still an improvement :)


snowleopard:
I'm not sure about this. We have quite a big increase in complexity and the only benefit is removing the dependency on `Relation`. 

I'm sure this function can be implemented more efficiently and without relying on `Relation`, but I think the right solution is still to be found.

@nobrakal Let's wait until we have the benchmark suite finalised and then we will start opening issues for optimising individual functions, like `edgeSet`. That will surely require some discussions and experiments.

nobrakal:
Ok, you are right, it will be more convenient to discuss on issue :)

I'm closing this then

PR #72, opened 2018-06-04
NFData for AdjacencyMap

nobrakal:
Finally with the great work of @anfelor ,  a clear instance of `NFData` for `AdjacencyMap` is emerging.

This is obviously related to #54


anfelor:
I think there should be an equivalent one for IntAdjacencyMap?

nobrakal:
Of course :)

PR #73, opened 2018-06-07
Speedup vertices

nobrakal:
Hi,

I noticed that the function `vertices` was made of a map and a fold, it can be reduced to a fold alone: 
It divide the time took by the original function by 2

Also, this is pure make up, I used `foldr1` when possible, I think it is more clearer.


src/Algebra/Graph/HigherKinded/Class.hs
```diff
@@ -212,7 +212,9 @@ edge x y = connect (vertex x) (vertex y)
 -- 'vertexSet'   . vertices == Set.'Set.fromList'
 -- @
 vertices :: Graph g => [a] -> g a
-vertices = overlays . map vertex
```
snowleopard: I wonder if we can achieve the same speedup using rewrite rules? What if we add a function `overlaysMap` which will be implemented as you define below (but with an arbitrary function `f` instead of `vertex`) and then add a rewrite rule to optimise `overlays . map` to `overlaysMap`?

snowleopard:
@nobrakal Thanks! I would prefer not to use `foldr1` because it is a partial function. I know, we're safe here, but I'd like to avoid partial functions whenever possible.

Regarding your optimisation -- please see my comment. Can we instead introduce fused non-exported functions like `overlaysMap` and `connectsMap` and add rewrite rules from `overlays . map` and similarly from `connects . map` in `clique`?

This would be a lot more valuable, because it will automatically handle other similar cases, not only in the library (e.g. see function `edges`) but also in user code.

nobrakal:
Ok, I just reverted the last commit with `foldr1`, sorry for the noise.

I was not aware that RULES was used in user code ! So it is a very good idea, but I think the RULES code, and overlays, etc, must be moved in the `Algebra.Graph` module for this trick to work.

Also, I just tried and it work like a charm with
```Haskell
{-# RULES
 "overlays/map"    forall f xs.  overlays (map f xs) = overlaysMap f xs
  #-}
```
But it throw me warnings:
```
Rule "overlays/map" may never fire
      because ‘overlays’ might inline first
    Probable fix: add an INLINE[n] or NOINLINE[n] pragma for ‘overlays’
```

I solved it by adding a 
```Haskell
{-# NOINLINE overlays #-}
```

But I am not a GHC expert, and I don't know if this can really influence the code produced for `overlays`

snowleopard:
@nobrakal Well, I guess you can make the following test. Benchmark these cases:
* `vertices` without any optimisation.
* `vertices` with the rewrite rule.
* `overlays . map vertex` with the rewrite rule -- this will test whether the rule applies in user code.

I've never used rewrite rules before, so I suggest you study this topic. And then share what you learn :)

nobrakal:
So I have benchmarked three things:

* `vertices` is the proper vertices function
* `overlays` is the proper overlays function
* `RuleInUserCode` is the direct code of `vertices` inside the benchmark code

I did it against three variants:

* `REW` use the rewrite rules, with a NOINLINE for `overlays`
* `Witness` is the actual alga state
* `DirectOpt` is the code I proposed in the PR (so with only a direct optimization in `vertices`)

All benchs were run against a list of `10^6 Int`

```
benchmarking vertices/REW             
time                 2.917 ms   (2.866 ms .. 2.977 ms)
                     0.996 R²   (0.992 R² .. 0.998 R²)
mean                 3.086 ms   (3.031 ms .. 3.168 ms)
std dev              226.5 μs   (169.6 μs .. 319.8 μs)
variance introduced by outliers: 50% (severely inflated)
                                      
benchmarking vertices/Witness         
time                 5.469 ms   (5.363 ms .. 5.607 ms)
                     0.998 R²   (0.995 R² .. 1.000 R²)
mean                 5.382 ms   (5.352 ms .. 5.428 ms)
std dev              110.8 μs   (70.25 μs .. 191.7 μs)
                                      
benchmarking vertices/DirectOpt       
time                 3.118 ms   (3.094 ms .. 3.143 ms)
                     0.999 R²   (0.999 R² .. 1.000 R²)
mean                 3.138 ms   (3.122 ms .. 3.156 ms)
std dev              58.31 μs   (45.50 μs .. 79.68 μs)
                                      
benchmarking overlays/REW             
time                 3.572 ms   (3.489 ms .. 3.647 ms)
                     0.997 R²   (0.996 R² .. 0.999 R²)
mean                 3.478 ms   (3.451 ms .. 3.517 ms)
std dev              97.94 μs   (78.70 μs .. 129.6 μs)
variance introduced by outliers: 13% (moderately inflated)
                                      
benchmarking overlays/Witness         
time                 3.620 ms   (3.551 ms .. 3.696 ms)
                     0.997 R²   (0.994 R² .. 0.999 R²)
mean                 3.715 ms   (3.665 ms .. 3.793 ms)
std dev              186.7 μs   (128.2 μs .. 266.0 μs)
variance introduced by outliers: 30% (moderately inflated)
                                      
benchmarking overlays/DirectOpt       
time                 3.577 ms   (3.517 ms .. 3.663 ms)
                     0.998 R²   (0.997 R² .. 1.000 R²)
mean                 3.602 ms   (3.580 ms .. 3.631 ms)
std dev              75.78 μs   (59.49 μs .. 95.96 μs)
                                      
benchmarking RuleInUserCode/REW       
time                 3.143 ms   (3.065 ms .. 3.212 ms)
                     0.995 R²   (0.991 R² .. 0.997 R²)
mean                 3.123 ms   (3.073 ms .. 3.171 ms)
std dev              156.9 μs   (130.0 μs .. 186.1 μs)
variance introduced by outliers: 31% (moderately inflated)
                                      
benchmarking RuleInUserCode/Witness   
time                 4.637 ms   (4.564 ms .. 4.708 ms)
                     0.998 R²   (0.996 R² .. 0.999 R²)
mean                 4.826 ms   (4.766 ms .. 4.903 ms)
std dev              206.6 μs   (151.0 μs .. 308.4 μs)
variance introduced by outliers: 21% (moderately inflated)
                                      
benchmarking RuleInUserCode/DirectOpt 
time                 4.677 ms   (4.632 ms .. 4.731 ms)
                     0.999 R²   (0.999 R² .. 1.000 R²)
mean                 4.662 ms   (4.636 ms .. 4.694 ms)
std dev              86.64 μs   (63.17 μs .. 136.7 μs)
```

The conclusion: rewrite RULES are fantastics, they apply even in the user code! And the NOINLINE pragma on overlays don't seems to drop its performances. So its is the best choice from my point of view.

nobrakal:
Mmmhh, I was a bit optimistic yesterday, the optimization come from two parts.

Once again, `Algebra.Graph.HigherKinded.Class` is dropping `Algebraic.Graph` capabilities.  Please see below:

* NotHH is `Algebraic.Graph` with the `vertices` and `overlays` definitions of `Algebra.Graph.HigherKinded.Class` copy/pasted

* HH is the actual `Algebraic.Graph`

* REW is like `NotHH` PLUS a REWRITE rule like we discussed

* Dir is direct optimization of vertices INSIDE `Algebra.Graph`

I have benchmarked two things:

* The `vertices` function

* The definition of `vertices` but direclty into the benchmarking code, to see if the rewrite rule apply.

```
benchmarking vertices/NotHH           
time                 43.02 ms   (41.54 ms .. 44.74 ms)
                     0.996 R²   (0.990 R² .. 1.000 R²)
mean                 35.01 ms   (30.82 ms .. 37.45 ms)
std dev              6.041 ms   (3.404 ms .. 9.275 ms)
variance introduced by outliers: 60% (severely inflated)
                                      
benchmarking vertices/HH              
time                 52.47 ms   (50.77 ms .. 53.82 ms)
                     0.998 R²   (0.995 R² .. 1.000 R²)
mean                 44.47 ms   (39.52 ms .. 46.89 ms)
std dev              5.875 ms   (2.972 ms .. 8.973 ms)
variance introduced by outliers: 44% (moderately inflated)
                                      
benchmarking vertices/REW             
time                 29.38 ms   (28.84 ms .. 30.12 ms)
                     0.998 R²   (0.994 R² .. 1.000 R²)
mean                 23.71 ms   (21.21 ms .. 25.27 ms)
std dev              3.487 ms   (2.125 ms .. 5.197 ms)
variance introduced by outliers: 58% (severely inflated)
                                      
benchmarking vertices/Dir             
time                 22.88 ms   (22.62 ms .. 23.13 ms)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 18.30 ms   (16.77 ms .. 19.26 ms)
std dev              2.404 ms   (1.464 ms .. 3.457 ms)
variance introduced by outliers: 57% (severely inflated)
                                      
benchmarking overlays . map vertex/NotHH
time                 32.67 ms   (32.50 ms .. 32.85 ms)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 26.51 ms   (24.55 ms .. 27.99 ms)
std dev              3.320 ms   (2.033 ms .. 4.869 ms)
variance introduced by outliers: 48% (moderately inflated)
                                      
benchmarking overlays . map vertex/HH 
time                 46.29 ms   (44.71 ms .. 48.71 ms)
                     0.995 R²   (0.986 R² .. 0.999 R²)
mean                 39.92 ms   (36.61 ms .. 41.80 ms)
std dev              4.333 ms   (2.395 ms .. 6.638 ms)
variance introduced by outliers: 40% (moderately inflated)
                                      
benchmarking overlays . map vertex/REW
time                 30.73 ms   (29.32 ms .. 32.58 ms)
                     0.989 R²   (0.973 R² .. 0.999 R²)
mean                 24.27 ms   (21.85 ms .. 26.11 ms)
std dev              4.083 ms   (2.737 ms .. 6.101 ms)
variance introduced by outliers: 65% (severely inflated)
                                      
benchmarking overlays . map vertex/Dir
time                 33.62 ms   (33.13 ms .. 33.99 ms)
                     0.998 R²   (0.994 R² .. 1.000 R²)
mean                 27.67 ms   (24.56 ms .. 29.29 ms)
std dev              3.699 ms   (2.036 ms .. 5.458 ms)
variance introduced by outliers: 48% (moderately inflated)

```

So: 

* The rewrite rule does  apply in the userCode, which is a good news (if it has not applied, the time of `REW/overlays . map vertex` will have looks like then one from `overlays . map vertex/NotHH`). The time difference is not so big, I think it can come from other ghc-optimization inside the benchmarking code (I am passing to it a big list of Int, so the function made is not polymorphic)

* Direct implementation seems to be faster than using `overlaysMap`, I am not sure why

* A big optimization is done only by moving code from `Algebra.Graph.HigherKinded.Class` to `Algebra.Graph`, I will do some investigation about that ^^

snowleopard:
@nobrakal Thanks for benchmarking!

I think type classes often prevent some optimisations, so I'm not surprised that implementing thing directly in `Algebra.Graph` is faster. 

This PR pushed me to make a major refactoring of the library, which I planned for some time now: make the `Algebra.Graph` module self-contained and remove all dependencies on type classes: see #74.

nobrakal:
Oh great,  that is a big game changer :)

Some benchmarks using the `mono-graph` branch:

* DirectImplementation correspond to : 
```Haskell
vertices []     = Empty
vertices [x]    = Vertex x
vertices (x:xs) = Vertex x `Overlay` vertices xs
```Haskell

* OverlaysMap correspond to a rewritten `overlays . map Vertex` using a ghc rule using
```Haskell
overlaysMap _ []     = Empty
overlaysMap f [x]    = f x
overlaysMap f (x:xs) = f x `Overlay` vertices xs
```
*OverlaysPointMap is the current implementation 

```
benchmarking DirectImplementation
time                 27.96 ms   (27.52 ms .. 28.33 ms)
                     0.999 R²   (0.999 R² .. 1.000 R²)
mean                 28.76 ms   (28.46 ms .. 29.49 ms)
std dev              838.5 μs   (436.5 μs .. 1.417 ms)

benchmarking OverlaysMap                        
time                 30.44 ms   (30.22 ms .. 30.67 ms)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 30.51 ms   (30.40 ms .. 30.64 ms)
std dev              241.1 μs   (171.6 μs .. 346.6 μs)

benchmarking OverlaysPointMap
time                 40.13 ms   (39.85 ms .. 40.48 ms)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 39.92 ms   (39.67 ms .. 40.13 ms)
std dev              417.8 μs   (278.1 μs .. 608.9 μs)
```

So the direct implementation is faster. I think this is done by the abstraction of the function `f` in `overlaysMap`.
So I don't know what to do. Use a specialized function for each `overlays . map` in the code ? Or use the rewrite rule ?

snowleopard:
Hmm, not sure. Ideally we would use rewrite rules... Perhaps, we can analyse the resulting Core of the direct implementation and the one with the rewrite rule and to see why the former is faster?

nobrakal:
Here there are:
* Using rewrite rules
```Haskell
{-# RULES
  "overlays/map" forall f lst.  overlays (map f lst) = overlaysMap f lst
  #-}

overlaysMap :: (a -> Graph b) -> [a] -> Graph b
overlaysMap _ []     = empty
overlaysMap f [x]    = f x
overlaysMap f (x:xs) = f x `overlay` overlaysMap f xs
```
```
-- RHS size: {terms: 5, types: 7, coercions: 0, joins: 0/0}
vertices
vertices = \ @ a_a7O3 x_abEM -> overlaysMap Vertex x_abEM

Rec {
-- RHS size: {terms: 21, types: 23, coercions: 0, joins: 0/0}
overlaysMap
overlaysMap
  = \ @ a_a81u @ b_a81v ds_dbBU ds1_dbBV ->
      case ds1_dbBV of {
        [] -> Empty;
        : x_a67u ds2_dbC6 ->
          case ds2_dbC6 of wild1_XS {
            [] -> ds_dbBU x_a67u;
            : ipv_sbWJ ipv1_sbWK ->
              Overlay (ds_dbBU x_a67u) (overlaysMap ds_dbBU wild1_XS)
          }
      }
end Rec }
```

* With Direct implementation
```Haskell
vertices :: [a] -> Graph a
vertices []     = empty
vertices [x]    = vertex x
vertices (x:xs) = vertex x `overlay` vertices xs
```
```
Rec {
-- RHS size: {terms: 18, types: 19, coercions: 0, joins: 0/0}
vertices
vertices
  = \ @ a_a7LP ds_dbxs ->
      case ds_dbxs of {
        [] -> Empty;
        : x_a656 ds1_dbxC ->
          case ds1_dbxC of wild1_XQ {
            [] -> Vertex x_a656;
            : ipv_sbUd ipv1_sbUe -> Overlay (Vertex x_a656) (vertices wild1_XQ)
          }
      }
end Rec }
```

* So I ended with a new rewrite rule: (simply `overlaysMap` directly encoded)
```Haskell
{-# RULES
  "overlays/map" forall f lst.  overlays (map f lst) = fix (\ovMap lst' -> case lst' of
                                                                            []     -> empty
                                                                            [x]    -> f x
                                                                            (x:xs) -> f x `overlay` ovMap xs
                                                           ) lst
  #-}

```
```
Rec {
-- RHS size: {terms: 18, types: 19, coercions: 0, joins: 0/0}
biclique1
biclique1
  = \ @ a_a7NG lst'_a6fU ->
      case lst'_a6fU of {
        [] -> Empty;
        : x_a6fV ds_dbDG ->
          case ds_dbDG of wild1_X6 {
            [] -> Vertex x_a6fV;
            : ipv_sbWJ ipv1_sbWK ->
              Overlay (Vertex x_a6fV) (biclique1 wild1_X6)
          }
      }
end Rec }

-- RHS size: {terms: 4, types: 5, coercions: 0, joins: 0/0}
vertices
vertices = \ @ a_a7NG x_abEA -> biclique1 x_abEA
```

The compiler is naming the recursion function `biclique1` not sure why.........

Which perform between the two:
```
benchmarking RewriteSimple                    
time                 31.53 ms   (30.89 ms .. 32.54 ms)
                     0.998 R²   (0.996 R² .. 0.999 R²)
mean                 26.12 ms   (23.45 ms .. 27.57 ms)
std dev              3.728 ms   (2.009 ms .. 5.772 ms)
variance introduced by outliers: 55% (severely inflated)

benchmarking RewriteWithFix                     
time                 25.09 ms   (24.30 ms .. 26.17 ms)
                     0.996 R²   (0.993 R² .. 0.999 R²)
mean                 20.77 ms   (19.06 ms .. 21.82 ms)
std dev              2.696 ms   (1.563 ms .. 4.133 ms)
variance introduced by outliers: 54% (severely inflated)
                                      
benchmarking Direct                     
time                 21.47 ms   (20.91 ms .. 21.99 ms)
                     0.998 R²   (0.995 R² .. 0.999 R²)
mean                 17.41 ms   (15.84 ms .. 18.26 ms)
std dev              2.181 ms   (900.1 μs .. 3.078 ms)
variance introduced by outliers: 51% (severely inflated)
```

snowleopard:
@nobrakal Thanks! What about something like this instead?

```haskell
overlaysMap :: (a -> Graph b) -> [a] -> Graph b
overlaysMap _ []     = empty
overlaysMap f (x:xs) = foldl (\r -> overlay r . f) (f x) xs
```
This should capture `f` in a closure instead of passing it as a parameter.

nobrakal:
The code produced is almost similar (overlaysMap does not appear in the code of `vertices`, I believe GHC is optimizing the fold)
```
Rec {
-- RHS size: {terms: 14, types: 14, coercions: 0, joins: 0/0}
biclique1
biclique1
  = \ @ a_a7My ds_abQ6 eta_B1 ->
      case ds_abQ6 of {
        [] -> eta_B1;
        : y_abQb ys_abQc ->
          biclique1 ys_abQc (Overlay eta_B1 (Vertex y_abQb))
      }
end Rec }

-- RHS size: {terms: 11, types: 12, coercions: 0, joins: 0/0}
vertices
vertices
  = \ @ a_a7My x_abDi ->
      case x_abDi of {
        [] -> Empty;
        : x1_a65t xs_a65u -> biclique1 xs_a65u (Vertex x1_a65t)
      }
```
But for an ignored reason, it multiplies by ten the time took by the function by 10 :p
```
benchmarking REW                      
time                 315.0 ms   (280.0 ms .. 353.6 ms)
                     0.998 R²   (0.993 R² .. 1.000 R²)
mean                 304.0 ms   (298.9 ms .. 307.2 ms)
std dev              4.769 ms   (0.0 s .. 5.439 ms)
variance introduced by outliers: 19% (moderately inflated)
                                      
benchmarking DIR                      
time                 37.80 ms   (37.49 ms .. 38.13 ms)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 30.09 ms   (27.25 ms .. 31.99 ms)
std dev              3.982 ms   (2.583 ms .. 5.916 ms)
variance introduced by outliers: 52% (severely inflated)
```

snowleopard:
@nobrakal Aha, interesting -- what about using `foldl'` instead of `foldl`?

nobrakal:
I have tried, this is not a strictness problem (results are the same).
Don't really know where lies the issue in there

snowleopard:
OK, I'm out of ideas for now :) I leave this optimisation puzzle to you: how can we achieve the performance of the direct implementation using a more compositional approach where we don't duplicate a lot of code?

nobrakal:
@snowleopard Just re-tested, the rewrite rule I proposed 
```Haskell
{-# RULES
 "overlays/map" forall f lst.
    overlays (map f lst) = fix (\ovMap lst' ->
                                  case lst' of
                                    []     -> empty
                                    [x]    -> f x
                                    (x:xs) -> f x `overlay` ovMap xs
                               ) lst
#-}
```
```
benchmarking REW                      
time                 37.57 ms   (37.23 ms .. 37.89 ms)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 30.47 ms   (27.38 ms .. 32.05 ms)
std dev              3.617 ms   (2.221 ms .. 5.567 ms)
variance introduced by outliers: 44% (moderately inflated)
                                      
benchmarking DIR                      
time                 37.97 ms   (37.75 ms .. 38.20 ms)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 30.48 ms   (27.79 ms .. 32.33 ms)
std dev              3.828 ms   (2.421 ms .. 5.711 ms)
variance introduced by outliers: 45% (moderately inflated)
```
Is now showing me the same time as the direct implementation, so I think there was a bug in my yesterday results :p 
I think it is the best implementation, as we canno't modify function definition with RULES.

About duplicated code, this rule will produce a different function for *every* `x` in `overlays . map x`. Very like a custom implementation for each :)

I will update my branch soon with this improvements, adding its counterpart for `connect`

snowleopard:
@nobrakal Nice!

Still, I'm somewhat unsatisfied with this solution. Can you retry the more clear `overlayMap` variant -- perhaps it will now work too?

snowleopard:
Perhaps, instead of `NOINLINE` you need to use `INLINE[1]`?

snowleopard:
@nobrakal Thanks, I see.

Requiring `-O2` seems a bit unsatisfactory. It's very rarely used, and I'm worried that it will not give us predictable performance -- some further changes in GHC might result in these optimisations not happening. I'd say if it only works in `-O2` it means we just got lucky.

Let's try a couple of more `foldr`-based implementations:
* `overlays = foldr overlay empty` -- this one adds an unnecessary `+ empty` at the end, but is perhaps subject to built-in GHC optimisations for `foldr/map`.
* `overlays [] = empty; vertices (x:xs) = foldr overlay x xs` -- this will be close to your original suggestion with `foldr1`.

If nothing works, then perhaps we can stick to direct implementation for now, but open an issue where we could explore ways of doing this in a better way.

nobrakal:
I had left the `foldr` because it added an `Empty` graph, but it is in fact the best approach: GHC optimized it right away, and without flag or rewrite rules ^^:

```
benchmarking Foldr                      
time                 20.85 ms   (19.92 ms .. 21.49 ms)
                     0.990 R²   (0.967 R² .. 0.999 R²)
mean                 16.83 ms   (15.15 ms .. 18.01 ms)
std dev              2.722 ms   (1.379 ms .. 4.020 ms)
variance introduced by outliers: 63% (severely inflated)
                                      
benchmarking DIR                      
Progtime             22.47 ms   (22.11 ms .. 22.74 ms)
                     0.999 R²   (0.998 R² .. 1.000 R²)
mean                 17.27 ms   (15.76 ms .. 18.33 ms)
std dev              2.583 ms   (1.802 ms .. 3.621 ms)
variance introduced by outliers: 63% (severely inflated)
```

`overlays [] = empty; overlays (x:xs) = foldr overlay x xs` does not achieve these times unfortunately.

I have digged a bit what is doing GHC to optimize the simple `foldr` but I was not able to reproduce the same performance without the hidden `Empty`

snowleopard:
Good old `foldr` :-) Let's use it!

The only issue with it is that some expression size bounds may start to fail. For example, this one:

`size (removeEdge x y z)         <= 3 * size z` 

https://hackage.haskell.org/package/algebraic-graphs-0.1.1.1/docs/Algebra-Graph.html#v:removeEdge

But we can always bump the bound a bit, or do some custom optimisation.

nobrakal:
Just did it, and tests are failing as you said.

> or do some custom optimisation.

What do you have in mind ?

snowleopard:
> What do you have in mind ?

I've pushed one minor fix: 

https://github.com/snowleopard/alga/commit/480a73137e9b38ad3f1bc2c628847953d2fb3e25

However, this may be not enough :)

nobrakal:
> However, this may be not enough :)

Tests are passing :)

snowleopard:
Cool :) Please switch other implementations of `overlays` to using `foldr`.

src/Algebra/Graph/AdjacencyMap.hs
```diff
@@ -104,7 +104,7 @@ edges = fromAdjacencySets . map (fmap Set.singleton)
 -- 'isEmpty' . overlays == 'all' 'isEmpty'
 -- @
 overlays :: Ord a => [AdjacencyMap a] -> AdjacencyMap a
-overlays = AM . Map.unionsWith Set.union . map adjacencyMap
+overlays = foldr overlay empty
```
snowleopard: Wait, not here :) Adjacency maps and relations have a more efficient implementation!

src/Algebra/Graph/AdjacencyMap.hs
```diff
@@ -117,9 +117,7 @@ overlays = AM . Map.unionsWith Set.union . map adjacencyMap
 -- 'isEmpty' . connects == 'all' 'isEmpty'
 -- @
 connects :: Ord a => [AdjacencyMap a] -> AdjacencyMap a
-connects []     = empty
-connects [x]    = x
-connects (x:xs) = x `connect` connects xs
+connects = foldr connect empty
```
snowleopard: Here is fine, since there is nothing more clever.

snowleopard: Although in future we might add a more efficient `connects` for adjacency maps too.

src/Algebra/Graph/AdjacencyMap.hs
nobrakal: I had some doubt, I am going to revert !

snowleopard:
@nobrakal Is there anything else you plan to do in this PR? Looks like I can't merge it due to conflicts.

nobrakal:
I did not plan to add anything, you can merge :)

snowleopard:
Done!

PR #74, opened 2018-06-08
Drop dependency of Algebra.Graph on graph type classes

snowleopard:
This is work in progress. Some further changes are possible, e.g. removing all inefficient (based on generic implementations in type classes) functions from the API of `Relation`, `AdjacencyMap`, etc.

In general, this is a move towards simplifying the library and focusing on `Algerba.Graph` as the main data type for algerbaric graph manipulation.

This also addresses some performance issues raised by @nobrakal.

@anfelor Do you have any feedback?


nobrakal:
This is I think a good improvement :D That will speedup many functions from `Algebra.Graph`.

PR #79, opened 2018-06-12
Rename toAdjacencyMap to adjacencyMap

snowleopard:
Trying to come up with a more consistent naming for properties.


snowleopard:
@nobrakal I've done quite a major renaming here, which will probably require you to do some changes in the benchmark suite -- apologies about this. 

But I really think the new naming is more consistent. Previously we had `adjacencyMap` with two different types: one returning `Map a (Set a)`, another returning `IntMap IntSet`. Now we have `adjacencyMap` and `adjacencyIntMap` -- just like `vertexSet` and `vertexIntSet`. And with this change it also makes sense to rename `IntAdjacencyMap` to `AdjacencyIntMap`.

nobrakal:
@snowleopard Don't worry, I am not using these functions directly.

The naming is great !

PR #81, opened 2018-06-13
Specialise overloaded functions to Graph Int

nobrakal:
Here the code for #75 

I can add some benchs if you want :)

### DIFF
--- a/src/Algebra/Graph.hs
+++ b/src/Algebra/Graph.hs
@@ -395,6 +395,8 @@ foldg e v o c = go
 isSubgraphOf :: Ord a => Graph a -> Graph a -> Bool
 isSubgraphOf x y = overlay x y == y
 
+{-# SPECIALISE isSubgraphOf :: Graph Int -> Graph Int -> Bool #-}
+
 -- | Structural equality on graph expressions.
 -- Complexity: /O(s)/ time.
 --
@@ -413,6 +415,8 @@ Empty           === Empty           = True
 _               === _               = False
 
 infix 4 ===
+
+{-# SPECIALISE (===) :: Graph Int -> Graph Int -> Bool #-}
 
 -- | Check if a graph is empty. A convenient alias for 'null'.
 -- Complexity: /O(s)/ time.
@@ -454,6 +458,8 @@ size = foldg 1 (const 1) (+) (+)
 hasVertex :: Eq a => a -> Graph a -> Bool
 hasVertex x = foldg False (==x) (||) (||)
 
+{-# SPECIALISE hasVertex :: Int -> Graph Int -> Bool #-}
+
 -- TODO: Benchmark to see if this implementation is faster than the default
 -- implementation provided by the ToGraph type class.
 -- | Check if a graph contains a given edge.
@@ -468,6 +474,8 @@ hasVertex x = foldg False (==x) (||) (||)
 -- @
 hasEdge :: Ord a => a -> a -> Graph a -> Bool
 hasEdge u v = (edge u v `isSubgraphOf`) . induce (`elem` [u, v])
+
+{-# SPECIALISE hasEdge :: Int -> Int -> Graph Int -> Bool #-}
 
 -- | The number of vertices in a graph.
 -- Complexity: /O(s * log(n))/ time.
@@ -499,6 +507,8 @@ vertexIntCount = IntSet.size . vertexIntSet
 -- @
 edgeCount :: Ord a => Graph a -> Int
 edgeCount = length . edgeList
+
+{-# SPECIALISE edgeCount :: Graph Int -> Int #-}
 
 -- | The sorted list of vertices of a given graph.
 -- Complexity: /O(s * log(n))/ time and /O(n)/ memory.
@@ -532,6 +542,8 @@ vertexIntList = IntSet.toList . vertexIntSet
 -- @
 edgeList :: Ord a => Graph a -> [(a, a)]
 edgeList = AM.edgeList . toAdjacencyMap
+
+{-# SPECIALISE edgeList :: Graph Int -> [(Int,Int)] #-}
 
 -- | The set of vertices of a given graph.
 -- Complexity: /O(s * log(n))/ time and /O(n)/ memory.
@@ -570,6 +582,8 @@ vertexIntSet = foldg IntSet.empty IntSet.singleton IntSet.union IntSet.union
 edgeSet :: Ord a => Graph a -> Set.Set (a, a)
 edgeSet = AM.edgeSet . toAdjacencyMap
 
+{-# SPECIALISE edgeSet :: Graph Int -> Set.Set (Int,Int) #-}
+
 -- | The sorted /adjacency list/ of a graph.
 -- Complexity: /O(n + m)/ time and /O(m)/ memory.
 --
@@ -582,6 +596,8 @@ edgeSet = AM.edgeSet . toAdjacencyMap
 -- @
 adjacencyList :: Ord a => Graph a -> [(a, [a])]
 adjacencyList = AM.adjacencyList . toAdjacencyMap
+
+{-# SPECIALISE adjacencyList :: Graph Int -> [(Int,[Int])] #-}
 
 -- | The /path/ on a list of vertices.
 -- Complexity: /O(L)/ time, memory and size, where /L/ is the length of the
@@ -767,6 +783,8 @@ deBruijn len alphabet = skeleton >>= expand
 removeVertex :: Eq a => a -> Graph a -> Graph a
 removeVertex v = induce (/= v)
 
+{-# SPECIALISE removeVertex :: Int -> Graph Int -> Graph Int #-}
+
 -- | Remove an edge from a given graph.
 -- Complexity: /O(s)/ time, memory and size.
 --
@@ -781,6 +799,8 @@ removeVertex v = induce (/= v)
 removeEdge :: Eq a => a -> a -> Graph a -> Graph a
 removeEdge s t = filterContext s (/=s) (/=t)
 
+{-# SPECIALISE removeEdge :: Int -> Int -> Graph Int -> Graph Int #-}
+
 -- TODO: Export
 -- | Filter vertices in a subgraph context.
 filterContext :: Eq a => a -> (a -> Bool) -> (a -> Bool) -> Graph a -> Graph a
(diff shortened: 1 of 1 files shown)


snowleopard:
Thanks! Yes, it would be nice to see some benchmarks. I'm not sure if it takes a lot of time for you -- ideally we should automate this somehow, creating a regression suite to see performance effects of all PRs. Do you think you could implement something like this?

src/Algebra/Graph.hs
```diff
@@ -395,6 +395,8 @@ foldg e v o c = go
 isSubgraphOf :: Ord a => Graph a -> Graph a -> Bool
 isSubgraphOf x y = overlay x y == y
 
+{-# SPECIALISE isSubgraphOf :: Graph Int -> Graph Int -> Bool #-}
```
snowleopard: Can you skip the empty line above? It's not immediately obvious to which function the specialisations apply.

nobrakal: Of course, I thought that rendered not so well, but you are right

nobrakal: In fact, because it looks like a type definition, maybe adding it just below the actual type definition is a good idea

snowleopard: Another option is to group all these pragmas together, say at the bottom of the file with a comment explaining what it achieves.

nobrakal:
Mmhh that can certainly be done. I did not think about it. A bash-script with ~ 10 lines can do the trick

snowleopard:
> A bash-script with ~ 10 lines can do the trick

Yep, and we should call this script from our CI scripts.

src/Algebra/Graph.hs
nobrakal: Yes, but one can forget to check this. I like that when I am reading the code and I have everything under my eyes

snowleopard: OK, let's place it next to function definitions. Perhaps, below the comment but on top of function type signature? Unless this messes up Haddock generation.

nobrakal: Ok I will check this

nobrakal:
Hop, a script is living here: https://github.com/nobrakal/benchAlgaPr

It will bench a directory containing a different version of alga again `alga/master/HEAD`. Fore example here the content of `res.md` on this PR, using
```Bash
./compare.sh ../alga res.md hasVertex removeVertex hasEdge removeEdge
```


Using [("Mesh",3),("Clique",3)] as graphs

## hasVertex

Descritpion: Test if the given vertex is in the graph

### Mesh
<TABLE>
   <TR>
      <TH>
      </TH>
      <TH CLASS = "thinright">
         1
      </TH>
      <TH CLASS = "thinright">
         10
      </TH>
      <TH>
         100
      </TH>
   </TR>
   <TR>
      <TH>
         Alga
      </TH>
      <TD CLASS = "thinright">
         24.75 ns
      </TD>
      <TD CLASS = "thinright">
         81.59 ns
      </TD>
      <TD>
         81.29 ns
      </TD>
   </TR>
   <TR>
      <TH>
         AlgaOld
      </TH>
      <TD CLASS = "thinright">
         31.46 ns
      </TD>
      <TD CLASS = "thinright">
         114.3 ns
      </TD>
      <TD>
         112.3 ns
      </TD>
   </TR>
</TABLE>

### Clique
<TABLE>
   <TR>
      <TH>
      </TH>
      <TH CLASS = "thinright">
         1
      </TH>
      <TH CLASS = "thinright">
         10
      </TH>
      <TH>
         100
      </TH>
   </TR>
   <TR>
      <TH>
         Alga
      </TH>
      <TD CLASS = "thinright">
         25.27 ns
      </TD>
      <TD CLASS = "thinright">
         396.2 ns
      </TD>
      <TD>
         37.96 μs
      </TD>
   </TR>
   <TR>
      <TH>
         AlgaOld
      </TH>
      <TD CLASS = "thinright">
         31.52 ns
      </TD>
      <TD CLASS = "thinright">
         590.3 ns
      </TD>
      <TD>
         55.66 μs
      </TD>
   </TR>
</TABLE>


SUMMARY:

 * Alga was the fastest 16 times


ABSTRACT:

 * Alga was 1.36 times faster than AlgaOld

## hasEdge

Descritpion: Test if the given edge is in the graph (with arguments both in the graph and not in the graph (where applicable))

### Mesh
<TABLE>
   <TR>
      <TH>
      </TH>
      <TH CLASS = "thinright">
         1
      </TH>
      <TH CLASS = "thinright">
         10
      </TH>
      <TH>
         100
      </TH>
   </TR>
   <TR>
      <TH>
         Alga
      </TH>
      <TD CLASS = "thinright">
         343.6 ns
      </TD>
      <TD CLASS = "thinright">
         1.332 μs
      </TD>
      <TD>
         1.339 μs
      </TD>
   </TR>
   <TR>
      <TH>
         AlgaOld
      </TH>
      <TD CLASS = "thinright">
         552.4 ns
      </TD>
      <TD CLASS = "thinright">
         2.292 μs
      </TD>
      <TD>
         2.280 μs
      </TD>
   </TR>
</TABLE>

### Clique
<TABLE>
   <TR>
      <TH>
      </TH>
      <TH CLASS = "thinright">
         1
      </TH>
      <TH CLASS = "thinright">
         10
      </TH>
      <TH>
         100
      </TH>
   </TR>
   <TR>
      <TH>
         Alga
      </TH>
      <TD CLASS = "thinright">
         345.2 ns
      </TD>
      <TD CLASS = "thinright">
         6.372 μs
      </TD>
      <TD>
         573.9 μs
      </TD>
   </TR>
   <TR>
      <TH>
         AlgaOld
      </TH>
      <TD CLASS = "thinright">
         543.3 ns
      </TD>
      <TD CLASS = "thinright">
         7.943 μs
      </TD>
      <TD>
         637.9 μs
      </TD>
   </TR>
</TABLE>


SUMMARY:

 * Alga was the fastest 18 times


ABSTRACT:

 * Alga was 1.42 times faster than AlgaOld

## removeVertex

Descritpion: Remove a vertex of the graph

### Mesh
<TABLE>
   <TR>
      <TH>
      </TH>
      <TH CLASS = "thinright">
         1
      </TH>
      <TH CLASS = "thinright">
         10
      </TH>
      <TH>
         100
      </TH>
   </TR>
   <TR>
      <TH>
         Alga
      </TH>
      <TD CLASS = "thinright">
         36.52 ns
      </TD>
      <TD CLASS = "thinright">
         339.7 ns
      </TD>
      <TD>
         325.5 ns
      </TD>
   </TR>
   <TR>
      <TH>
         AlgaOld
      </TH>
      <TD CLASS = "thinright">
         35.05 ns
      </TD>
      <TD CLASS = "thinright">
         327.6 ns
      </TD>
      <TD>
         337.4 ns
      </TD>
   </TR>
</TABLE>

### Clique
<TABLE>
   <TR>
      <TH>
      </TH>
      <TH CLASS = "thinright">
         1
      </TH>
      <TH CLASS = "thinright">
         10
      </TH>
      <TH>
         100
      </TH>
   </TR>
   <TR>
      <TH>
         Alga
      </TH>
      <TD CLASS = "thinright">
         36.51 ns
      </TD>
      <TD CLASS = "thinright">
         3.067 μs
      </TD>
      <TD>
         440.7 μs
      </TD>
   </TR>
   <TR>
      <TH>
         AlgaOld
      </TH>
      <TD CLASS = "thinright">
         35.28 ns
      </TD>
      <TD CLASS = "thinright">
         3.007 μs
      </TD>
      <TD>
         422.8 μs
      </TD>
   </TR>
</TABLE>


SUMMARY:

There was 6 ex-aequo


ABSTRACT:

 * AlgaOld was 1.00 times faster than Alga

## removeEdge

Descritpion: Remove an edge of the graph

### Mesh
<TABLE>
   <TR>
      <TH>
      </TH>
      <TH CLASS = "thinright">
         10
      </TH>
      <TH>
         100
      </TH>
   </TR>
   <TR>
      <TH>
         Alga
      </TH>
      <TD CLASS = "thinright">
         2.001 μs
      </TD>
      <TD>
         2.008 μs
      </TD>
   </TR>
   <TR>
      <TH>
         AlgaOld
      </TH>
      <TD CLASS = "thinright">
         2.204 μs
      </TD>
      <TD>
         2.211 μs
      </TD>
   </TR>
</TABLE>

### Clique
<TABLE>
   <TR>
      <TH>
      </TH>
      <TH CLASS = "thinright">
         10
      </TH>
      <TH>
         100
      </TH>
   </TR>
   <TR>
      <TH>
         Alga
      </TH>
      <TD CLASS = "thinright">
         15.67 μs
      </TD>
      <TD>
         7.655 ms
      </TD>
   </TR>
   <TR>
      <TH>
         AlgaOld
      </TH>
      <TD CLASS = "thinright">
         17.04 μs
      </TD>
      <TD>
         7.907 ms
      </TD>
   </TR>
</TABLE>


SUMMARY:

 * Alga was the fastest 4 times
There was 8 ex-aequo


ABSTRACT:

 * Alga was 1.07 times faster than AlgaOld


The specialization seems to be useful only on graph inspection...

snowleopard:
@nobrakal Great! Could you send a separate PR to invoke this script from our CI scripts?

nobrakal:
For CI I don't know if this is a good idea.

It's output is not very good to work with, and because I cannot say which benchmark is good to do, it will run the whole bench-suite, producing many unrelevant output.

snowleopard:
> It's output is not very good to work with

Yes, right now it's difficult to view without rendering in a Markdown viewer, but still it's better than nothing. As a maintainer of Alga I will be able to review the performance effect of any PR, which is very important. Otherwise, it's really easy to accidentally introduce a performance regression.

Perhaps, in future we can add an option to render the output in a more terminal friendly way. It's probably not even too difficult, we just need to redirect output to a Markdown file and then run an existing text-only renderer for Markdown files (right in the CI script).

> I cannot say which benchmark is good to do, it will run the whole bench-suite

Yes, that's the only way. It's the same with the testsuite really -- we always run all tests even if we change a comment, because there is no easy way to do better.

nobrakal:
> Perhaps, in future we can add an option to render the output in a more terminal friendly way. 

It can even be done directly in the bench suite, only require some tweaks

> As a maintainer of Alga I will be able to review the performance effect of any PR, which is very important. 

Ok, so I will send a PR. Since it is only for a quick review, do you only want to "Best,Abstract" part, and not the detail ? It will be simpler to check I think

snowleopard:
@nobrakal Thanks! 

I've merged this PR.

PR #82, opened 2018-06-15
Bench on commit

nobrakal:
Hi,

As you requested, here is the .travis.yml updated to bench alga against an older version of itself.

It basically calls a script I made here: https://github.com/nobrakal/benchAlgaPr/blob/master/compare.sh

It is (very) tricky, here is some explanation:

This script will clone alga, and reset it to `HEAD^1" if it is a direct push, or leave it as "HEAD" if it is a PR (so PR will be benchmarked against HEAD, and direct push against HEAD^1). (The script itself can be parametrized to use an other commit id)

Then I rename the algebraic-graph package to "old", renaming the "Algebra.Graph" module to "Algebra.GraphOld".

Then it will clone and modify https://github.com/haskell-perf/graphs/ to match this.

And then it will run the benchmarks.
I disabled the output, so only the ABSTRACT and SUMMARY sections are displayed

Please note that the script can both using `cabal new-*` and `stack`, and can be forced to bench only some functions.

The output can be seen here: https://travis-ci.com/nobrakal/alga/jobs/129609679

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -114,7 +114,7 @@ script:
 
   # Benchmark
   - cd $TRAVIS_BUILD_DIR
-  - if [ $DOBENCH ]; then ./compare.sh Cabal ./ Ascii $(if [ $TRAVIS_PULL_REQUEST = "false" ]; then "HEAD^1" else "HEAD" ); fi;
+  - if [ $DOBENCH ]; then ./compare.sh Cabal ./ Ascii $(if [ $TRAVIS_PULL_REQUEST = "false" ]; then "HEAD^1"; else "HEAD";fi; ); fi;
 
 # REGENDATA ["--osx=8.2.2","algebraic-graphs.cabal","-o",".travis.yml"]
 # EOF
--- a/.travis.yml
+++ b/.travis.yml
@@ -114,7 +114,7 @@ script:
 
   # Benchmark
   - cd $TRAVIS_BUILD_DIR
-  - if [ $DOBENCH ]; then ./compare.sh Cabal ./ Ascii $(if [ $TRAVIS_PULL_REQUEST = "false" ]; then "HEAD^1"; else "HEAD";fi; ); fi;
+  - if [ $DOBENCH ]; then ./compare.sh Cabal ./ Ascii $(if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then echo "HEAD^1"; else echo "HEAD";fi; ); fi;
 
 # REGENDATA ["--osx=8.2.2","algebraic-graphs.cabal","-o",".travis.yml"]
 # EOF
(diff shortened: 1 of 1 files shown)


snowleopard:
This is awesome!

What about printing the results in a much more compact form, e.g.:
```
isEmpty: -5%
vertexList: +1% 
...
```

Here the numbers correspond to the time increase/decrease, that is, `-5%` means the new implementation is 5% faster, and `+1%` means it is 1% slower.

Then it's very easy to evaluate the impact of the commit/PR.

nobrakal:
This is now implemented: https://travis-ci.com/nobrakal/alga/jobs/129716066

I have keep the ratio format, it represent the `Alga/AlgaOld` ratio. So under 1, this is an improvement, above it is a problem.

Please note that results are to be analyzed, in the results I sent you, `vertexCount` is `0.85` times slower, but it is the same code behind (my PR don't change the code). (I believe this come from Criterion OLS regression)

It will not replace an analyzed table of benchmarks on larger graphs (The benchs are done with 10^4 vertices ONLY), so the impact of a commit on smaller graph is not evaluated.

snowleopard:
@nobrakal Many thanks, this looks great!

Can you drop the unnecessary `Doing: ...` list?

I think it's normal to see some variation in performance.

snowleopard:
@nobrakal Thanks! One more request: at some we might need to regenerate the Travis script, and it should be clear what modifications should be made to the generated script. At the moment, the top comment says: "This Travis job script has been generated by a script via", but this is not true -- a part of it you wrote by hand. Could you expand the top comment by precisely specifying which new lines were added to the script?

nobrakal:
Just added some comments, does they fits you

snowleopard:
@nobrakal Great, thank you, merged!

One thing that bothers me is that performance variance is higher than I expected, which makes it difficult to take quick decisions. For example, in the latest commit here we see `hasVertex: 1.37`. Is there anything we can do to bring the variance down? Perhaps, we can give Criterion more time for benchmarking?

nobrakal:
Just spent some time on this issue, giving more time to criterion don't seems to ameliorate things (the base time is 5s, with 10s, it keeps throwing strange number). 

Maybe doing more tests can be a good idea, hoping that without sense values are diminished thanks to the average.

nobrakal:
Just did it, results seems a bit better https://travis-ci.com/nobrakal/alga/jobs/129907802#L3081

snowleopard:
@nobrakal Thanks! Looks better. As far as I understand, this will take an immediate effect, i.e. we don't need to change anything in Alga.

nobrakal:
Exactly, this is only an improvement inside the haskell-perf/graph repositiory

nobrakal:
For information, I had some problems with travis (it killed the job because the benchs used too much memory), so I lowered the graphs used and came back to 10^3 vertices. Should enough to show if there is a real drop in performance. I prefer to do more little benchs than only one or two massive ones.

snowleopard:
@nobrakal Thanks, I'm sure it will be very useful, even if there is a lot of variance in results.

Issue #84, opened 2018-06-21
Benchmark hasEdge


nobrakal:
Hi,

This is just for the record:

I saw: https://github.com/snowleopard/alga/blob/master/src/Algebra/Graph.hs#L449 : 

> -- TODO: Benchmark to see if this implementation is faster than the default
-- implementation provided by the ToGraph type class.

So here is the benchmarks:

`Alga` use the the ToGraph definition, `AlgaOld` is the current implementation:

## hasEdge

Description: Test if the given edge is in the graph (with arguments both in the graph and not in the graph (where applicable))

### Mesh
<TABLE>
   <TR>
      <TH>
      </TH>
      <TH CLASS = "thinright">
         1
      </TH>
      <TH CLASS = "thinright">
         10
      </TH>
      <TH CLASS = "thinright">
         100
      </TH>
      <TH>
         1000
      </TH>
   </TR>
   <TR>
      <TH>
         Alga
      </TH>
      <TD CLASS = "thinright">
         30.38 ns
      </TD>
      <TD CLASS = "thinright">
         783.9 ns
      </TD>
      <TD CLASS = "thinright">
         11.84 μs
      </TD>
      <TD>
         270.4 μs
      </TD>
   </TR>
   <TR>
      <TH>
         AlgaOld
      </TH>
      <TD CLASS = "thinright">
         385.7 ns
      </TD>
      <TD CLASS = "thinright">
         2.330 μs
      </TD>
      <TD CLASS = "thinright">
         19.89 μs
      </TD>
      <TD>
         219.1 μs
      </TD>
   </TR>
</TABLE>

### Clique
<TABLE>
   <TR>
      <TH>
      </TH>
      <TH CLASS = "thinright">
         1
      </TH>
      <TH CLASS = "thinright">
         10
      </TH>
      <TH CLASS = "thinright">
         100
      </TH>
      <TH>
         1000
      </TH>
   </TR>
   <TR>
      <TH>
         Alga
      </TH>
      <TD CLASS = "thinright">
         30.28 ns
      </TD>
      <TD CLASS = "thinright">
         2.733 μs
      </TD>
      <TD CLASS = "thinright">
         1.758 ms
      </TD>
      <TD>
         870.8 ms
      </TD>
   </TR>
   <TR>
      <TH>
         AlgaOld
      </TH>
      <TD CLASS = "thinright">
         384.7 ns
      </TD>
      <TD CLASS = "thinright">
         6.709 μs
      </TD>
[...]

PR #86, opened 2018-06-28
Has loop

nobrakal:
This is related to #84 

Here is the `hasLoop` implemented, with default implementation in `ToGraph` and a better one for `Algebra.Graph` and `Algebra.Graph.NonEmpty`.

I have also sneaked a little improvement in `Algebra.Graph.NonEmpty` concerning `foldg1` which `Maybe` directly. A bench of `removeVertex` for this improvement:

### Mesh
<TABLE>
   <TR>
      <TH>
      </TH>
      <TH CLASS = "thinright">
         1
      </TH>
      <TH CLASS = "thinright">
         10
      </TH>
      <TH>
         100
      </TH>
   </TR>
   <TR>
      <TH>
         NonEmpty
      </TH>
      <TD CLASS = "thinright">
         34.78 ns
      </TD>
      <TD CLASS = "thinright">
         1.080 μs
      </TD>
      <TD>
         14.85 μs
      </TD>
   </TR>
   <TR>
      <TH>
         NonEmptyOld
      </TH>
      <TD CLASS = "thinright">
         44.27 ns
      </TD>
      <TD CLASS = "thinright">
         1.853 μs
      </TD>
      <TD>
         27.71 μs
      </TD>
   </TR>
</TABLE>

### Clique
<TABLE>
   <TR>
      <TH>
      </TH>
      <TH CLASS = "thinright">
         1
      </TH>
      <TH CLASS = "thinright">
         10
      </TH>
      <TH>
         100
      </TH>
   </TR>
   <TR>
      <TH>
         NonEmpty
      </TH>
      <TD CLASS = "thinright">
         34.70 ns
      </TD>
      <TD CLASS = "thinright">
         3.589 μs
      </TD>
      <TD>
         520.5 μs
      </TD>
   </TR>
   <TR>
      <TH>
         NonEmptyOld
      </TH>
      <TD CLASS = "thinright">
         46.32 ns
      </TD>
      <TD CLASS = "thinright">
         6.523 μs
      </TD>
      <TD>
         1.404 ms
      </TD>
   </TR>
</TABLE>

SUMMARY:

 * NonEmpty was the fastest 12 times

ABSTRACT:

 * NonEmpty was 2.21 times faster than NonEmptyOld

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,11 +1,8 @@
 # Change log
 
-## 0.2.1
-
-* #86: Add `hasSelfLoop` into the API.
-
 ## 0.2
 
+* #86: Add `hasSelfLoop` into the API.
 * #79: Improve the API consistency: rename `IntAdjacencyMap` to `AdjacencyIntMap`,
        and then rename the function that extracts its adjacency map to
        `adjacencyIntMap` to avoid the clash with `AdjacencyMap.adjacencyMap`,
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # Change log
 
+## 0.2.1
+
+* #86: Add `hasSelfLoop` into the API.
+
 ## 0.2
 
 * #79: Improve the API consistency: rename `IntAdjacencyMap` to `AdjacencyIntMap`,
--- a/src/Algebra/Graph.hs
+++ b/src/Algebra/Graph.hs
@@ -471,7 +471,7 @@ hasEdge u v = (edge u v `isSubgraphOf`) . induce (`elem` [u, v])
 -- hasLoop x ('edge' x x)       == True
 -- hasLoop x                    == hasEdge x x
 -- hasLoop x . 'removeEdge' x x == const False
--- hasEdge x                    == 'elem' (x,x) . 'edgeList'
+-- hasLoop x                    == 'elem' (x,x) . 'edgeList'
 -- @
 {-# SPECIALISE hasLoop :: Int -> Graph Int -> Bool #-}
 hasLoop :: Eq a => a -> Graph a -> Bool
--- a/src/Algebra/Graph.hs
+++ b/src/Algebra/Graph.hs
@@ -33,7 +33,7 @@ module Algebra.Graph (
     isSubgraphOf, (===),
 
     -- * Graph properties
-    isEmpty, size, hasVertex, hasEdge, hasLoop, vertexCount, edgeCount,
+    isEmpty, size, hasVertex, hasEdge, hasSelfLoop, vertexCount, edgeCount,
     vertexList, edgeList, vertexSet, vertexIntSet, edgeSet, adjacencyList,
 
     -- * Standard families of graphs
@@ -466,20 +466,20 @@ hasEdge u v = (edge u v `isSubgraphOf`) . induce (`elem` [u, v])
 -- Complexity: /O(s)/ time.
 --
 -- @
--- hasLoop x 'empty'            == False
--- hasLoop x ('vertex' z)       == False
--- hasLoop x ('edge' x x)       == True
--- hasLoop x                    == hasEdge x x
--- hasLoop x . 'removeEdge' x x == const False
--- hasLoop x                    == 'elem' (x,x) . 'edgeList'
+-- hasSelfLoop x 'empty'            == False
+-- hasSelfLoop x ('vertex' z)       == False
+-- hasSelfLoop x ('edge' x x)       == True
+-- hasSelfLoop x                    == hasEdge x x
+-- hasSelfLoop x . 'removeEdge' x x == const False
+-- hasSelfLoop x                    == 'elem' (x,x) . 'edgeList'
 -- @
-{-# SPECIALISE hasLoop :: Int -> Graph Int -> Bool #-}
-hasLoop :: Eq a => a -> Graph a -> Bool
-hasLoop l = hasLoop' . induce (==l)
-  where -- hasLoop' is working because Algebra.Graph.induce is removing empty leaves.
-    hasLoop' (Overlay x y) = hasLoop' x || hasLoop' y
-    hasLoop' Connect{} = True
-    hasLoop' _ = False
+{-# SPECIALISE hasSelfLoop :: Int -> Graph Int -> Bool #-}
+hasSelfLoop :: Eq a => a -> Graph a -> Bool
+hasSelfLoop l = hasSelfLoop' . induce (==l)
+  where -- hasSelfLoop' is working because Algebra.Graph.induce is removing empty leaves.
+    hasSelfLoop' (Overlay x y) = hasSelfLoop' x || hasSelfLoop' y
+    hasSelfLoop' Connect{} = True
+    hasSelfLoop' _ = False
 
 -- | The number of vertices in a graph.
 -- Complexity: /O(s * log(n))/ time.
--- a/src/Algebra/Graph.hs
+++ b/src/Algebra/Graph.hs
@@ -33,8 +33,8 @@ module Algebra.Graph (
     isSubgraphOf, (===),
 
     -- * Graph properties
-    isEmpty, size, hasVertex, hasEdge, vertexCount, edgeCount, vertexList,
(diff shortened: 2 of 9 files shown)


src/Algebra/Graph/NonEmpty.hs
```diff
@@ -681,7 +699,14 @@ transpose = foldg1 vertex overlay (flip connect)
 -- induce1 p '>=>' induce1 q == induce1 (\\x -> p x && q x)
 -- @
 induce1 :: (a -> Bool) -> NonEmptyGraph a -> Maybe (NonEmptyGraph a)
-induce1 p = toNonEmptyGraph . G.induce p . T.toGraph
+induce1 p = foldg1
+  (\x -> if p x then Just (Vertex x) else Nothing)
+  (k Overlay)
```
snowleopard: Isn't this just `liftA2 Overlay` and `liftA2 Connect`?

src/Algebra/Graph/NonEmpty.hs
```diff
@@ -377,9 +377,27 @@ hasVertex v = foldg1 (==v) (||) (||)
...
-hasEdge :: Ord a => a -> a -> NonEmptyGraph a -> Bool
+hasEdge :: Eq a => a -> a -> NonEmptyGraph a -> Bool
 hasEdge = T.hasEdge
 
+-- | Check if a graph contains a given loop.
+-- Complexity: /O(s)/ time.
+--
+-- @
+-- hasLoop x 'empty'            == False
+-- hasLoop x ('vertex' z)       == False
```
snowleopard: Why `z` instead of `y`? Seems surprising :)

src/Algebra/Graph/NonEmpty.hs
```diff
@@ -377,9 +377,27 @@ hasVertex v = foldg1 (==v) (||) (||)
...
+-- hasLoop x ('edge' x x)       == True
+-- hasLoop x                    == hasEdge x x
+-- hasLoop x . 'removeEdge' x x == const False
+-- hasEdge x                    == 'elem' (x,x) . 'edgeList'
```
snowleopard: Should be `hasLoop` instead of `hasEdge`. In fact, I would probably drop this example -- the ones above are sufficient.

src/Algebra/Graph/NonEmpty.hs
```diff
@@ -377,9 +377,27 @@ hasVertex v = foldg1 (==v) (||) (||)
...
+-- @
+hasLoop :: Eq a => a -> NonEmptyGraph a -> Bool
+hasLoop l = maybe False hasLoop' . induce1 (==l)
+  where -- hasLoop' is working because induce is removing empty leaves.
```
snowleopard: The comment is misleading -- there no empty leaves in non-empty graphs.

src/Algebra/Graph/ToGraph.hs
```diff
@@ -100,6 +100,14 @@ class ToGraph t where
         o (xs, xt, xst) (ys, yt, yst) = (xs || ys, xt || yt,             xst || yst)
         c (xs, xt, xst) (ys, yt, yst) = (xs || ys, xt || yt, xs && yt || xst || yst)
 
+    -- | Check if a graph contains a given lopp.
+    --
+    -- @
+    -- hasEdge x == hasEdge x x
```
snowleopard: Again, `hasLoop x = ...`.

snowleopard:
Many thanks @nobrakal! I've added a bunch of comments.

As a general comment, I'm tempted to use `hasSelfLoop` instead of `hasLoop`, even though it's longer, because it is more self-!-explanatory. What do you think?

src/Algebra/Graph/NonEmpty.hs
nobrakal: Totally, copying/pasting is a bad thing :)

src/Algebra/Graph/NonEmpty.hs
nobrakal: Again, copy/paste and some suppression of `y` from `hasEdge` ^^ corrected

src/Algebra/Graph/NonEmpty.hs
nobrakal: Well, no it is not, because `liftA2 (+) (Nothing) (Just 1) == Nothing`, we don't want this behavior

snowleopard: Oops, you are right :)

src/Algebra/Graph.hs
```diff
@@ -462,6 +462,25 @@ hasVertex x = foldg False (==x) (||) (||)
...
 hasEdge u v = (edge u v `isSubgraphOf`) . induce (`elem` [u, v])
 
+-- | Check if a graph contains a given loop.
+-- Complexity: /O(s)/ time.
+--
+-- @
+-- hasSelfLoop x 'empty'            == False
+-- hasSelfLoop x ('vertex' z)       == False
+-- hasSelfLoop x ('edge' x x)       == True
+-- hasSelfLoop x                    == hasEdge x x
```
snowleopard: I think these examples do not line up in Haddock. Could you please check? You might need to insert a few spaces here and there.

src/Algebra/Graph/ToGraph.hs
```diff
@@ -100,6 +100,14 @@ class ToGraph t where
         o (xs, xt, xst) (ys, yt, yst) = (xs || ys, xt || yt,             xst || yst)
         c (xs, xt, xst) (ys, yt, yst) = (xs || ys, xt || yt, xs && yt || xst || yst)
 
+    -- | Check if a graph contains a given lopp.
+    --
+    -- @
+    -- hasSelfLoop x == hasSelfLoop x x
```
snowleopard: `hasSelfLoop x x` should be `hasEdge x x` :)

test/Algebra/Graph/Test/Generic.hs
```diff
@@ -522,6 +523,25 @@ testHasEdge (Testsuite prefix (%)) = do
...
+    test "hasSelfLoop x (vertex y)       == False" $ \x y ->
+          hasSelfLoop x % vertex y       == False
+
+    test "hasSelfLoop x (edge x x)       == True" $ \x ->
+          hasSelfLoop x % edge x x       == True
+
+    test "hasSelfLoop x . removeEdge x x == const False" $ \x y ->
+         (hasSelfLoop x . removeEdge x x) y == const False % y
+
+    test "hasSelfLoop x                  == elem (x,x) . edgeList" $ \x y -> do
```
snowleopard: I think we should replace this test with `hasSelfLoop x == hasEge x x`, which is in the Haddock.

snowleopard:
@nobrakal Thanks! A few more minor tweaks and we can merge.

nobrakal:
Thanks for the review, I have addressed all your comments :)

snowleopard:
Wait, I just realised that you didn't add the function to `Adjacency(Int)Map` and `Relation` -- could you do that too? They will be able to provide a faster implementation.

Also, please add an item to `CHANGES.md`.

nobrakal:
I have added a default implementation (`hasSelfLoop x = hasEdge x x`) in the `ToGraph` class, is this not sufficient ?

I don't know if writing it by hand will be better for AdjacencyMap. Is there a better implementation for `Relation` ?

snowleopard:
Ah, you are right: `hasEdge` is already implemented efficiently for `Adjacency(Int)Map` and `Relation`, so the default implementation is fine. There is also module `Fold`.

However, I prefer reexporting monomorphic versions of functions from modules, because they are easier to understand and easier to find in documentation.

For example, see: 

https://github.com/snowleopard/alga/blob/master/src/Algebra/Graph/Fold.hs#L355-L366

CHANGES.md
```diff
@@ -1,5 +1,9 @@
 # Change log
 
+## 0.2.1
+
+* #86: Add `hasSelfLoop` into the API.
```
snowleopard: This should go to `0.2`, which still hasn't been released.

nobrakal: Oops

snowleopard:
@nobrakal Many thanks! Merged.

Issue #90, opened 2018-07-03
Replacing Algebra.Graph by Maybe (Algebra.Graph.NonEmpty)


nobrakal:
Hi,

I have been working on a different implementation of `Algebra.Graph`, based on `Algebra.Graph.NonEmpty`.

The idea is to have:
```Haskell
newtype Graph a = G (Maybe (NonEmptyGraph a))
```

Using patterns, with:
```Haskell
pattern Empty :: Graph a
pattern Empty = G Nothing

pattern Vertex :: a -> Graph a
pattern Vertex a = G (Just (N.Vertex a))

pattern Overlay :: NonEmptyGraph a -> NonEmptyGraph a -> Graph a
pattern Overlay a b = G (Just (N.Overlay a b))

pattern Connect :: NonEmptyGraph a -> NonEmptyGraph a -> Graph a
pattern Connect a b = G (Just (N.Connect a b))
```
One can reproduce the traditional `Algebra.Graph` interface.

This big change can be summarized with some functions:

```Haskell
overlay :: Graph a -> Graph a -> Graph a
overlay l@(G a) r@(G b) = maybe r (\x -> maybe l (Overlay x) b) a
```

We now check the neutrality of `empty` *directly* into the `overlay` and `connect` functions.

```Haskell
edges :: [(a, a)] -> Graph a
edges = G . fmap N.edges1 . NL.nonEmpty
```
`edges` become more complicated and slower.

```Haskell
foldg :: b -> (a -> b) -> (b -> b -> b) -> (b -> b -> b) -> Graph a -> b
foldg e v o c (G g) = maybe e (N.foldg1 v o c) g
```

`foldg` is directly based on `foldg1` (and this is the case for others functions, when possible).

```Haskell
removeEdge :: Eq a => a -> a -> Graph a -> Graph a
removeEdge s t (G g) = G $ N.removeEdge s t <$> g
```
`fmap` allow a pretty syntax when a function will act as the identity on the empty graph.

## Some benefits

* Type safety in the inner representation: one can be sure when pattern matching `Overlay` or `Connect` that the two graphs are both non-empty.
* Faster graph inspection. Because of the dropped `Empty` constructor, `foldg1` is faster than `foldg`. Thus, inspection functions are faster.

## Some problems:

* The creation takes longer from a list of edges, or anything, because we pattern match to detect the empty graph.
* Graph modification takes longer, mainly this is affecting`induce` because we have to wrap the intermediate results into the `Maybe` monad.
[...]

PR #92, opened 2018-07-10
Add regression suite for NonEmpty and update current one

nobrakal:
Hi,

I have updated the script to support the benchmarks of NonEmpty graphs and created two Travis jobs dedicated to the benchmark of the `Algebra.Graph` and `Algebra.Graph.NonEmpty` module.

The tricky part is that the benchmark suite need criterion which need many dependencies. I needed to set a timeout for the action, so it fail before the Travis time limit and allow caching these dependencies (if the time limit is reached, the is no caching).

So if the build first fail, it certainly due to benchamrks dependencies, so re-run it and, as they have been cached, the build should pass.

Example here: https://travis-ci.com/nobrakal/alga/builds/78616254

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,18 +43,21 @@ matrix:
     - compiler: "ghc-8.2.2"
     # env: TEST=--disable-tests BENCH=--disable-benchmarks
       addons: {apt: {packages: [ghc-ppa-tools,cabal-install-head,ghc-8.2.2], sources: [hvr-ghc]}}
-    - compiler: "ghc-8.4.1"
-    # env: TEST=--disable-tests BENCH=--disable-benchmarks
-      addons: {apt: {packages: [ghc-ppa-tools,cabal-install-head,ghc-8.4.1], sources: [hvr-ghc]}}
-      env:
-      - TEST="" DOBENCH=0 # Added by hand
-      - TEST=--disable-tests DOBENCH=1 # Added by hand, will enable benchmarks.
-      - TEST=--disable-tests DOBENCH=2 # Added by hand, will enable benchmarks.
-
     - compiler: "ghc-8.2.2"
     # env: TEST=--disable-tests BENCH=--disable-benchmarks
       addons: {apt: {packages: [ghc-ppa-tools,cabal-install-head,ghc-8.2.2], sources: [hvr-ghc]}}
       os: osx
+
+# Added by hand
+    - compiler: "ghc-8.4.1"
+      addons: {apt: {packages: [ghc-ppa-tools,cabal-install-head,ghc-8.4.1], sources: [hvr-ghc]}}
+      env: DOBENCH=0
+    - compiler: "ghc-8.4.1"
+      addons: {apt: {packages: [ghc-ppa-tools,cabal-install-head,ghc-8.4.1], sources: [hvr-ghc]}}
+      env: TEST=--disable-tests DOBENCH=1
+    - compiler: "ghc-8.4.1"
+      addons: {apt: {packages: [ghc-ppa-tools,cabal-install-head,ghc-8.4.1], sources: [hvr-ghc]}}
+      env: TEST=--disable-tests DOBENCH=2
 
 before_install:
   - HC=${CC}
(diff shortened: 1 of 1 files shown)


nobrakal:
(as discussed in #90 )

snowleopard:
@nobrakal Thanks! Why did both benchmark suites fail?

Also, it's not really obvious which module is benchmarked where -- could make it visible in the Travis report by using a (possibly unused) environment setting like `NonEmpty=1`?

nobrakal:
It fails because of the time needed to download and build the dependencies of the bench suite. It simply takes too long. Normally, it fails only the first time, then the dependencies are cached and everything is ok after it. Please re-run the failed build to see.

I have added better variable names, I hope they are effectively better ^^

snowleopard:
> It simply takes too long

Travis has 50 min time limit, but tasks failed after 45 min. Are you forcing the termination earlier?

snowleopard:
> I have added better variable names, I hope they are effectively better

@nobrakal Yep, that's better, thank you :)

nobrakal:
> Are you forcing the termination earlier ?

Yep, either Travis does not cache what was built

nobrakal:
Sorry if I wasn't clear:

If the time limit is reached, travis __does not__ cache anything. But is a job fails, it cache. So the idea is to "manually" fail when we approach the time limit to force the cache to be done. For the current state of this PR , see: <https://travis-ci.org/snowleopard/alga/jobs/402850821#L2019> .

If now you are restarting it, it should build just fine.

snowleopard:
OK, I've restarted the jobs. Let's see how this goes...

snowleopard:
@nobrakal Aha, it worked!

I noticed that one of the benchmarks is called `creation` which is a bit vague -- could you rename it to `edges` and `edges1`, which I guess is how graphs are created?

nobrakal:
Great :)

> I noticed that one of the benchmarks is called creation which is a bit vague -- could you rename it to edges and edges1, which I guess is how graphs are created?

`creation` is not very clear, but neither is `edges` (it can return the list of edges in the graph). What about `graphFromEdges` ?

Note that I didn't replaced functions name with their `*1` counterpart, since it is messy `sed` work, and I don't think it very necessary (the used script is already not very beautiful and readable). I can do it if want to.

nobrakal:
Ah sorry, I thought you was suggesting something for the graph suite itself, not this specific version.

Ok, it can be dine with some `sed` magic :)

snowleopard:
> Ah sorry, I thought you was suggesting something for the graph suite itself, not this specific version.

Oops, sorry for being ambiguous :) I meant just this regression suite.

nobrakal:
No problem ^^

I have updated the script, please try to restart the two corresponding travis jobs :)

(only `creation`, `vertexList` and `removeVertex` needed renaming)

snowleopard:
@nobrakal Many thanks, looks good!

One more suggestion while we are at it: could you find a way to hide all build/compilation information which is not really useful for performance regression?

We should of course show all warnings etc on usual Travis instances, but for performance regression we'd like to see only performance figures, if that's possible.

nobrakal:
So I have reworked it.

Now there is a separated `script` sections for corresponding matrix entries. This disable all the the "classic" `script` and replace it by the regression suite.

It should drop unwanted outputs, and save time :)

I have also tweaked the script to hide all output.

snowleopard:
Hmm, for some reason Travis CI is now missing -- could you push another commit to try to trigger it?

nobrakal:
Ooops it was my fault, the yaml syntax is a bit tricky when it comes to new line ^^

snowleopard:
@nobrakal We now get:
```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```
Perhaps, you could output something to indicate progress, e.g. just dots on a line `...`?

nobrakal:
Fun that my tests ran without hitting this error !

I have corrected my script in https://github.com/nobrakal/benchAlgaPr/commit/9a0547ff1c95f768541819c99c67f57c542dc73b . It output a line every 5 minutes when installing dependencies.

snowleopard:
@nobrakal Many thanks! All looks good now -- merged.

PR #93, opened 2018-07-14
Use rewrite rules instead of some specialization

nobrakal:
Hi,

Trying to improve the `MaybeNonEmpty` version, I discovered that the rewrite rules are very powerful for `edgeList` (it comes from the quickness of AdjacencyIntMap):

## edgeList

Description: Produce a list of the edges in the graph

### Mesh
<TABLE>
   <TR>
      <TH>
      </TH>
      <TH>
         100
      </TH>
   </TR>
   <TR>
      <TH>
         Alga
      </TH>
      <TD>
         77.91 μs
      </TD>
   </TR>
   <TR>
      <TH>
         AlgaOld
      </TH>
      <TD>
         206.3 μs
      </TD>
   </TR>
</TABLE>

### Clique
<TABLE>
   <TR>
      <TH>
      </TH>
      <TH>
         100
      </TH>
   </TR>
   <TR>
      <TH>
         Alga
      </TH>
      <TD>
         2.347 ms
      </TD>
   </TR>
   <TR>
      <TH>
         AlgaOld
      </TH>
      <TD>
         7.619 ms
      </TD>
   </TR>
</TABLE>

### Circuit
<TABLE>
   <TR>
      <TH>
      </TH>
      <TH>
         100
      </TH>
   </TR>
   <TR>
      <TH>
         Alga
      </TH>
      <TD>
         39.63 μs
      </TD>
   </TR>
   <TR>
      <TH>
         AlgaOld
      </TH>
      <TD>
         118.9 μs
      </TD>
   </TR>
</TABLE>

SUMMARY:

 * Alga was the fastest 3 times

ABSTRACT:

 * Alga was 2.94 times faster than AlgaOld

### DIFF
--- a/src/Algebra/Graph.hs
+++ b/src/Algebra/Graph.hs
@@ -541,9 +541,14 @@ vertexIntList = IntSet.toList . vertexIntSet
 -- edgeList . 'edges'        == 'Data.List.nub' . 'Data.List.sort'
 -- edgeList . 'transpose'    == 'Data.List.sort' . map 'Data.Tuple.swap' . edgeList
 -- @
-{-# SPECIALISE edgeList :: Graph Int -> [(Int,Int)] #-}
 edgeList :: Ord a => Graph a -> [(a, a)]
 edgeList = AM.edgeList . fromGraphAM
+{-# INLINE[1] edgeList #-}
+{-# RULES "edgeList/Int" edgeList = edgeIntList #-}
+
+-- | Like 'edgeList' but specialised for graphs with vertices of type 'Int'.
+edgeIntList :: Graph Int -> [(Int,Int)]
+edgeIntList = AIM.edgeList . fromGraphAIM
 
 -- | The set of vertices of a given graph.
 -- Complexity: /O(s * log(n))/ time and /O(n)/ memory.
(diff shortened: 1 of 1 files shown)


snowleopard:
> edgeCount: 0.41 (good)
> edgeList: 0.32 (good)

@nobrakal Awesome, thank you!

Will the same come for non-empty graphs?

nobrakal:
> Will the same come for non-empty graphs?

I think yes, as well as many `SPECIALIZE` and other `RULES` of the `Algebra.Graph` module :)

So since we make no use for now of rules pragma in `Algebra.Graph.NonEmpty`, I don't know if it worth it adding only one or two here.

(In fact, I already added almost all of these rules in `Algebra.Graph.NonEmpty` for my work related to #90 )

snowleopard:
Right, I see. We could split the work on non-empty graphs into a separate PR.

> as well as many `SPECIALIZE` and other `RULES` of the `Algebra.Graph` module :)

Are these essentially the same as in this PR? In this case, I think it makes sense to combine all such changes together in one PR. Do you agree? If yes, I'll wait for further commits.

nobrakal:
> Are these essentially the same as in this PR? 

Yep this is mainly specialize and rules pragma. In fact, it would be great to add these pragmas directly into the `ToGraph` class when possible, so it benefit other implementations. What do you think ?

snowleopard:
If you manage to do this at the `ToGraph` class level it would be great, of course. But will that actually work?

nobrakal:
> But will that actually work?

You were right, I was a bit optimistic... I as not able to sneak this into the `ToGraph` class (I think this is due to TypeFamilies, which make the simple type-match of rewrite rules a bit complicated).

So I think it is a good idea to open an other PR concerning non-empty graphs, since there will be some different work unrelated to `edgeList`

snowleopard:
OK, sounds good. Then I'm waiting for further `Graph`-related commits.

nobrakal:
> I'm waiting for further Graph-related commits.

I do not have any :) Do you see anything missing for `Algebra.Graph` ?

snowleopard:
Ah, I think I just misunderstood one of your earlier messages :) Merged now!

PR #94, opened 2018-07-16
NonEmpty graphs optimizations

nobrakal:
Hi,

As discussed in #93 , here are several optimizations to `Algebra.Graph.NonEmpty`.

Mainly with:

* `SPECIALIZE` pragma when possible and useful
* `RULES` when possible and useful
* Previous improvements founds while working on #90

The benchmarks:
```
vertexList1: 0.08 (good)
vertexCount: 0.07 (good)
hasVertex: 3.71 (bad)
edgeCount: 0.36 (good)
edgeList: 0.37 (good)
hasEdge: 0.25 (good)
hasSelfLoop: 1.02 (OK)
addEdge: 1.00 (OK)
addVertex: 1.02 (OK)
removeVertex1: 1.01 (OK)
equality: 0.34 (good)
removeEdge: 0.96 (OK)
transpose: 1.00 (OK)
dff: 1.01 (OK)
topSort: 1.01 (OK)
edges1: 0.83 (good)
``` 

So everything is green, the drop of `hasVertex` as already been documented here: https://github.com/snowleopard/alga/issues/90#issuecomment-403087848

### DIFF
--- a/src/Algebra/Graph/NonEmpty.hs
+++ b/src/Algebra/Graph/NonEmpty.hs
@@ -57,12 +57,14 @@ import Control.DeepSeq (NFData (..))
 import Control.Monad.Compat
 import Data.List.NonEmpty (NonEmpty (..))
 
-import qualified Algebra.Graph         as G
-import qualified Algebra.Graph.ToGraph as T
-import qualified Data.IntSet           as IntSet
-import qualified Data.List.NonEmpty    as NonEmpty
-import qualified Data.Set              as Set
-import qualified Data.Tree             as Tree
+import qualified Algebra.Graph                 as G
+import qualified Algebra.Graph.AdjacencyMap    as AM
+import qualified Algebra.Graph.AdjacencyIntMap as AIM
+import qualified Algebra.Graph.ToGraph         as T
+import qualified Data.IntSet                   as IntSet
+import qualified Data.List.NonEmpty            as NonEmpty
+import qualified Data.Set                      as Set
+import qualified Data.Tree                     as Tree
 
 {-| The 'NonEmptyGraph' data type is a deep embedding of the core graph
 construction primitives 'vertex', 'overlay' and 'connect'. As one can guess from
--- a/src/Algebra/Graph/NonEmpty.hs
+++ b/src/Algebra/Graph/NonEmpty.hs
@@ -466,7 +466,7 @@ vertexIntCount = IntSet.size . vertexIntSet
 -- @
 {-# SPECIALISE edgeCount :: NonEmptyGraph Int -> Int #-}
 edgeCount :: Ord a => NonEmptyGraph a -> Int
-edgeCount = T.edgeCount
+edgeCount = length . edgeList
 
 -- | The sorted list of vertices of a given graph.
 -- Complexity: /O(s * log(n))/ time and /O(n)/ memory.
--- a/src/Algebra/Graph/NonEmpty.hs
+++ b/src/Algebra/Graph/NonEmpty.hs
@@ -496,7 +496,7 @@ vertexIntList1 = NonEmpty.fromList . IntSet.toAscList . vertexIntSet
 {-# RULES "edgeList/Int" edgeList = edgeIntList #-}
 {-# INLINE[1] edgeList #-}
 edgeList :: Ord a => NonEmptyGraph a -> [(a, a)]
-edgeList = AM.edgeList . foldg1 AM.vertex AM.overlay AM.connect
+edgeList = T.edgeList
 
 -- | Like 'edgeList' but specialised for NonEmptyGraph with vertices of type 'Int'.
 edgeIntList :: NonEmptyGraph Int -> [(Int,Int)]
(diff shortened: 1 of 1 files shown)


src/Algebra/Graph/NonEmpty.hs
```diff
@@ -274,7 +287,9 @@ vertices1 = overlays1 . fmap vertex
 -- 'edgeCount' . edges1   == 'Data.List.NonEmpty.length' . 'Data.List.NonEmpty.nub'
 -- @
 edges1 :: NonEmpty (a, a) -> NonEmptyGraph a
-edges1 = overlays1 . fmap (uncurry edge)
+edges1 (x :| xs) = foldr (Overlay . edgeTuple) (edgeTuple x) xs
+  where
+    edgeTuple = uncurry edge
```
snowleopard: I would drop `edgeTuple` -- I don't think duplicating `uncurry edge` is such a big deal.

src/Algebra/Graph/NonEmpty.hs
```diff
@@ -377,8 +395,31 @@ hasVertex v = foldg1 (==v) (||) (||)
 -- hasEdge x y . 'removeEdge' x y == const False
 -- hasEdge x y                  == 'elem' (x,y) . 'edgeList'
 -- @
+{-# SPECIALISE hasEdge :: Int -> Int -> NonEmptyGraph Int -> Bool #-}
 hasEdge :: Eq a => a -> a -> NonEmptyGraph a -> Bool
-hasEdge = T.hasEdge
```
snowleopard: Since you now provide much faster implementations compared to the default ones in `ToGraph`, you should use them in the `instance ToGraph` declaration.

snowleopard:
@nobrakal Thanks, I've added a couple of comments.

> vertexList1: 0.08 (good)
> vertexCount: 0.07 (good)

Amazing! For these cases we really should be saying `too good to be true` instead of just `good` ;-)

src/Algebra/Graph/NonEmpty.hs
```diff
@@ -440,8 +495,14 @@ vertexList1 = NonEmpty.fromList . T.vertexList
...
 -- @
+{-# RULES "edgeList/Int" edgeList = edgeIntList #-}
+{-# INLINE[1] edgeList #-}
 edgeList :: Ord a => NonEmptyGraph a -> [(a, a)]
-edgeList = T.edgeList
+edgeList = AM.edgeList . foldg1 AM.vertex AM.overlay AM.connect
+
+-- | Like 'edgeList' but specialised for NonEmptyGraph with vertices of type 'Int'.
+edgeIntList :: NonEmptyGraph Int -> [(Int,Int)]
+edgeIntList = AIM.edgeList . foldg1 AIM.vertex AIM.overlay AIM.connect
```
snowleopard: Presumably, `foldg1 AM.vertex AM.overlay AM.connect` and `foldg1 AIM.vertex AIM.overlay AIM.connect` will appear in the `instance ToGraph`, so you could simply write `toAdjacencyIntMap` here?

nobrakal:
@snowleopard Ahaha, yes this is great :)

Thanks for the comments, I will fix address them shortly.

The travis build is failing here because I just added some missing algorithms to the bench suite, and now the suite take too much time to run. Since the added algorithms are based on their counterpart from AdjacencyMap, I will drop them. A PR is coming soon to fix that !

src/Algebra/Graph/NonEmpty.hs
nobrakal: Writing this improvement in the `ToGraph` instance drop a bit the performances:

I think this is because I can't use `NonEmpty` graphs in `ToGraph`, so I can 't use `Algebra.Graph.NonEmpty.Vertex` but only `Algebra.Graph.Vertex`, and thus, in the final step, I can only use `foldg` instead of `foldg1`.

src/Algebra/Graph/NonEmpty.hs
nobrakal: > Presumably, foldg1 AM.vertex AM.overlay AM.connect and foldg1 AIM.vertex AIM.overlay AIM.connect will appear in the instance ToGraph

You are right for `edgeList`, it is a mistake.

For `edgeIntList` there is only `adjacencyIntMap` in `ToGraph`, so I cannot use it for `AM.edgeSet` since `adjacencyIntMap` only return a `IntMap IntSet`.

I can add `toAdjacencyIntMap` in `ToGraph` if you want.

src/Algebra/Graph/NonEmpty.hs
snowleopard: @nobrakal I meant this:

```haskell
instance ToGraph NonEmpty where
    ...
    hasEdge = hasEdge -- use faster implementation instead of the default one
```

Surely this can't reduce the performace of `hasEdge` itself!

snowleopard: But users of the typeclass might get the benefit.

So, the rule is: if you use the default implementation, i.e. `x = T.x` then the instance declaration uses default methods. But otherwise you need to redefine instance methods, i.e. `x = x`.

src/Algebra/Graph/NonEmpty.hs
snowleopard: Thanks, you are right, this seems to work only with `edgeList`.

src/Algebra/Graph/NonEmpty.hs
nobrakal: Oh yes, it is done now :)

snowleopard:
@nobrakal Thanks, merged!

PR #97, opened 2018-07-19
Optimise hasEdge, drop hasSelfLoop

snowleopard:
Work in progress

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,7 +9,7 @@
 * #76: Remove benchmarks.
 * #74: Drop dependency of `Algebra.Graph` on graph type classes.
 * #62: Move King-Launchbury graphs into `Data.Graph.Typed`.
-* #67, #68, #69, #77, #81: Various performance improvements.
+* #67, #68, #69, #77, #81, #97: Various performance improvements.
 * #66, #72: Add missing `NFData` instances.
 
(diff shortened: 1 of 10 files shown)


snowleopard:
@nobrakal Here is my low-level optimisation :) Need to add some comments!

Let's see what the regression suite says...

nobrakal:
> hasEdge: 0.17 (good)

That is a pretty cool improvement :D

snowleopard:
> hasEdge: 0.17 (good)

🎉

snowleopard:
@nobrakal Looks like we were watching in sync :D

snowleopard:
@nobrakal I now think it wasn't necessary to add `hasSelfLoop`: we can always take care of this special case in `hasEdge` itself. This makes the API smaller, and removes a potential source of confusion from users.

So, I think I'll remove `hasSelfLoop` as part of this PR. What do you think?

snowleopard:
Hmm, my "further simplification" reduced the speed up to just `hasEdge: 0.36 (good)`...

snowleopard:
OK, although the achieved speed up is not as great as initially hoped, I'm merging this. I'm not entirely happy about code duplication and the fact that default implementation in `ToGraph` may be slower than before -- I'd like to come back to this at some point.

Issue #99, opened 2018-07-25
Optimise size of mesh and torus


nobrakal:
Hi,

Working on #71 I just discovered that the `mesh` function is making graphs with a very big `size`:

```
> let m = mesh [1..10] [1..10] :: Graph (Int,Int)
> size m
685
> length $ edgeList m
180
> size $ edges $ edgeList m
361
```

One part of this is due to `Empty` graphs hidden at the end of each `path` (here `path` introduces `36` empty leaves), but there still a strange too big size :)
`box` seems to introduce duplicated edges...

snowleopard:
Yes, `box` in general is not very efficient in terms of the size of the resulting expressions :) 

In principle, one could relatively easily obtain an efficient implementation of `mesh` without `box`, but I'd like to rename this issue to make it a bit more general.

snowleopard:
One reason `box` is currently inefficient, is that it is fully polymorphic:
```haskell
box :: Graph a -> Graph b -> Graph (a, b)
box x y = overlays $ xs ++ ys
  where
    xs = map (\b -> fmap (,b) x) $ toList y
    ys = map (\a -> fmap (a,) y) $ toList x
```

Simply by switching from `toList` to `vertexList` in the implementation one could obtain a good improvement, but that requires `Ord a`.

nobrakal:
Without requiring at least an `Eq` instance, we cannot distinct vertices, so I don't know if one can make an optimized `box`.

> one could relatively easily obtain an efficient implementation of mesh without box

It certainly affects `torus` as well, so it can worth it to have a better _polymorphic_ version of `mesh` and `torus` without relying on `box`.

snowleopard:
> so it can worth it to have a better polymorphic version of `mesh` and `torus` without relying on `box`.

@nobrakal Yes, I think this is doable and the resulting functions are not going to be simple. Both `mesh` and `torus` could probably be expressed by using the `stars` combinator which was mentioned in the `edges` issue #71: just list all sources in a chessboard pattern, for example. But maybe there are even more compact representations -- an interesting puzzle :)

nobrakal:
So for `mesh`: I came to something ugly version but it gives some results and can surely be prettified.

The idea is based on `edgesOrd` results, because it can actually reduce the size.
```Haskell
mesh :: [a] -> [b] -> Graph (a, b)
mesh xs ys = overlays $ map mkAngle [ (x, y) | x <- [0..lx] , y <- [0..ly] ]
  where
[...]

PR #100, opened 2018-07-30
Rename fromAdjacencyList to stars

snowleopard:
See https://github.com/snowleopard/alga/issues/99#issuecomment-408714915

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## 0.2
 
+* #100: Rename `fromAdjacencyList` to `stars`.
 * #79: Improve the API consistency: rename `IntAdjacencyMap` to `AdjacencyIntMap`,
        and then rename the function that extracts its adjacency map to
        `adjacencyIntMap` to avoid the clash with `AdjacencyMap.adjacencyMap`,
--- a/src/Algebra/Graph.hs
+++ b/src/Algebra/Graph.hs
@@ -573,11 +573,11 @@ edgeIntSet = AIM.edgeSet . fromGraphAIM
 -- Complexity: /O(n + m)/ time and /O(m)/ memory.
 --
 -- @
--- adjacencyList 'empty'               == []
--- adjacencyList ('vertex' x)          == [(x, [])]
--- adjacencyList ('edge' 1 2)          == [(1, [2]), (2, [])]
--- adjacencyList ('star' 2 [3,1])      == [(1, []), (2, [1,3]), (3, [])]
--- 'stars' . adjacencyList == id
+-- adjacencyList 'empty'          == []
+-- adjacencyList ('vertex' x)     == [(x, [])]
+-- adjacencyList ('edge' 1 2)     == [(1, [2]), (2, [])]
+-- adjacencyList ('star' 2 [3,1]) == [(1, []), (2, [1,3]), (3, [])]
+-- 'stars' . adjacencyList        == id
 -- @
 {-# SPECIALISE adjacencyList :: Graph Int -> [(Int,[Int])] #-}
 adjacencyList :: Ord a => Graph a -> [(a, [a])]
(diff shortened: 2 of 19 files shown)


snowleopard:
@nobrakal Any comments or suggestions?

snowleopard:
@nobrakal One thing that bothers me is `starTranspose` and possible `starsTranspose` which may similarly be useful. It feels that the API shouldn't contain them, because `transpose . stars` does the same thing as `starsTranspose`, but unfortunately `transpose . stars` would be slower. I wonder if there is a way to fuse `transpose` with other graph construction functions nicely?

For example, we could have a rule `transpose (connect x y) = connect (transpose y) (transpose x)`, as well as `transpose . vertex = vertex`. Could you investigate this?

nobrakal:
> I wonder if there is a way to fuse transpose with other graph construction functions nicely?

Yep, this is possible using rewrite rules :)

```Haskell
{-# RULES
"transpose/connect"  forall g1 g2. transpose (connect g1 g2) = connect (transpose g2) (transpose g1)
"transpose/vertex"   forall v. transpose (vertex v) = vertex v
"transpose/vertices" forall xs. transpose (vertices xs) = vertices xs
"transpose/stars" forall xs. transpose (stars xs) = overlays (map (transpose . uncurry star) xs)
 #-}
```

(I have also added some `INLINE` and `NOOINLINE` pragmas for concerned functions).

Sadly, all of this includes `overlays` which is very particular: due to its simplicity, adding any `INLINE` pragma drops its performances (and I don't know why). Butif I want my rule to fire, I have to delay `overlays`'s inlining...

So I had to trick the thing using less generic rules.

Anyway the result is here:

Using:

```Haskell
import Algebra.Graph

import Criterion.Main

import Control.DeepSeq

main :: IO ()
main = defaultMain
  [
    bench "starTranspose" $ nf (starTranspose 0) $!! intList
  , bench "star . transpose" $ nf (transpose . star 0) $!! intList
  , bench "starsTranspose" $ nf starsTranspose $!! toStars
  , bench "transpose . stars" $ nf starsTranspose' $!! toStars
  ]

starsTranspose :: [(a,[a])] -> Graph a
starsTranspose = overlays . map (uncurry starTranspose)

starsTranspose' :: [(a,[a])] -> Graph a
starsTranspose' = transpose . stars

toStars :: [(Int,[Int])]
toStars = map (\x -> (x,intList)) intList

intList :: [Int]
intList = [1..1000]
```

With `-O2`:

```
benchmarking starTranspose
time                 19.75 μs   (19.67 μs .. 19.85 μs)
[...]

benchmarking star . transpose
time                 19.83 μs   (19.78 μs .. 19.88 μs)
[...]

benchmarking starsTranspose
time                 20.42 ms   (20.31 ms .. 20.60 ms)
[...]

benchmarking transpose . stars
time                 20.43 ms   (20.27 ms .. 20.74 ms)
[...]
```

And using `-O2 -fno-enable-rewrite-rules`:
```
benchmarking starTranspose
time                 19.71 μs   (19.61 μs .. 19.82 μs)
[...]

benchmarking star . transpose
time                 39.91 μs   (39.13 μs .. 40.71 μs)
[...]

benchmarking starsTranspose
time                 21.23 ms   (20.72 ms .. 21.88 ms)
[...]

benchmarking transpose . stars
time                 39.41 ms   (38.30 ms .. 40.93 ms)
[...]
```

snowleopard:
@nobrakal Cool, thank you! Presumably, you don't necessarilty need to use `-O2`? Will `-O` be sufficient?

> Sadly, all of this includes `overlays` which is very particular: due to its simplicity, adding any INLINE pragma drops its performances (and I don't know why).

Yes, we need to figure out what to do with `overlays`. Anyway, thanks for the useful experiment -- I will remove `starTranspose` and we can add your rewrite rules later, in a separate PR.

nobrakal:
> Presumably, you don't necessarilty need to use -O2? Will -O be sufficient?

Yep, I only  used`-O2` because it is the flag of alga build :)

Ok, prefect, I just want to check what happen to `overlays` and try to find better rules before

snowleopard:
> Yep, I only used `-O2` because it is the flag of alga build :)

Actually, let me try to drop it -- see #101.

PR #101, opened 2018-07-30
Drop -O2

snowleopard:
We most probably don't need it, but let's see at the results of the regression suite.


snowleopard:
@nobrakal We should probably also double-check that this doesn't lead to loss of performance in your main benchmark suite.

Issue #105, opened 2018-08-02
Add isAcyclic, topSort, dfs and other graph analysis functions to ToGraph


fosskers:
Or likewise, `isCircular`. Basically, a top-level function to determine if some `Graph` type is a DAG or not.

As [last time](https://github.com/snowleopard/alga/issues/85) (thanks again), I suspect there's some composition of the exposed primitives that "just works", but nothing jumped out at me after a brief scan of the API.

Thoughts?

snowleopard:
@fosskers Thanks! I agree, this is a useful function to add. If you don't necessarily need a polymorphic function, the implementation is quite simple:

```haskell
isDag :: Ord a => AdjacencyMap a -> Bool
isDag = isJust . topSort
```

You'll just need to convert your graph data type to `AdjacencyMap`.

With the new `ToGraph` type class (coming soon in version `0.2`), I believe we could even add it as a separate method.

fosskers:
Ah ha! So it was just a simple thing. And luckily for me, I'm already using `AdjacencyMap`.

fosskers:
So `isJust . topSort` tells us *if* there's a cycle... would it be hard to report the values of the vertices that are *causing* the cycle?

snowleopard:
@fosskers To find the cycle, you can use the [`scc` function](https://hackage.haskell.org/package/algebraic-graphs-0.1.1.1/docs/Algebra-Graph-AdjacencyMap.html#v:scc). It returns you a graph whose vertices are strongly-connected components, and if any component contains more than a single vertex it is a cycle. 

Here is one possible way to use is: 

```haskell
cycles :: Ord a => AdjacencyMap a -> [AdjacencyMap a]
cycles x = [ induce (`Set.member` c) x | c <- cs ]
  where
    cs = filter (\c -> Set.size c > 1) $ vertexList (scc x)
```

We obtain the list `cs` which contains non-singleton components (i.e. cycles), and then compute corresponding induced subgraphs of the original graph. As a result you can see all cycles in isolation.

For example:

```haskell
> cycles $ path [1, 2, 3]
[]

> cycles $ circuit [1, 2, 3]
[edges [(1,2),(2,3),(3,1)]]

> cycles $ circuit [1, 2, 3] `overlay` path [3, 4, 5] `overlay` circuit [5, 6, 7]
[edges [(1,2),(2,3),(3,1)],edges [(5,6),(6,7),(7,5)]]
```

fosskers:
[...]

PR #106, opened 2018-08-03
Extend ToGraph

snowleopard:
This is work in progress in preparation for #105.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,7 @@
 ## 0.2
 
 * #106: Extend `ToGraph` with algorithms based on adjacency maps.
-* #106: Add `isAcyclic`.
+* #106: Add `isAcyclic` and `reachable`.
 * #106: Rename `isTopSort` to `isTopSortOf`.
 * #102: Switch the master branch to GHC 8.4.3. Add a CI instance for GHC 8.6.1.
 * #101: Drop `-O2` from the `ghc-options` section of the Cabal file.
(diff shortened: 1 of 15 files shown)


snowleopard:
@nobrakal Note that this PR adds an implementation of `reachable`, which I think is part of your graph benchmark suite.

nobrakal:
@snowleopard That is cool to have it included properly in the API! It is the definition I used, so no changes in the results.

snowleopard:
@nobrakal For some reason I get much better speed up of `edgeCount` in non-empty graphs:

`edgeCount: 0.34 (good)`

I have a feeling that the `edgeCount/Int` rule does not fire in `Algebra.Graph`. Do you have any idea why this happens and how to fix it?

nobrakal:
@snowleopard 

The rule actually fired:

> Rule fired: Rule fired: edgeCount/Int (Algebra.Graph)

It will dig in the produced core, it is a really interesting speedup :)

snowleopard:
@nobrakal Thank you! :) 

Also, it looks like `edgeSet` is not part of the testsuite -- I'll try to send you a PR with it soon.

snowleopard:
Thanks, interesting! But why was `edgeCount` slower for non-empty graphs? That's quite strange.

Anyway, I hope to merge this PR later today.

fosskers:
:)

snowleopard:
@fosskers Hope to release this soon :)

PR #107, opened 2018-08-11
Rewrite rules

nobrakal:
Hi,

This is related to #100 

The rules are explaining themselves :)

```Haskell
{-# RULES
"transpose/overlays" forall xs. transpose (overlays xs) = overlays (map transpose xs)
"transpose/connect"  forall g1 g2. transpose (connect g1 g2) = connect g2 g1
 #-}
```

Then benchmarks, using:

```Haskell
main :: IO ()
main = defaultMain
  [
    bench "starTranspose" $ nf (starTranspose 0) $!! intList
  , bench "transpose . star" $ nf (transpose . star 0) $!! intList
  , bench "starsTranspose" $ nf starsTranspose $!! toStars
  , bench "transpose . stars" $ nf starsTranspose' $!! toStars
  ]

starsTranspose :: [(a,[a])] -> Graph a
starsTranspose = overlays . map (uncurry starTranspose)

starsTranspose' :: [(a,[a])] -> Graph a
starsTranspose' = transpose . stars

toStars :: [(Int,[Int])]
toStars = map (\x -> (x,intList)) intList

intList :: [Int]
intList = [1..1000]
```

```
benchmarking starTranspose
time                 38.55 μs   (38.42 μs .. 38.77 μs)
                     0.999 R²   (0.998 R² .. 1.000 R²)
mean                 38.76 μs   (38.55 μs .. 39.50 μs)
std dev              1.139 μs   (387.8 ns .. 2.487 μs)
variance introduced by outliers: 30% (moderately inflated)

benchmarking transpose . star
time                 39.13 μs   (38.95 μs .. 39.40 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 39.07 μs   (39.00 μs .. 39.21 μs)
std dev              338.3 ns   (204.6 ns .. 542.3 ns)

benchmarking starsTranspose
time                 39.22 ms   (39.03 ms .. 39.36 ms)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 39.30 ms   (39.23 ms .. 39.37 ms)
std dev              154.7 μs   (125.0 μs .. 189.0 μs)

benchmarking transpose . stars
time                 39.48 ms   (39.30 ms .. 39.63 ms)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 39.24 ms   (39.12 ms .. 39.41 ms)
std dev              290.4 μs   (196.7 μs .. 449.5 μs)

```

So everything seems to work :)

So this PR removes the need of `starTranspose`... Do you think it is a good idea to remove it of the API ?
It can maybe be useful somewhere else...


snowleopard:
Great! Yes, I think we should remove `starTranspose`, since it only saves a few characters, but comes with a maintenance/complexity cost.

I guess we should add similar rules for `empty`, `vertex`, `overlay`, `connects` and non-empty graphs. At some point it would be interesting to generalise this to arbitrary homomorphisms (like `removeVertex`) and antihomomorphisms (like `transpose`). All such transformations can benefit from similar rewrite rules -- perhaps we could achieve this with some tricky use of `newtype`s?

Also I wonder if it is possible to add tests that make sure we never accidentally disable these (or any other!) rules from firing.

nobrakal:
Ok for `starTranspose` :)

> All such transformations can benefit from similar rewrite rules -- perhaps we could achieve this with some tricky use of newtypes?

It can be great to use rules where we can but I don't know the impact on the compilation time...
Also I don't know how we can use newtypes, but it needs some investigation !

> Also I wonder if it is possible to add tests that make sure we never accidentally disable these (or any other!) rules from firing.

I have added some tests:

The function used lookup in `test/Rewrite/Test/` for files ending in `.hs`, and search if rules specified in comments are firing (by running directly ghc and looking at is output). The syntax is:

```Haskell
{-
rulename1
rulename2
-}

main :: IO ()
main = ...
```

I don't know yet if this work with Stack (because it runs ghc with system-installed libraries, and not the current one). But it seems to work with `Cabal`.

snowleopard:
I think some rules are incorrect, for example: `transpose (Overlay g1 g2) = Overlay g1 g2` is missing recursive `transpose` calls. I wonder if more general graph tests are needed to make sure this is tested :)

Good start on the tests for rewrite rules! I would prefer less magic like parsing comments, and more explicit expectations for documentation purposes. For example, you could put `G.transpose (G.vertex 1))` into a separate file, and create another file with a list of rules that are expected to fire in this case, e.g. one expected rule per line. Maybe this deserves a separate issue for discussion.

nobrakal:
> I think some rules are incorrect,

Haha, you are right, I had `edge` in my mind. Main consequence is that the beautiful solution is not working anymore, because of the "propagated" transpose. Indeed, for the `vertices` part of `star`, the `overlays . map vertex` merge too quickly, and the rules are not firing after. Modifying anything drops the (fragile) work I made on `overlays` and `connects`.

 Some solutions:

* Making a rule for `transpose . vertices`
* Making a rule for `transpose . star`

Less general, but working :)

>  I wonder if more general graph tests are needed to make sure this is tested :)

Yes, it can be useful ^^. Fun that the test-suite passed !

> For example, you could put G.transpose (G.vertex 1)) into a separate file, and create another file with a list of rules that are expected to fire in this case, e.g. one expected rule per line.

That is what GHC is doing (more or less) for its test. I found the "one file" solution better, but it is simple to switch to your solution.
Note that it will require some other magic (like writing a temporary file with a main function and right imports) if we stick strictly to your idea. I will prefer a two files solution, one with a plain Haskell file (and not just a line like `.transpose (G.vertex 1))`, and another with the list of rules that are required to fire.

> Maybe this deserves a separate issue for discussion.

Do you want tests for specialize rules ? If not, the only rules that we will test are in this PR :)

snowleopard:
@nobrakal Apologies for long silence, I'm slowly returning to working mode after a holiday :)

Let's limit this PR to solving the issue with `transpose`/`stars`. 

Regarding testing rewrite rules -- I am worried by the added complexity of the testsuite. Let's keep things simple for now. I think if someone accidentally disables a rewrite rule, this will be caught by the performance regression testsuite.

nobrakal:
Hi,

I am slowly getting back to work too !

Here is some work:

* I reverted the tests
* I added (hopefully) right rewrite rules corresponding to the algebra PLUS one targeting `vertices` because I am not able to rewrite `overlays . map vertex`.
* I remove `starTranspose` from the API, and kept the function only in `Fold` since it is used in `Fold.filterContext`

snowleopard:
@nobrakal Hi, welcome back!

Looks great, but I have a couple of questions:

> PLUS one targeting `vertices`

Perhaps we should also add a rule for `clique`, the twin of `vertices`?

> I remove `starTranspose` from the API, and kept the function only in `Fold` since it is used in `Fold.filterContext`

I don't understand, why didn't you replace it with `star . transpose` in `Fold.filterContext`?

nobrakal:
> Perhaps we should also add a rule for clique, the twin of vertices?

Done !

> I don't understand, why didn't you replace it with star . transpose in Fold.filterContext?

Because I have not added any rules in `Fold` and thus `star . transpose` will be slower than `starTranspose` :)

snowleopard:
> Because I have not added any rules in `Fold`

Well, this begs the question: Why not? :-) Will the same rules work for `Fold` too? If yes, let's add them!

nobrakal:
After a day of battle, I have only a `transpose . star` working rule. I was not able to achieve the `transpose . stars` equivalent (I suppose a problem with the core of `Fold` itself: graph primitives are functions).

Do you think it is a good idea to add these rules to others graphs implementations (to name them: `Relation`, `AdjacencyMap` and `IntAdjacencyMap`) ?

I was thinking about a generalization of 
```Haskell
"transpose/Empty"    transpose Empty = Empty
"transpose/Vertex"   forall x. transpose (Vertex x) = Vertex x
"transpose/Overlay"  forall g1 g2. transpose (Overlay g1 g2) = Overlay (transpose g1) (transpose g2)
"transpose/Connect"  forall g1 g2. transpose (Connect g1 g2) = Connect (transpose g2) (transpose g1)
```
rules to something catching any use of `foldg` (ie. `foldg e v o c Empty == e`). But this seems like just an inlining of `foldg` so I don't really know what to do ^^

snowleopard:
@nobrakal Thanks for trying :) I also realised that `Fold` is missing the recent improvements you made to `Graph` and `NonEmptyGraph` based on `concatg` -- perhaps this is the reason you can't make the rules work?

> Do you think it is a good idea to add these rules to others graphs implementations (to name them: `Relation`, `AdjacencyMap` and `IntAdjacencyMap`) ?

Yes, I think this will be useful. We do have `starTranspose` in them and if we could get rid of it using rules that would be great.

> I was thinking about a generalization of

I think there may be several possible generalisations, but let's not go into this right now :) I think having the rules for `transpose` is a great starting point -- we will build further generalisations on top.

nobrakal:
> perhaps this is the reason you can't make the rules work?

I think yes, you are certainly right !

> Yes, I think this will be useful. We do have starTranspose in them and if we could get rid of it using rules that would be great.

Ok, I will do it soon. `transpose . stars` is not defined by `overlays . map star` so the corresponding rewrite rule will be more complicated and targeting *only* `stars`. Do you think it will worth it ? I am starting to think that this will seems very "magical" to an end user.

> we will build further generalisations on top.

Ok :)

snowleopard:
> `transpose . stars` is not defined by `overlays . map star` so the corresponding rewrite rule will be more complicated and targeting only `stars`. Do you think it will worth it ?

No, I don't think we should write different/complicated rules for every function. Ideally we should write just a few basic ones and automatically get all the benefits, like in `Graph` :)

snowleopard:
@nobrakal I'm keen to merge this soon. If you don't have time to implement rewrite rules for other modules, shall we just merge this for now and create an issue to finish the rest?

snowleopard:
Just to clarify: I'd like to make the new release of Alga tomorrow and I think it would be nice to get rid of `starTranspose` in it.

nobrakal:
> shall we just merge this for now and create an issue to finish the rest?

Of course :) 

For information:

* `transpose . star` is rewritten in:
  * Algebra.Graph
  * Algebra.Graph.NonEmpty
  * Algebra.Graph.AdjacencyMap
  * Algebra.Graph.AdjacencyIntMap
  * Algebra.Graph.Fold

* `transpose . stars` is rewritten in:
  * Algebra.Graph
  * Algebra.Graph.NonEmpty

I have trouble for `Relation` :)

> Just to clarify: I'd like to make a new release of Alga tomorrow and I think it would be nice to get rid of starTranspose in it.

Great ! I agree. `starTranspose` removal is now handled in (almost) all modules with my recent commits. The `stars`  version is more problematic yet !

snowleopard:
@nobrakal Many thanks! Will merge once the tests are complete.

Could you create an issue with the above list of what's been done here and what still needs to be done?

> `starTranspose` removal is now handled in (almost) all modules with my recent commits. 

I just noticed that there is still `GraphAPI.starTranspose` -- shall this be removed too?

nobrakal:
> Could you create an issue with the above list of what's been done here and what still needs to be done?

Done in #116

> I just noticed that there is still GraphAPI.starTranspose -- shall this be removed too?

Done, this should also the regression suite to complete :)

snowleopard:
Merged, thank you!

PR #110, opened 2018-08-23
Make mesh and torus without box

nobrakal:
Hi,

This is related to #99 

I added a implementation for `NonEmptyGraph` (even if I am not satisfied, the solution is a bit tricky) and some tests to ensure that size bounds are satisfied.

For `torus`:

```
size (torus xs ys) <= max 1 (3 * (length xs) * (length ys))
```

And for `mesh`

```
size (mesh xs ys) <= max 1 (3*(length xs - 1)*(length ys - 1) + 2*(length xs + length ys - 2))
```


snowleopard:
@nobrakal Thanks! I'm still a bit unhappy about the use of partial functions `init` and `last`, but it may be more productive for me to try to simplify things myself after merging your implementation :)

> `size (torus xs ys) <= max 1 (3 * (length xs) * (length ys))`

Is this really an inequality? Could this actually be strengthened to equality? Same question about `mesh`.

test/Algebra/Graph/Test/Graph.hs
```diff
@@ -53,39 +53,45 @@ testGraph = do
...
+    test "torus xs     ys    == box (circuit xs) (circuit ys)" $ \(xs :: [Int]) (ys :: [Int]) ->
+          torus xs     ys    == box (circuit xs) (circuit ys)
 
-    test "torus [1,2] \"ab\" == <correct result>" $
-          torus [1,2]  "ab"  == edges [ ((1,'a'),(1,'b')), ((1,'a'),(2,'a')), ((1,'b'),(1,'a')), ((1,'b'),(2,'b'))
+    test "torus [1,2]  \"ab\"  == <correct result>" $
+          torus [1,2]   "ab"   == edges [ ((1,'a'),(1,'b')), ((1,'a'),(2,'a')), ((1,'b'),(1,'a')), ((1,'b'),(2,'b'))
                                       , ((2,'a'),(1,'a')), ((2,'a'),(2,'b')), ((2,'b'),(1,'b')), ((2,'b'),(2 :: Int,'a')) ]
 
+    test "size (torus xs ys) <= max 1 (3 * (length xs) * (length ys))" $ \(xs :: [Int]) (ys :: [Int]) ->
```
snowleopard: No need for parentheses around `length`.

snowleopard:
By the way, I think the size expression for `mesh` can be simplified by opening parentheses.

nobrakal:
> Is this really an inequality? Could this actually be strengthened to equality? Same question about mesh.

Not it is an equality :) 

> By the way, I think the size expression for mesh can be simplified by opening parentheses.

Yep it is done

snowleopard:
@nobrakal Thanks, this looks better :) Merged!

Issue #78, opened 2018-06-12
Failing tests for latest master


srid:
```
[ 5 of 14] Compiling Algebra.Graph.Test.Generic ( test/Algebra/Graph/Test/Generic.hs, dist/build/test-alga/test-alga-tmp/Algebra/Graph/Test/Generic.dyn_o )

test/Algebra/Graph/Test/Generic.hs:371:77: error:
    Variable not in scope: applyFun2 :: t0 -> Int -> Int -> Int

test/Algebra/Graph/Test/Generic.hs:371:94: error:
    Variable not in scope: applyFun2 :: t1 -> Int -> Int -> Int
builder for '/nix/store/83kmp44pqq2fnb7zwmv4xs5adz857hcn-algebraic-graphs-0.2.drv' failed with exit code 1
```

snowleopard:
@srid Hmm, that's strange -- tests are passing on our CI bots (and locally for me).

This looks like a problem with the version of QuickCheck. Which version are you using?

srid:
Ah, looks like it. I must be using a really older version (whatever is available in reflex-platform).

PR #112, opened 2018-08-31
remove setProduct, use cartesianProduct (fixes #111)

### LINKED ISSUE #111: Use new cartesianProduct from containers?
Hi, 
Awesome project you have! I have one very minor comment to make.
I noticed that you're using your own `setProduct` function: https://github.com/snowleopard/alga/blob/76d01ae5ea12279505c92238fb7dc771dcb1e19e/src/Algebra/Graph/Relation/Internal.hs#L173 I just wanted to make sure you are aware that there was a `cartesianProduct` function recently added to containers (Since: containers-0.5.11)! And luckily you're on a stackage lts version which already has it:
 https://www.stackage.org/haddock/lts-12.4/containers-0.5.11.0/Data-Set.html#v:cartesianProduct
Cheers.

subttle:
fixes #111 
Changes pass `stack test` on my machine.

### DIFF
--- a/src/Algebra/Graph/Relation/Internal.hs
+++ b/src/Algebra/Graph/Relation/Internal.hs
@@ -12,7 +12,7 @@
 -----------------------------------------------------------------------------
 module Algebra.Graph.Relation.Internal (
     -- * Binary relation implementation
-    Relation (..), empty, vertex, overlay, connect, setProduct, consistent,
+    Relation (..), empty, vertex, overlay, connect, consistent,
     referredToVertexSet
   ) where
 
@@ -164,14 +164,10 @@ overlay x y = Relation (domain x `union` domain y) (relation x `union` relation 
 -- @
 connect :: Ord a => Relation a -> Relation a -> Relation a
 connect x y = Relation (domain x `union` domain y)
-    (relation x `union` relation y `union` (domain x `setProduct` domain y))
+    (relation x `union` relation y `union` (domain x `Set.cartesianProduct` domain y))
 
 instance NFData a => NFData (Relation a) where
     rnf (Relation d r) = rnf d `seq` rnf r `seq` ()
-
--- | Compute the Cartesian product of two sets. /Note: this function is for internal use only/.
-setProduct :: Set a -> Set b -> Set (a, b)
-setProduct x y = Set.fromDistinctAscList [ (a, b) | a <- Set.toAscList x, b <- Set.toAscList y ]
 
 instance (Ord a, Num a) => Num (Relation a) where
     fromInteger = vertex . fromInteger
--- a/src/Algebra/Graph/Relation/Internal.hs
+++ b/src/Algebra/Graph/Relation/Internal.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module     : Algebra.Graph.Relation.Internal
@@ -12,7 +13,7 @@
 -----------------------------------------------------------------------------
 module Algebra.Graph.Relation.Internal (
     -- * Binary relation implementation
-    Relation (..), empty, vertex, overlay, connect, consistent,
+    Relation (..), empty, vertex, overlay, connect, setProduct, consistent,
     referredToVertexSet
   ) where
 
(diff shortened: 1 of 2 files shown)


src/Algebra/Graph/Relation/Internal.hs
```diff
@@ -164,15 +164,11 @@ overlay x y = Relation (domain x `union` domain y) (relation x `union` relation
...
 connect x y = Relation (domain x `union` domain y)
-    (relation x `union` relation y `union` (domain x `setProduct` domain y))
+    (relation x `union` relation y `union` (domain x `Set.cartesianProduct` domain y))
 
 instance NFData a => NFData (Relation a) where
     rnf (Relation d r) = rnf d `seq` rnf r `seq` ()
 
--- | Compute the Cartesian product of two sets. /Note: this function is for internal use only/.
-setProduct :: Set a -> Set b -> Set (a, b)
-setProduct x y = Set.fromDistinctAscList [ (a, b) | a <- Set.toAscList x, b <- Set.toAscList y ]
```
snowleopard: It looks like we need to keep this function, but provide two different implementation: the current one for earlier GHC versions that do not ship with `containers-0.5.11`, and a new one based on `Set.cartesianProduct`. This will require some CPP, which is ugly, but seems unavoidable if we want to support earlier GHC versions.

snowleopard:
Many thanks @subttle!

Alas, it looks like we'll have to add some support for earlier versions of GHC (I think for GHC < 8.4), which didn't ship with `cartesianProduct`. See my comment above.

snowleopard:
Actually, perhaps this will also involve some conditionals in the Cabal file, since we need to depend on different versions of the `containers` library... This looks like a job for `containers-compact`, but it doesn't seem to exist yet (we do use `base-compact` already for compatibility with earlier versions of GHC).

snowleopard:
Here is the list of GHC versions which can't compile this PR:

https://travis-ci.org/snowleopard/alga/builds/423003930

subttle:
Ah, my apologies. I'm not a stack expert by any means but I can look into adding the conditionals. As of right now I think I have a solution which doesn't involve any CPP, rather just adding to the `stack.yaml` file:
```
extra-deps:
- containers-0.5.11.0
```
and then adding this to the `algebraic-graphs.cabal` file:
```
    build-depends:      array       >= 0.4      && < 0.6,
                        base        >= 4.7      && < 5,
                        base-compat >= 0.9.1    && < 0.11,
                        containers  >= 0.5.11.0 && < 0.8,
                        deepseq     >= 1.3.0.1  && < 1.5
```
```
    build-depends:      algebraic-graphs,
                        array        >= 0.4      && < 0.6,
                        base         >= 4.7      && < 5,
                        base-compat  >= 0.9.1    && < 0.11,
                        base-orphans >= 0.5.4    && < 0.8,
                        containers   >= 0.5.11.0 && < 0.8,
                        extra        >= 1.5      && < 2,
                        QuickCheck   >= 2.9      && < 2.12
```
I'm not sure if using `extra-deps` appeals to you, however. I will look into doing this the other way.

snowleopard:
@subttle Hmm, I'm puzzled about how the `extra-deps` trick is supposed to work. For GHC < 8.4, won't there be a conflict between two different versions of `containers`? 

In any case, the `extra-deps` trick seems to be a Stack-only solution, but we want Cabal users to be able to build Alga with earlier GHCs too.

src/Algebra/Graph/Relation/Internal.hs
subttle: I will give this approach a try and get back to you!

subttle:
To be honest I haven't looked into stack too much because I plan on switching to Nix in the future, but I tested it yesterday with `stack --resolver=lts-10.0 build` and then `stack --resolver=lts-10.0 test` on a fresh clone with the changes I described above and it worked even though GHC < 8.4 and containers is < 0.5.11.0 for lts-10.0. Oddly enough when I tested it with lts-9.0 just now it built fine but failed the test portion with the same error as #78 . At any rate, it doesn't matter now because I will try the other way that you described, you're right that a not-stack-only solution would of course be better!

snowleopard:
Thanks @subttle!

subttle:
Hi again, I added the CPP conditional the way you described but I did not make any changes to
 `algebraic-graphs.cabal`, please let me know if there are changes you would like made there.

snowleopard:
@subttle Many thanks! I think that's exactly what we need. All tests pass, so it looks like we don't need to add anything to the Cabal file.

Merged!

Issue #17, opened 2017-07-12
Edge labels


snowleopard:
The current implementation has no support for edge labels and this is a serious limitation. Let's use this issue to discuss possible designs.

As an example, consider a graph with vertices `1`, `2`, `3` and labelled edges `1 -{a}-> 2` and `2 -{b}-> 3`.

[The paper](https://github.com/snowleopard/alga-paper) mentions this approach:
```haskell
type LabelledGraph b a = b -> Graph a

g :: LabelledGraph Char Int
g 'a' = 1 * 2
g 'b' = 2 * 3
g _   = empty
```
Here `b` stands for the type of labels that can be attached not only to edges, but also vertices and arbitrary subgraphs. Function application `g x` acts as a projection: it extracts the subgraph labelled with `x`. 

A similar, but somewhat simpler approach is to use lists of pairs instead of functions:
```haskell
type LabelledGraph b a = [(b, Graph a)]

g :: LabelledGraph Char Int
g = [('a', 1 * 2), ('b', 2 * 3)]
```
To obtain a projection, we filter the list by the edge label and overlay the graphs. The advantage of this approach compared to the above is that it's more explicit, and we can easily implement equality checking.

A very different approach is to add a new labelling operator `[b]` to the algebra. Here to obtain a nice algebraic structure the type of labels `b` need to form an idempotent semiring. For example, this works very well for Boolean labels:

* `[0]a = empty`
* `[1]a = a`
* `[x]a + [y]a = [x | y]a`
* `[x][y]a = [x & y]a`
* `[x](a + b) = [x]a + [x]b`
* `[x](a * b) = [x]a * [x]b`

This should also work well for numeric labels, e.g. usually a distance label `x` on an edge corresponds to the interval `[x; infinity)` in terms of possible time it takes to traverse the edge, and such intervals, I believe form an idempotent semiring.

The cost is a new graph construction primitive:
```haskell
data G b a = Empty
           | Vertex a
           | Overlay (G b a) (G b a)
           | Connect (G b a) (G b a)
           | Label b (G b a)

g :: G Char Int
g = Overlay (Label 'a' $ Vertex 1 `Connect` Vertex 2) (Label 'b' $ Vertex 2 `Connect` Vertex 3)
```

I haven't yet tried any of these approaches on a serious example.

I'm pretty sure there are other approaches and the design space is quite large. Do you have any suggestions?

justinlynn:
[...]

PR #114, opened 2018-09-03
Add a basic implementation of edge-labelled graphs

snowleopard:
See #17.


nobrakal:
> I believe fgl also has support for them.

Indeed, all graphs in FGL are labelled ^^. Unlabelled graphs are made using `()` as label !

Concerning benchmarks, I think it will do it :)

PR #121, opened 2018-09-23
Remove foldable

nobrakal:
This is related to #95 

So:

* It removes `Foldable` and `Traversable` instance from concerned Graph data-types.
* It removes comments related to `Foldable` instances.
* It drops `isEmpty`, `hasVertex`, `vertexCount`, `vertexList`, `vertexSet`, `vertexIntSet` and `box` from `Algebra.Graph.HigherKinded.Class` because of the removal of the `Foldable` constraint. `mesh` and ` torus` were rewrited without `box`.
* I don't really know what is happening in `Algebra.Graph.Labelled` but I had to add a `foldgUnDiff` and a real `hasVertex` (because `elem` was used previously).
* I added a comment related to the `Foldable` instance in `Algebra.Graph.ToGraph` (I don't know if it is in the right module).

### DIFF
--- a/src/Algebra/Graph.hs
+++ b/src/Algebra/Graph.hs
@@ -132,19 +132,7 @@ be computed as follows:
 m == 'edgeCount' g
 s == 'size' g@
 
-Note that 'size' is slightly different from the 'length' method of the
-'Foldable' type class, as the latter does not count 'empty' leaves of the
-expression:
-
-@'length' 'empty'           == 0
-'size'   'empty'           == 1
-'length' ('vertex' x)      == 1
-'size'   ('vertex' x)      == 1
-'length' ('empty' + 'empty') == 0
-'size'   ('empty' + 'empty') == 2@
-
-The 'size' of any graph is positive, and the difference @('size' g - 'length' g)@
-corresponds to the number of occurrences of 'empty' in an expression @g@.
+Note that 'size' count all leaves of the expression.
 
 Converting a 'Graph' to the corresponding 'AM.AdjacencyMap' takes /O(s + m * log(m))/
 time and /O(s + m)/ memory. This is also the complexity of the graph equality test,
@@ -356,10 +344,9 @@ concatg combine = fromMaybe empty . foldr1Safe combine
 -- @
 -- foldg 'empty' 'vertex'        'overlay' 'connect'        == id
 -- foldg 'empty' 'vertex'        'overlay' (flip 'connect') == 'transpose'
--- foldg []    return        (++)    (++)           == 'Data.Foldable.toList'
--- foldg 0     (const 1)     (+)     (+)            == 'Data.Foldable.length'
 -- foldg 1     (const 1)     (+)     (+)            == 'size'
 -- foldg True  (const False) (&&)    (&&)           == 'isEmpty'
+-- foldg False ((==) v)      (||)    (||)           == 'hasVertex v'
 -- @
 foldg :: b -> (a -> b) -> (b -> b -> b) -> (b -> b -> b) -> Graph a -> b
 foldg e v o c = go
--- a/src/Algebra/Graph.hs
+++ b/src/Algebra/Graph.hs
@@ -52,7 +52,7 @@ module Algebra.Graph (
   ) where
 
 import Prelude ()
-import Prelude.Compat
+import Prelude.Compat hiding ((<>))
 
 import Control.Applicative (Alternative)
 import Control.DeepSeq (NFData (..))
--- a/src/Algebra/Graph.hs
+++ b/src/Algebra/Graph.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE DeriveFunctor, DeriveFoldable, DeriveTraversable #-}
+{-# LANGUAGE DeriveFunctor #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module     : Algebra.Graph
@@ -155,7 +155,7 @@ data Graph a = Empty
              | Vertex a
              | Overlay (Graph a) (Graph a)
              | Connect (Graph a) (Graph a)
-             deriving (Foldable, Functor, Show, Traversable)
+             deriving (Functor, Show)
 
 instance NFData a => NFData (Graph a) where
     rnf Empty         = ()
@@ -844,7 +844,6 @@ removeVertex v = induce (/= v)
 removeEdge :: Eq a => a -> a -> Graph a -> Graph a
 removeEdge s t = filterContext s (/=s) (/=t)
 
-
 -- TODO: Export
 -- | Filter vertices in a subgraph context.
 {-# SPECIALISE filterContext :: Int -> (Int -> Bool) -> (Int -> Bool) -> Graph Int -> Graph Int #-}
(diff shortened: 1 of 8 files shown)


src/Algebra/Graph.hs
```diff
@@ -132,19 +132,7 @@ be computed as follows:
...
-@'length' 'empty'           == 0
-'size'   'empty'           == 1
-'length' ('vertex' x)      == 1
-'size'   ('vertex' x)      == 1
-'length' ('empty' + 'empty') == 0
-'size'   ('empty' + 'empty') == 2@
-
-The 'size' of any graph is positive, and the difference @('size' g - 'length' g)@
-corresponds to the number of occurrences of 'empty' in an expression @g@.
+Note that 'size' count all leaves of the expression.
```
snowleopard: "count" -> "counts"

Also, I think we can keep `size`-related examples above, and replace `length` examples by `vertexCount` examples.

src/Algebra/Graph.hs
```diff
@@ -356,10 +344,9 @@ concatg combine = fromMaybe empty . foldr1Safe combine
 -- @
 -- foldg 'empty' 'vertex'        'overlay' 'connect'        == id
 -- foldg 'empty' 'vertex'        'overlay' (flip 'connect') == 'transpose'
--- foldg []    return        (++)    (++)           == 'Data.Foldable.toList'
--- foldg 0     (const 1)     (+)     (+)            == 'Data.Foldable.length'
 -- foldg 1     (const 1)     (+)     (+)            == 'size'
 -- foldg True  (const False) (&&)    (&&)           == 'isEmpty'
+-- foldg False ((==) v)      (||)    (||)           == 'hasVertex v'
```
snowleopard: Let's switch to `foldg False (==x)      (||)    (||)           == hasVertex x` to match the testsuite as well as actual `hasVertex` implementation.

src/Algebra/Graph.hs
```diff
@@ -1000,8 +986,9 @@ simple op x y
 box :: Graph a -> Graph b -> Graph (a, b)
 box x y = overlays $ xs ++ ys
   where
-    xs = map (\b -> fmap (,b) x) $ toList y
-    ys = map (\a -> fmap (a,) y) $ toList x
+    xs = map (\b -> fmap (,b) x) $ toListGr y
+    ys = map (\a -> fmap (a,) y) $ toListGr x
+    toListGr = foldg [] pure (++) (++)
```
snowleopard: I suggest to use the name `toList` since the function is equivalent to the standard `toList`.

By the way, are you sure this is fast enough? Do we need to worry about potential quadratic list concatenation?

You might want to use `Algebra.Graph.Internal.List` instead to achieve constant-time list concatenation.

src/Algebra/Graph/Fold.hs
```diff
@@ -125,19 +125,7 @@ computed as follows:
...
-@'length' 'empty'           == 0
-'size'   'empty'           == 1
-'length' ('vertex' x)      == 1
-'size'   ('vertex' x)      == 1
-'length' ('empty' + 'empty') == 0
-'size'   ('empty' + 'empty') == 2@
-
-The 'size' of any graph is positive, and the difference @('size' g - 'length' g)@
-corresponds to the number of occurrences of 'empty' in an expression @g@.
+Note that 'size' count all leaves of the expression.
```
snowleopard: See https://github.com/snowleopard/alga/pull/121/files#r219704369

src/Algebra/Graph/Fold.hs
```diff
@@ -341,10 +323,9 @@ connects = foldr connect empty
 -- @
 -- foldg 'empty' 'vertex'        'overlay' 'connect'        == id
 -- foldg 'empty' 'vertex'        'overlay' (flip 'connect') == 'transpose'
--- foldg []    return        (++)    (++)           == 'Data.Foldable.toList'
--- foldg 0     (const 1)     (+)     (+)            == 'Data.Foldable.length'
 -- foldg 1     (const 1)     (+)     (+)            == 'size'
 -- foldg True  (const False) (&&)    (&&)           == 'isEmpty'
+-- foldg False ((==) v)      (||)    (||)           == 'hasVertex v'
```
snowleopard: See https://github.com/snowleopard/alga/pull/121/files#r219704458

src/Algebra/Graph/HigherKinded/Class.hs
```diff
@@ -436,6 +359,23 @@ star :: Graph g => a -> [a] -> g a
...
+-- stars []                      == 'empty'
+-- stars [(x, [])]               == 'vertex' x
+-- stars [(x, [y])]              == 'edge' x y
+-- stars [(x, ys)]               == 'star' x ys
+-- stars                         == 'overlays' . map (uncurry 'star')
+-- stars . 'adjacencyList'         == id
+-- 'overlay' (stars xs) (stars ys) == stars (xs ++ ys)
+-- @
+stars :: Graph g => [(a, [a])] -> g a
+stars = overlays . map (uncurry star)
```
snowleopard: Does it also make sense to switch to your better `overlays` implementation based on `foldr1Safe`?

src/Algebra/Graph/Labelled.hs
```diff
@@ -112,11 +112,41 @@ instance Dioid e => C.Graph (Graph e a) where
...
+-- The order of arguments is: empty, vertex, overlay and connect.
+-- Complexity: /O(s)/ applications of given functions. As an example, the
+-- complexity of 'size' is /O(s)/, since all functions have cost /O(1)/.
+--
+-- @
+-- foldgUnDiff 1     (const 1)     (+)  == 'size'
+-- foldgUnDiff True  (const False) (&&) == 'isEmpty'
+-- foldgUnDiff False ((==) v)      (||) == 'hasVertex v'
+-- @
+foldgUnDiff :: b -> (a -> b) -> (b -> b -> b) ->  Graph e a -> b
```
snowleopard: Let's not make any changes to this module in this PR -- I've got a big refactoring of labelled graphs in a local branch where I implemented a similar fold function and removed `Foldable` -- I'll make the corresponding PR soon.

src/Algebra/Graph/NonEmpty.hs
```diff
@@ -122,9 +122,7 @@ if @g@ is a 'NonEmptyGraph' then /n/, /m/ and /s/ can be computed as follows:
 m == 'edgeCount' g
 s == 'size' g@
 
-The 'size' of any graph is positive and coincides with the result of 'length'
-method of the 'Foldable' type class. We define 'size' only for the consistency
-with the API of other graph representations, such as "Algebra.Graph".
+Note that 'size' count all leaves of the expression.
```
snowleopard: See https://github.com/snowleopard/alga/pull/121/files#r219704369

src/Algebra/Graph.hs
nobrakal: Indeed, it is better !

src/Algebra/Graph/HigherKinded/Class.hs
nobrakal: I don't know, I think yes ! I will do some benchs tomorrow

nobrakal: Mmhh, I spoke too quickly, the `concatg` solution seems to not work, at least not out-of-the-box...
It will certainly require further work, maybe this is linked with the same problem in `Algebra.Graph.Fold`

snowleopard:
@nobrakal Is this ready or do you plan to do any more work?

nobrakal:
I think this is ready. The `stars` issue is certainly linked to other things, I will have a look at it later :)

snowleopard:
@nobrakal Thanks again, merged!

PR #123, opened 2018-10-04
Optimize foldg, fmap, induce and transpose composition

nobrakal:
Hi,

This is the code concerning the optimization I described in my last blog article: [Rewrite rules and a specific fold: use optimization techniques from GHC.Base](https://blog.nyarlathotep.one/2018/09/rewrite-rules-and-a-specific-fold/).

It is the same ideas than the rules behind `foldr . map` optimization.

These rules are made only for `Algebra.Graph` using `Algebra.Graph.Fold` (to avoid  circular dependencies, I have moved a part of the `Algebra.Graph.Fold` into a new module `Algebra.Graph.Fold.Internal` ).

A quick benchmark, using Criterion:

```Haskell
import Algebra.Graph
import Control.DeepSeq
import Criterion.Main

main :: IO ()
main = defaultMain
  [ bgroup "foldgFusion"
    [ bench "foldg" $ nf (foldg 0 (+1) (+) (+)) $!! grInt
    , bench "foldg . induce" $ nf (foldg 0 (+1) (+) (+) . induce (/= 2)) $!! grInt
    , bench "foldg . transpose" $ nf (foldg 0 (+1) (+) (+) . transpose) $!! grInt
    , bench "foldg . fmap" $ nf (foldg 0 (+1) (+) (+) . fmap (+1)) $!! grInt
    ]
    
  , bgroup "fmapFusion"
    [ bench "fmap" $ nf (fmap (+1)) $!! grInt
    , bench "fmap . fmap" $ nf (fmap (+1) . fmap (+1)) $!! grInt
    ]
    
  , bgroup "transposeRewriteBack"
    [ bench "transpose" $ nf (transpose . vertices) $!! intList
    , bench "transpose'" $ nf (foldg Empty Vertex Overlay (flip Connect) . vertices) $!! intList
    ]

  ,  bgroup "induceRewriteBack"
    [ bench "induce" $ nf (induce (/=1) . vertices) $!! intList
    , bench "induce'" $ nf (induce' (/= 1) . vertices) $!! intList
    ]
  ]

grInt :: Graph Int
grInt = vertices intList

intList :: [Int]
intList = [0..9999]

induce' :: (a -> Bool) -> Graph a -> Graph a
induce' p = foldg Empty (\x -> if p x then Vertex x else Empty) (k Overlay) (k Connect)
  where
    k _ x     Empty = x -- Constant folding to get rid of Empty leaves
    k _ Empty y     = y
    k f x y = f x y
```

* "foldgFusion" test if the fusion between foldg and an other function happen correctly
* "fmapFusion" test if the fusion between two fmap and an other function happen correctly
* "transposeRewriteBack" test that there is no impact on "normal" code (like a `transpose` alone)
* "induceRewriteBack" test that there is no impact on "normal" code (like a `fmap` alone)

```
benchmarking foldgFusion/foldg
time                 103.8 μs   (103.4 μs .. 104.3 μs)
[...]

benchmarking foldgFusion/foldg . induce
time                 110.1 μs   (109.8 μs .. 110.5 μs)
[...]

benchmarking foldgFusion/foldg . transpose
time                 100.2 μs   (100.1 μs .. 100.3 μs)
[...]

benchmarking foldgFusion/foldg . fmap
time                 97.71 μs   (97.56 μs .. 97.90 μs)
[...]

benchmarking fmapFusion/fmap
time                 224.7 μs   (224.5 μs .. 225.0 μs)
[...]

benchmarking fmapFusion/fmap . fmap
time                 224.5 μs   (224.2 μs .. 225.0 μs)
[...]

benchmarking transposeRewriteBack/transpose
time                 134.5 μs   (132.7 μs .. 137.2 μs)
[...]

benchmarking transposeRewriteBack/transpose'
time                 214.8 μs   (209.2 μs .. 219.5 μs)
[...]

benchmarking induceRewriteBack/induce
time                 383.6 μs   (381.7 μs .. 386.8 μs)
[...]

benchmarking induceRewriteBack/induce'
time                 376.3 μs   (372.4 μs .. 382.6 μs)
[...]
```

Note that this will not improve directly alga's benchmarks and will "only" improve user code, as displayed int the benchmarks.

There is certainly a lot of polishing to do, but the main idea is here :)


snowleopard:
Many thanks @nobrakal! The results look great.

You say these rules do not make Alga's code faster, but perhaps they can help us simplify some code? For example, `hasEdge`? Perhaps, the answer is "No", which is OK, but it would be great if the answer turned out to be "Yes" :)

snowleopard:
@nobrakal There now seem to be some merge conflicts after I merged your previous PR.

nobrakal:
> You say these rules do not make Alga's code faster, but perhaps they can help us simplify some code? For example, `hasEdge`? 

The old `hasEdge` was really poorly optimized and even with the rules, it is still behind your new version of `hasEdge` :)

Otherwise, I don't see any other function that can be simplified...

> There now seem to be some merge conflicts

Fixed :)

src/Algebra/Graph.hs
```diff
@@ -859,7 +871,7 @@ filterContext s i o g = maybe g go $ context (==s) g
 -- @
 {-# SPECIALISE replaceVertex :: Int -> Int -> Graph Int -> Graph Int #-}
 replaceVertex :: Eq a => a -> a -> Graph a -> Graph a
-replaceVertex u v = fmap $ \w -> if w == u then v else w
+replaceVertex u v = mapG $ \w -> if w == u then v else w
```
snowleopard: Can you keep `fmap` here? It will be rewritten to `mapG` anyway. I think `fmap` is easier to understand.

src/Algebra/Graph.hs
```diff
@@ -873,7 +885,7 @@ replaceVertex u v = fmap $ \w -> if w == u then v else w
 -- mergeVertices odd  1 (3 + 4 * 5) == 4 * 1
 -- @
 mergeVertices :: (a -> Bool) -> a -> Graph a -> Graph a
-mergeVertices p v = fmap $ \w -> if p w then v else w
+mergeVertices p v = mapG $ \w -> if p w then v else w
```
snowleopard: Same here.

src/Algebra/Graph.hs
```diff
@@ -1041,3 +1054,48 @@ sparsify graph = res
         m <- get
         put (m + 1)
         overlay <$> s `x` m <*> m `y` t
+
+buildG :: forall a. F.Fold a -> Graph a
+buildG g = F.foldg Empty Vertex Overlay Connect g
+{-# INLINE [1] buildG #-}
+
+-- | FB functions
```
snowleopard: Please add some comments for all these FB functions. Don't hesitate to write long comments: you can even add a part of your blog post here!

src/Algebra/Graph.hs
```diff
@@ -1041,3 +1054,48 @@ sparsify graph = res
...
+mapGFB :: (b -> c) -> (a -> b) -> a -> c
+mapGFB = (.)
+{-# INLINE [0] mapGFB #-}
+
+induceFB :: b -> (a -> b) -> (a -> Bool) -> a -> b
+induceFB e v p = \x -> if p x then v x else e
+{-# INLINE [0] induceFB #-}
+
+transposeFB :: (b -> b -> b) -> (b -> b -> b)
```
snowleopard: Wait, I thought we no longer need anything special for `transpose`, because it is written as a `foldg`. Instead, we could add rules that rewrite `foldg e _ _ _ Empty` into `e`, etc.

src/Algebra/Graph/Fold/Internal.hs~
```diff
@@ -0,0 +1,43 @@
+newtype Fold a = Fold { runFold :: forall b. b -> (a -> b) -> (b -> b -> b) -> (b -> b -> b) -> b }
```
snowleopard: I guess we don't need this file?

snowleopard:
@nobrakal Thanks! I left some some comments above. Also, could explain which import cycle you're trying to avoid? I would prefer not to introduce `Fold.Internal` unless it's absolutely necessary.

src/Algebra/Graph.hs
nobrakal: Yes, indeed "FB functions" are maybe not enough ^^

src/Algebra/Graph.hs
nobrakal: W need such function to "rewrite back" un-merged expression :)

Concerning the `foldg` rules, this is not so simple due to the `transpose . star` problem...

nobrakal:
The problem is that `Algebra.Graph.Fold` is based on `Algebra.Graph.ToGraph` which require `Algebra.Graph`. But I need only the `newtype` in `Algebra.Graph`. So the solution is to define the `newtype` in a internal module and make `ToGraph` and `Algebra.Graph` depend on it :)

snowleopard:
@nobrakal Aha, I see where the import cycle happens... Let me think about it for some time, perhaps, I'll come up with a way to refactor the module structure.

src/Algebra/Graph.hs
nobrakal: I added a modified version of the comment in `GHC.Base`: http://hackage.haskell.org/package/base-4.11.0.0/docs/src/GHC.Base.html#mapFB

src/Algebra/Graph.hs
snowleopard: My response ended up being too long, so I posted it in the comments below.

snowleopard:
@nobrakal Thank you for further explanations :) So, I can see now that the rewrite rules get stuck here:

```haskell
foldg Empty Vertex Overlay (flip Connect) $ vertices xs
```

But why can't we proceed as follows?

1) Inline `vertices`:

```haskell
foldg Empty Vertex Overlay (flip Connect) (overlays . map vertex $ xs)
```

2) Inline `overlays`:

```haskell
foldg Empty Vertex Overlay (flip Connect) (concatg overlay . map vertex $ xs)
```

3) Now we could continue with inlining `concatg`, and perhaps get stuck somewhere later, but one simple option would be to create two rewrite rules for `foldg` applied to the result of `concatg`:

```haskell
foldg e v o c (concatg Overlay list) = concatg o (map (foldg e v o c) list)
foldg e v o c (concatg Connect list) = concatg c (map (foldg e v o c) list)
```

4) If we apply the above rules we get:

```haskell
concatg Overlay (map (foldg Empty Vertex Overlay (flip Connect)) . map vertex $ xs)
```

5) We can now fuse the two `map`s:

```haskell
concatg Overlay (map (foldg Empty Vertex Overlay (flip Connect) . vertex) $ xs)
```

6) And after propagating `foldg` through `vertex`, we get what we wanted:

```haskell
concatg Overlay (map Vertex xs)
```

@nobrakal Can you make this work? :)

nobrakal:
> @nobrakal Can you make this work? :)

This sounded like a great idea, but I have spent my weekend on it and sadly I am not able to make it:

* First of all, we don't want to target `concatg` since we need it to fire quickly to allow good composition with a possible `build` following it (like in `vertices`).
* Even if we target `overlays`, we are stuck at:

```Haskell
concatg Overlay (map (foldg Empty Vertex Overlay (flip Connect)) . map vertex $ xs)
```
because the two maps won't be merged, since we are late in the simplifier phases when this happen.

* Even if we merge the two `map` by hand (remember, `concatg` is only `foldr` so it is pretty easy to do it), `GHC` will inline `foldg` (and its a good idea for specialization) in the laste phase, and we are stuck with the `go` function, which will never be inlined because it is recursive. (In fact, compiling with `-O2` solves some cases, but not all), and we canno't target it with rewrite-rules (since it is a local-definition in `foldg`).

Moreover, even if I was able to make such rules working, they will mess-up entirely the `buildG/transpose` work :p

This was not totally a waste of time, I found some improvements on the `buildG` code, and I learned how `inspection-testing` was working. I will add all of this in the beginning of the week :)

nobrakal:
Hi there,

I spoke too quickly yesterday, and almost everything is working with a bit of work :)

Mainly:

* No more specific rules for `transpose`. Instead general rules for `foldg`
* Compatibility with the `buildG` optimisation.
* Adding tests to ensure that everything is well rewritten with the great `inspection-testing` (in ` test/Algebra/Graph/Test/GraphRules.hs`. (it is very cool to see the rules working here). 

Some points on which I want to draw your attention:

* There still an issue with the `induce`: indeed it was always there: converting `induce` to its `buildg` equivalent is not... Equivalent: I cannot make the "`k-reduce`" trick, since I cannot pattern-match against function.
Indeed I still have an issue to "rewrite-it" back if fusion not happened.
* I added a compilation flag (`-fspec-constr`) normally enabled with `-O2` to make this work. Note that it seems that it improves your `hasEdge` (it canno't be from my optimization, since `hasEdge` is made without `foldg`).
* I modified a bit the `foldr1Safe` function, mainly not in-lining it too quickly, so I have to target it with specific rules. I think it is not very wise to use it outside the context of `Algebra.Graph`.

snowleopard:
@nobrakal Looks awesome!

I particularly like how inspection testing works! :) But I think we better introduce it in a separate PR, ideally before this one, and this PR will only add new rewrite rules tests.

> I added a compilation flag (-fspec-constr) normally enabled with -O2 to make this work. Note that it seems that it improves your hasEdge 

This is odd: according to your performance regression suite, `hasEdge` actually got worse: `1.27 (bad)`.

Two more questions:

1) Couldn't you use `concatg` in your `foldg/overlays` and `foldg/connects` rules? I think it would be a bit easier to understand.

2) I'm afraid I didn't understand your comment about `induce` being problematic. Could you clarify?

nobrakal:
>  But I think we better introduce it in a separate PR, ideally before this one, and this PR will only add new rewrite rules tests.

Ok I will send a PR soon :)

> This is odd: according to your performance regression suite, hasEdge actually got worse: 1.27 (bad).

That is really strange, on my proper `buildG` branch, I have `hasEdge: 0.68 (good)`... I will investigate ^^

> Couldn't you use concatg in your foldg/overlays and foldg/connects rules? I think it would be a bit easier to understand.

Sadly I cannot. In the expression:

```Haskell
foldg e v o c (overlays lst) = fromMaybe e (foldr (mf o . foldg e v o c) Nothing lst
```
If `foldg e v o c :: Graph a -> b` then `mf o . foldg e v o c` is of type `Graph a -> Maybe b -> Maybe b`, and thus canno't be used in `foldr1Safe`, thus not with `concatg`.

> I'm afraid I didn't understand your comment about induce being problematic. Could you clarify?

Of course, there is in fact two problems:

* I have trouble rewriting-back the `buildG` "equivalent" of `induce` into a plain `induce`.
* More problematic, the rule

```Haskell
 induce p g  = buildG (F.Fold $ \e v o c -> foldg e (induceFB e v p) o c g)
```

is not really great:

```
> let g = vertices [1..100]
> size $ induce (const False) g
1
>  size  $ buildG (F.Fold $ \e v o c -> foldg e (induceFB e v (const False)) o c g)
10
```

What is happening ? The `buildG` version cannot remove `empty` leaves, since it can not pattern-match against them.

nobrakal:
> In this case I think it may be useful to just remove concatg altogether. Minus one concept to keep in mind, plus more similarity between rewrite rules and definitions. What do you think?

Ok this is simpler you are right :)

> Regarding rewrite rules for induce: after we merge this PR

No problem!

I will try to solve other problems (the benchmarking suite and the rewriting-back of un-merged function) soon

nobrakal:
About benchmarks, the last one shows:

> hasEdge: 1.00 (OK)

After some tests on my machine with bigger graphs, it shows no differences between the two... I think they were false positive.

src/Algebra/Graph.hs
```diff
@@ -943,21 +966,8 @@ splitVertex v us g = g >>= \w -> if w == v then vertices us else vertex w
...
-"transpose/Overlay"  forall g1 g2. transpose (Overlay g1 g2) = Overlay (transpose g1) (transpose g2)
-"transpose/Connect"  forall g1 g2. transpose (Connect g1 g2) = Connect (transpose g2) (transpose g1)
-
-"transpose/overlays" forall xs. transpose (overlays xs) = overlays (map transpose xs)
-"transpose/connects" forall xs. transpose (connects xs) = connects (reverse (map transpose xs))
-
-"transpose/vertices" forall xs. transpose (vertices xs) = vertices xs
-"transpose/clique"   forall xs. transpose (clique xs)   = clique (reverse xs)
- #-}
+transpose = foldg Empty Vertex Overlay (transposeFB Connect)
```
snowleopard: I think `transposeFB` should be called `flipFB`.

src/Algebra/Graph.hs
```diff
@@ -1082,3 +1093,73 @@ sparsify graph = res
...
+
+  It's important that these two rules aren't both active at once
+  (along with build's unfolding) else we'd get an infinite loop
+  in the rules. Hence the activation control below.
+
+* FB functions are here to remember f after applying a "builG/f" rule.
+  These functions are higher-order functionsand therefore benefits from
+  inlining.
+
+* The "mapGFB" rule optimises compositions of mapG
```
snowleopard: Where does the suffix `FB` come from? Perhaps from `foldr` and `build`? I think it is worth adding a remark about this.

src/Algebra/Graph.hs
```diff
@@ -1082,3 +1093,73 @@ sparsify graph = res
...
+-}
+
+buildG :: forall a. F.Fold a -> Graph a
+buildG g = F.foldg Empty Vertex Overlay Connect g
```
snowleopard: We can drop `g` on both sides I guess.

src/Algebra/Graph.hs
```diff
@@ -1082,3 +1093,73 @@ sparsify graph = res
...
+{-# INLINE [1] buildG #-}
+
+-- | FB functions
+mapGFB :: (b -> c) -> (a -> b) -> a -> c
+mapGFB = (.)
+{-# INLINE [0] mapGFB #-}
+
+induceFB :: b -> (a -> b) -> (a -> Bool) -> a -> b
```
snowleopard: This function doesn't look like `induce` at all, so the name is a bit misleading.

snowleopard: Perhaps, `matchFP` or something similar would be better.

src/Algebra/Graph.hs
snowleopard: Or, maybe it's better to come up with a different suffix. I would just use `R` for all such intermediate functions, where `R` stands for "rewrite" and "rules" simultaneously :-)

src/Algebra/Graph/Fold.hs
```diff
@@ -23,7 +23,7 @@ module Algebra.Graph.Fold (
     Fold,
 
     -- * Basic graph construction primitives
-    empty, vertex, edge, overlay, connect, vertices, edges, overlays, connects,
+    empty, vertex, overlay, connect, edge, vertices, edges, overlays, connects,
```
snowleopard: Is this accidental?

snowleopard:
@nobrakal I've left a few more comments :)

src/Algebra/Graph/Fold.hs
nobrakal: Yes indeed ^^

src/Algebra/Graph.hs
nobrakal: I don't know at all from were come these `FB` functions ^^. These names come from [`GHC.List`](http://hackage.haskell.org/package/base-4.11.0.0/docs/src/GHC.List.html#foldl).

I am not sure changing these suffix since they are common in GHC sources...

nobrakal:
I addressed most of them, I am still facing an issue with `induce`, I will try to dig in Core soon :)

src/Algebra/Graph.hs
snowleopard: The problem with the suffix `FB` is precisely that we both don't know what it means. And other readers of this code will most probably not know either :) So, why don't we make it easier for the readers to understand what's going on by using `R` and saying that throughout the Alga codebase the suffix `R` indicates functions used exclusively for implementing rewrite rules?

nobrakal:
I changed `FB` functions to `R` functions :)

Otherwise, it was my benchmarking code that was faulty :p So everything seems to be in order for me !

snowleopard:
> Otherwise, it was my benchmarking code that was faulty

So, why was it showing a speedup for `hasEdge`?

By the way, I've merged the non-empty refactoring, so could you please make the necessary changes in the regression suite?

Also, sorry about the resulting merge conflicts.

src/Algebra/Graph.hs
```diff
@@ -176,7 +178,15 @@ data Graph a = Empty
              | Vertex a
              | Overlay (Graph a) (Graph a)
              | Connect (Graph a) (Graph a)
-             deriving (Functor, Show)
+             deriving (Show)
+
+instance Functor Graph where
+  fmap = mapG
```
snowleopard: Should this also be called `mapR` to highlight that it is there only for rewrite rules?

src/Algebra/Graph/Fold/Internal.hs
```diff
@@ -0,0 +1,287 @@
+{-# LANGUAGE RankNTypes #-}
+-----------------------------------------------------------------------------
+-- |
+-- Module     : Algebra.Graph.Fold
```
snowleopard: `.Internal` has been lost.

nobrakal:
> So, why was it showing a speedup for hasEdge?

I really don't know where this came from... I was not able to reproduce this on my machine so it is very strange

> By the way, I've merged the non-empty refactoring, so could you please make the necessary changes in the regression suite?

Of course !

src/Algebra/Graph.hs
```diff
@@ -178,7 +180,15 @@ data Graph a = Empty
              | Vertex a
              | Overlay (Graph a) (Graph a)
              | Connect (Graph a) (Graph a)
-             deriving (Functor, Show)
+             deriving (Show)
+
```
snowleopard: Let's add the following note here:

```suggestion
{- Note [Functions for rewrite rules]
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

This module contains several functions whose only purpose is to guide GHC
rewrite rules. The names of all such functions are suffixed with "R" so that it
is easier to distinguish them from others. 

Why do we need them?

These functions are annotated with carefully chosen GHC pragmas that control
inlining, which would be impossible or unreliable if we used standard functions
instead. For example, the function 'eqR' has the following annotations:

{-# NOINLINE [1] eqR #-}
{-# RULES "eqIntR" eqR = eqIntR #-}

This tells GHC to rewrite 'eqR' to faster 'eqIntR' if possible (if the types
match), and -- importantly -- not to inline 'eqR' too early, before the rewrite
rule had a chance to fire.

We could have written the following rule instead:

{-# RULES "eqIntR" (==) = eqIntR #-}

But that would have to rely on appropriate inlining behaviour of (==) which is
not under our control. We therefore choose the safe and more explicit path of
creating our own intermediate functions for guiding rewrite rules when needed.
-}
```

Also, let's see how the new GitHub Suggestions feature works :-)

src/Algebra/Graph.hs
snowleopard: See the note I've added above. I think it is just simpler to choose one consistent naming convention and follow it throughout the code. So, let's add `R` everywhere: `eqR`, `eqIntR`, etc. 

> and more prosaically I already have a `mapGR` used

Let's not invent new names. The above, for example, should be `fmapR` (it's a substitute for `fmap`). And the function you call `mapGR` should probably be more appropriately called `composeR`:

```
composeR :: (b -> c) -> (a -> b) -> a -> c
composeR = (.)
```

I find `mapGR = (.)` to be quite confusing.

nobrakal: Ok, indeed it can be more simple to understand for anyone reading the code

src/Algebra/Graph.hs
nobrakal: Great feature :D It seems to work well !

snowleopard: Indeed, pretty cool! The only slight problem is that it looks like it replaces the line I clicked on, so this removed the empty line between `deriving (Show)` and the comment :-)

nobrakal: Haha, I just added it, and removed one that bothered me for a while

src/Algebra/Graph.hs
```diff
@@ -178,7 +180,43 @@ data Graph a = Empty
...
+We could have written the following rule instead:
+
+{-# RULES "eqIntR" (==) = eqIntR #-}
+
+But that would have to rely on appropriate inlining behaviour of (==) which is
+not under our control. We therefore choose the safe and more explicit path of
+creating our own intermediate functions for guiding rewrite rules when needed.
+-}
+instance Functor Graph where
```
snowleopard: Let's also add an empty line here, since it may trigger Haddock to attach the above comment to the Functor instance.

src/Algebra/Graph.hs
```diff
@@ -178,7 +180,43 @@ data Graph a = Empty
...
+  fmap = fmapR
+
+-- | 'fmap' on which we can apply rewrite rules
```
snowleopard: Another simple convention I recently started to follow is not to use `|` when commenting functions that are not exported. This makes it easy to see whether a function is exported or not, without checking the top of the module.

snowleopard:
@nobrakal Could you also rename functions like `eq` and `ord` to `eqR` etc., as otherwise my note doesn't match the sources :-)

src/Algebra/Graph.hs
nobrakal: Yes this is a good idea

nobrakal:
Haha, I get lost between all these names. Normally this is now coherent !

src/Algebra/Graph.hs
```diff
@@ -1066,3 +1105,73 @@ sparsify graph = res
...
+  in the rules. Hence the activation control below.
+
+* R functions (R for Rewrite) are here to remember f after applying a
+  "builG/f" rule. These functions are higher-order functions and therefore
+  benefits from inlining.
+
+* The "mapR/mapR" rule optimises compositions of fmapR
+-}
+
+buildG :: forall a. F.Fold a -> Graph a
```
snowleopard: You seem to have missed this function: it should be `buildR`. Also in the above comment.

src/Algebra/Graph.hs
```diff
@@ -1066,3 +1105,73 @@ sparsify graph = res
...
+-- Transform a fmapR into its build equivalent
+"buildG/map"       [~1] forall f g.
+  fmapR   f g  = buildG (F.Fold $ \e v o c -> foldg e (composeR v f) o c g)
+
+-- Transform a induce into its build equivalent
+"buildG/induce"    [~1] forall p g.
+  induce p g  = buildG (F.Fold $ \e v o c -> foldg e (matchR e v p) o c g)
+
+"buildG/transpose" [~1] forall g.
+  foldg Empty Vertex Overlay (flipR Connect) g = buildG (F.Fold $ \e v o c -> foldg e v o (flip c) g)
```
snowleopard: I wonder if it makes sense to make this rule more general and get rid of `flipR` by doing so:

```
"buildG/transpose" [~1] forall g.
  foldg Empty Vertex Overlay (f Connect) g = buildG (F.Fold $ \e v o c -> foldg e v o (f c) g)
```
Will this work?

snowleopard: Also, perhaps we could also add a similar general rule for `Overlay`:

```
"buildG/transpose" [~1] forall g.
  foldg Empty Vertex (f Overlay) Connect g = buildG (F.Fold $ \e v o c -> foldg e v (f o) c g)
```

Note that `induce` actually applies function `k` to `Overlay` and `Connect` and could therefore be subject to these rules. Is it possible that by adding these general rules, you can get rid of the special rule for `induce`?

src/Algebra/Graph.hs
```diff
@@ -1066,3 +1105,73 @@ sparsify graph = res
...
+ #-}
+
+-- Rules to merge rewrited functions
+{-# RULES
+-- Merge a foldg followed by a buildG
+"foldg/buildG" forall e v o c (g::F.Fold a).
+                              foldg e v o c (buildG g) = F.foldg e v o c g
+
+-- Merge two mapR
```
snowleopard: I don't understand this comment and/or rule. It doesn't merge two maps, it reduces a 3-way function composition to a 2-way function composition.

snowleopard: (I think the rule is fine, is just a misleading comment and name of the rule.)

snowleopard:
@nobrakal Some more comments! This is a complex PR and I want to understand every single detail in it :)

src/Algebra/Graph.hs
nobrakal: I will update this

src/Algebra/Graph.hs
snowleopard: Thinking about this again, I think my suggestion won't work because if we use `flip` instead of `flipR` then `flip` will probably inline too early, resulting in `\x y -> Connect y x` or something like this and the rule will no longer match... @nobrakal Could you please try and see what actually happens?

snowleopard: Perhaps, we could keep `flipR`, but still generalise the rule. Then it should work, and furthermore it will handle not only `flipR` but other functions too.

nobrakal: You made me re-think some parts :) , so mainly:

> get rid of flipR by doing so:

We can write (with some rankntypes hack) what you wrote, but this only remove `flipR` from the rule, not from the `transpose` definition :)

The similar for `overlay` is also acceptable, but not usable for `induce`: we need the full-polymorphism of the "k" function, which we don't have since there is a pattern-match inside.

I think this is solved using `Maybe (NonEmptyGraph)` since it uses `Nothing` in place of `Empty`

nobrakal: Note that having a fully polymorphic `k` will remove the need to rewrite `induce` back, and thus will simplify a bit all of this !

src/Algebra/Graph.hs
```diff
@@ -1066,3 +1105,79 @@ sparsify graph = res
...
+"buildR/induce"    [~1] forall p g.
+  induce p g = buildR (F.Fold $ \e v o c -> foldg e (matchR e v p) o c g)
+
+"buildR/foldg(fc)" [~1] forall (f::forall b. (b -> b -> b) -> (b -> b -> b))  g.
+  foldg Empty Vertex Overlay (f Connect) g = buildR (F.Fold $ \e v o c -> foldg e v o (f c) g)
+
+"buildR/foldg(fo)" [~1] forall (f::forall b. (b -> b -> b) -> (b -> b -> b))  g.
+  foldg Empty Vertex (f Overlay) Connect g = buildR (F.Fold $ \e v o c -> foldg e v (f o) c g)
+
+"buildR/foldg(fo)(hc)" [~1] forall (f::forall b. (b -> b -> b) -> (b -> b -> b)) (h::forall b. (b -> b -> b) -> (b -> b -> b)) g.
```
snowleopard: I think we don't need this two-way rule, because it can be achieved by independent applications of the two rules above.

nobrakal: I don't think so, imagine you have

```Haskell
foldg Empty Vertex (flipR Overlay) (flipR Connect)
```

None of the two single rules are matching with it...

snowleopard: Oh, you are right. Not sure what I was thinking :)

snowleopard:
@nobrakal So, I think you've addressed all my comments, but our plan was to first merge the inspection testing patch, right?

nobrakal:
@snowleopard I have corrected the merge.

I just have an issue with `inspection-testing` (see https://github.com/nomeata/inspection-testing/issues/23 ). I am waiting an answer :)

snowleopard:
@nobrakal Great, thank you for further fixes.

I have one further suggestion related to `Fold`. Although it was initially my suggestion to reuse the definition of `Fold` from `Algebra.Graph.Fold`, I think the cost of this reuse is too high: we had to split the module in two, and complicate the dependency structure.

Let's revert to a simpler solution, which I think you had initially: just define something like:

```haskell
type Foldg a = forall b. b -> (a -> b) -> (b -> b -> b) -> (b -> b -> b) -> b
```

And use it instead of `F.Fold`. 

Then in your future PR for non-empty graphs, you could follow a similar approach with

```haskell
type Foldg1 a = forall b. (a -> b) -> (b -> b -> b) -> (b -> b -> b) -> b
```

What do you think?

src/Algebra/Graph.hs
```diff
@@ -1066,3 +1104,77 @@ sparsify graph = res
...
+* composeR and matchR are here to remember the original function after
+  applying a "builR/f" rule. These functions are higher-order functions
+  and therefore benefits from inlining in final phase.
+
+* The "fmapR/fmapR" rule optimises compositions of fmapR
+-}
+
+type Foldg a = forall b. b -> (a -> b) -> (b -> b -> b) -> (b -> b -> b) -> b
+
+buildR :: forall a. Foldg a -> Graph a
```
snowleopard: I guess the `forall a.` part is redundant?

stack.yaml
```diff
@@ -4,4 +4,4 @@ packages:
 - '.'
 
 extra-deps:
-- inspection-testing-0.4
+- inspection-testing-0.4
```
snowleopard: This seems to be accidental.

snowleopard:
@nobrakal Great! Some final remarks above. Please fix them and I'll merge.

src/Algebra/Graph.hs
nobrakal: Oops yes

nobrakal:
@snowleopard  Done, sorry for these minor unwanted changes.

The Travis build is failing to build, but it seems that this is due to Travis.

snowleopard:
@nobrakal Thanks for fixing! I'm ready to merge, but at the moment the regression suite seems to be broken, so I can't see whether there are any performance changes. I guess we do not expect any, right?

nobrakal:
Oops I didn't realized that the benchmarking suite was broken, I just fixed that (it was due to your changes in `AdjacencyIntMap`).

> I guess we do not expect any, right?

Normally, not any yes. After a quick check, I didn't see any regression !

snowleopard:
@nobrakal Thanks! This is now, finally, merged :-) 

There is still this mysterious result: `hasEdge: 0.69 (good)`. If we believe the performance testsuite, every commit makes `hasEdge` faster and faster :-)

PR #124, opened 2018-10-06
Various updates: bump version and LTS, update change log and CI

snowleopard:
I've bumped the next version to `0.3` since we are breaking API, updated change log, and LTS version.

Also, I've switched the performance regression suite to GHC 8.6.1.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,40 +32,37 @@ before_cache:
 matrix:
   include:
     - compiler: "ghc-8.6.1"
-      env: GHCHEAD=true
-      addons: {apt: {packages: [ghc-ppa-tools,cabal-install-head,ghc-8.6.1], sources: [hvr-ghc]}}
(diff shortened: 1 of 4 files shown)


snowleopard:
@nobrakal I merged this PR in the hope that this is just a glitch, but it looks like I broke your performance suite. Hope you'll figure out how to fix it!

nobrakal:
@snowleopard this is fixed now, it was the bound of `algebraic-graphs` that changed, and the benchmarking suite was not building anymore :)

nobrakal:
Mhh I spoke too quickly,  the new GHC certainly broke some dependencies, I will investigate further :)

snowleopard:
@nobrakal Thanks! At least it's good to know that the problem is not with this PR. Thank you for looking into this!

nobrakal:
Now this is I hope fixed, I also had to update the upper bound of the base library ^^

snowleopard:
Thanks again!

Issue #128, opened 2018-10-16
SCC condensation doesn't distinguish cyclic and acyclic single-node components


michaelpj:
This can be very important. `Data.Graph` [does it](http://hackage.haskell.org/package/containers-0.6.0.1/docs/Data-Graph.html#t:SCC), it would be great if `alga` did too.

snowleopard:
@michaelpj Indeed! I propose to modify the API with the following goals in mind:

1. Address this issue, i.e. make it easy to find out whether a component is acyclic.
2. Get rid of the invariant that the sets (i.e. vertices) in the resulting condensed graph are non-empty.
3. (Optional) Make it easy to extract the graphs corresponding to components, as discussed in #105.

One option would be to use the `SCC` data type defined in `Data.Graph`, i.e. return `AdjacencyMap (SCC a)` from the function `scc`, but it's a bit awkward for two reasons: (i) `SCC` doesn't have an `Ord` instance, and (ii) it also comes with the invariant that the `CyclicSCC` list is non-empty.

Perhaps, we should define our own `data SCC a = AcyclicSCC a | CyclicSCC (NonEmpty a)` with an `Ord` instance? 

This is still not ideal. I would prefer to have a **non-empty set** in the `CyclicSCC` constructor, but alas such sets are not provided by the `containers` library.

Thinking of (3), we could also switch to something like:

```haskell
scc :: Ord a => AdjacencyMap a -> AdjacencyMap (NonEmptyGraph a)
```

Which I think ticks all the boxes, since we could then use `isAcyclic` on the resulting graphs.

Any other ideas?

michaelpj:
I think modelling SCCs as non empty graphs would be great, since it even preserves the inner graph structure of the component if you want it.

More generally, if you had a graph model that includes subgraphs, you could model them as subgraphs.

snowleopard:
Agree, this looks like the best option. I will give it a try in the next couple of days.

snowleopard:
I've finally finished #130. It took a long time, because it triggered a wave of refactoring (#131, #136). 

The resulting type signature is:
```haskell
import           Algebra.Graph.AdjacencyMap
import qualified Algebra.Graph.NonEmpty.AdjacencyMap as NonEmpty

scc :: Ord a => AdjacencyMap a -> AdjacencyMap (NonEmpty.AdjacencyMap a)
```

[...]

Issue #126, opened 2018-10-14
Ord instance for graphs


snowleopard:
While playing with [QuickSpec](http://hackage.haskell.org/package/quickspec) to see if it can discover laws of algebraic graphs (yes, it can), I needed to add `Ord a => Ord (Graph a)` instance. My implementation simply redirected to the `Ord` of the `AdjacencyMap` representation, but I'm not sure if this is the best/most convenient way to define an ordering on graphs.

Any ideas how `Ord` should work?

nobrakal:
The problem is that an `Ord` instance is required to satisfy:

```Haskell
x == y = compare x y == EQ
```

So I don't think that this is possible to implement a simple thing without `AdjacencyMap`, otherwise you will be able to implement graph equality.

The only problem I see with an `Ord` instance from `AdjacencyMap` is that its order not be very intuitive, because based on the order of the adjacency lists, itself based on the order of the vertices.

Hence we will have (in an `AdjacencyMap` context):

```Haskell
2*3 < 2 == False
2*3 < 3 == True
```

Or more generally,  the single vertex `a` will always be superior to any graph constructed with a least a vertex `<a`

snowleopard:
@nobrakal One option is to insist that the `Ord` instance refines the `isSubgraphOf` relation. In other words:
* If ``x `isSubgraphOf` y`` then `x <= y`.
* As a consequence, this forces `Ord` to agree with the current `Eq` instance.

Note that this tells us that for your examples we have:

* `2 < 2*3 == True`
* `3 < 2*3 == True`

Which seems to be better.

But it leaves it open the following cases:

1. `1 < 2*3`
2. `4 < 2*3`
3. `1*2 < 2*3`
4. `2*2 < 2*3`
5. `3*4 < 2*3`

If we fall back to `AdjacencyMap`'s `Ord` instance after first checking `isSubgraphOf` then I think we will get the following:

1. `1 < 2*3 = True`
2. `4 < 2*3 = True`
3. `1*2 < 2*3 = True`
4. `2*2 < 2*3 = True`
5. `3*4 < 2*3 = False`

Not sure this is particularly intuitive...

snowleopard:
I just had a chat with Victor Khomenko (@vkhomenko) and he suggested to use "size-lexicographical" orders on vertices and edges, as follows:

* Compare the size of graphs in terms of the number of vertices. In case of a tie continue.
[...]

PR #130, opened 2018-10-23
Add a draft implementation of new SCC algorithm

snowleopard:
See #128.

The draft implementation seems to work correctly, but I need to add tests and documentation.

For example, given a graph `edges [(1,2), (2,3), (3,4), (4,3), (2,1)]`, the new `scc` finds a graph with two vertices `1*2 + 2*1` and `3*4 + 4*3` of type `NonEmptyGraph Int` and one edge between them.

Before merging this PR, I will need to address #126 (since we need an `Ord` instance for `NonEmptyGraph`s to store them inside an `AdjacencyMap`).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,9 @@
 
 ## 0.3
 
+* #128, #130: Modify the SCC algorithm to return non-empty graph components.
+* #130: Move adjacency map algorithms to separate modules.
+* #130: Export `fromAdjacencySets` and `fromAdjacencyIntSets`.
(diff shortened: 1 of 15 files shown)


snowleopard:
Oh, what a mess. Thank you, git!

snowleopard:
Had to do a force push to clean up the mess. Apologies if anyone was on this branch.

snowleopard:
@nobrakal Apologies, looks like I've broken the performance suite, again :(

Issue #129, opened 2018-10-17
Test rewrite rules with `inspection-testing`?


michaelpj:
Just a drive-by thought since I noticed that you have some rewrite rules now. (inspection-testing)[https://github.com/nomeata/inspection-testing] is a nice library for testing that these are actually firing as you intend them to, without relying on inspecting benchmarks.

snowleopard:
Good idea! @nobrakal and I wanted to test rewrite rules before but didn't know about this library then.

@nobrakal Have a look at this approach, it might help you a lot with your on-going work on rewrite rules.

There is a nice Haskell eXchange talk about inspection testing:

https://skillsmatter.com/skillscasts/12590-a-promise-checked-is-a-promise-kept-inspection-testing

nobrakal:
This seems geat :D 

Indeed it can really help me testing rule firings, and as @michaelpj  pointed it out, it seems that this library can simply test that all is working good in alga's test-suite.

I definitely need to dig on this, thanks for the link !

snowleopard:
@nobrakal After merging your draft implementation of inspection testing, my `ghcid` setup is no longer working: the RewriteRules testsuite reports errors, I guess because `ghcid` compiles with no optimisation... Do you know if it's possible to switch off inspection testing with a flag somehow?

PR #132, opened 2018-10-24
Add rewrite tests

nobrakal:
This is related to #129 and #123 

It implements tests of rewrite rules of `Algebra.Graph` using the (fantastic) `inspection-testing`.

* The correct `foldr/map` merge of `overlays . map vertex`
* The transpose rules
* Plus the nontrivial two combined: the rewriting of `transpose . star`

I don't really know if want to have such tests for other modules, they can obviously be added :)

### DIFF
--- a/src/Algebra/Graph/Internal.hs
+++ b/src/Algebra/Graph/Internal.hs
@@ -21,7 +21,7 @@ module Algebra.Graph.Internal (
 
     -- * Graph traversal
     Focus (..), emptyFocus, vertexFocus, overlayFoci, connectFoci, Hit (..),
-    foldr1Safe,
+    foldr1Safe, mf,
 
     -- * Utilities
     setProduct, setProductWith
@@ -116,11 +116,12 @@ data Hit = Miss | Tail | Edge deriving (Eq, Ord)
 
 -- | A safe version of 'foldr1'.
 foldr1Safe :: (a -> a -> a) -> [a] -> Maybe a
-foldr1Safe f = foldr mf Nothing
-  where
-    mf x m = Just (case m of
-                        Nothing -> x
-                        Just y  -> f x y)
+foldr1Safe f = foldr (mf f) Nothing
+
+mf :: (a -> a -> a) ->  a -> Maybe a -> Maybe a
+mf f x m = Just (case m of
+                  Nothing -> x
+                  Just y  -> f x y)
 {-# INLINE foldr1Safe #-}
 
 -- | Compute the Cartesian product of two sets.
--- a/test/Algebra/Graph/Test/GraphRules.hs
+++ b/test/Algebra/Graph/Test/GraphRules.hs
@@ -1,0 +1,44 @@
+{-# LANGUAGE TemplateHaskell #-}
+-----------------------------------------------------------------------------
+-- |
+-- Module     : Algebra.Graph.Test.Graph
+-- Copyright  : (c) Andrey Mokhov 2016-2018
+-- License    : MIT (see the file LICENSE)
+-- Maintainer : andrey.mokhov@gmail.com
+-- Stability  : experimental
+--
+-- Testsuite for "Algebra.Graph" rewrite-rules
+-----------------------------------------------------------------------------
+module Algebra.Graph.Test.GraphRules where
+
+import Algebra.Graph
+
+import qualified Test.Inspection as I
+
+-- transpose tests
+--- transpose . star
+starTranspose, transposeDotStar :: a -> [a] -> Graph a
+starTranspose a [] = vertex a
+starTranspose a xs = connect (vertices xs) (vertex a)
+
+transposeDotStar x = transpose . star x
+
+I.inspect $ 'starTranspose I.=== 'transposeDotStar
+
+--- transpose . overlays
+overlays', transposeDotOverlays :: [Graph a] -> Graph a
+overlays'            = overlays . map transpose
+
+transposeDotOverlays = transpose . overlays
+
+I.inspect $ 'overlays' I.=== 'transposeDotOverlays
+
+{-
+--- transpose . vertices
+vertices', transposeDotVertices :: [a] -> Graph a
+vertices'            = overlays . map vertex
+
+transposeDotVertices = transpose . vertices
+
+I.inspect $ 'vertices' I.=== 'transposeDotVertices
+-}
--- a/test/Algebra/Graph/Test/GraphRules.hs
+++ b/test/Algebra/Graph/Test/GraphRules.hs
@@ -11,9 +11,19 @@
 -----------------------------------------------------------------------------
 module Algebra.Graph.Test.GraphRules where
 
+import Data.Maybe (fromMaybe)
+
 import Algebra.Graph
+import Algebra.Graph.Internal
 
 import qualified Test.Inspection as I
+
+-- overlays tests
+vertices', overlaysDotMapVertex :: [a] -> Graph a
+vertices'            = fromMaybe Empty . foldr (mf Overlay . Vertex) Nothing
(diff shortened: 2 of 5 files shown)


src/Algebra/Graph/Internal.hs
```diff
@@ -116,11 +116,12 @@ data Hit = Miss | Tail | Edge deriving (Eq, Ord)
...
 -- | A safe version of 'foldr1'.
 foldr1Safe :: (a -> a -> a) -> [a] -> Maybe a
-foldr1Safe f = foldr mf Nothing
-  where
-    mf x m = Just (case m of
-                        Nothing -> x
-                        Just y  -> f x y)
+foldr1Safe f = foldr (mf f) Nothing
+
+mf :: (a -> a -> a) ->  a -> Maybe a -> Maybe a
```
snowleopard: Let's make the type more general `(a -> b -> a) -> a -> Maybe b -> Maybe a`, as it makes it a bit easier to understand what the function does. Also, I'd prefer to have a more meaningful name, and a comment :)

snowleopard: Interesting: this type signature looks quite similar to `scanl`:

https://hackage.haskell.org/package/base-4.12.0.0/docs/Prelude.html#v:scanl

test/Algebra/Graph/Test/GraphRules.hs
```diff
@@ -0,0 +1,99 @@
...
+-- |
+-- Module     : Algebra.Graph.Test.Graph
+-- Copyright  : (c) Andrey Mokhov 2016-2018
+-- License    : MIT (see the file LICENSE)
+-- Maintainer : andrey.mokhov@gmail.com
+-- Stability  : experimental
+--
+-- Testsuite for "Algebra.Graph" rewrite-rules
+-----------------------------------------------------------------------------
+module Algebra.Graph.Test.GraphRules where
```
snowleopard: Let's call the module "RewriteRules"? I think it's a bit clearer.

test/Algebra/Graph/Test/GraphRules.hs
```diff
@@ -0,0 +1,99 @@
...
+
+import Data.Maybe (fromMaybe)
+
+import Algebra.Graph
+import Algebra.Graph.Internal
+
+import qualified Test.Inspection as I
```
snowleopard: Let's import non-qualified to reduce some `I.` noise?

test/Algebra/Graph/Test/GraphRules.hs
```diff
@@ -0,0 +1,99 @@
...
+
+-- overlays/connects tests
+--- overlays . map vertex
+vertices', overlaysDotMapVertex :: [a] -> Graph a
+vertices'            = fromMaybe Empty . foldr (mf Overlay . Vertex) Nothing
```
snowleopard: Why are you testing a function defined here instead of the actual library `vertices`?

snowleopard: Same for `clique`, `empty`, etc.

snowleopard:
@nobrakal This is pretty cool! Much better than what we discussed some time ago :)

I've left some comments above. 

Shall we also test that we always rewrite everything to use `AdjacencyIntMap` for `Int` graphs?

src/Algebra/Graph/Internal.hs
nobrakal: Ok, this done, with changing its name to `maybeF`, and a better definition :)

test/Algebra/Graph/Test/GraphRules.hs
nobrakal: Done

test/Algebra/Graph/Test/GraphRules.hs
nobrakal: I wanted to avoid the clash with `(===)` from `Algebra.Graph`, but hiding it is simpler :)

test/Algebra/Graph/Test/GraphRules.hs
nobrakal: Because otherwise the tests are failing :) It seems that with the declaration `f = g`, `g` is not inlined, even if it has an `INLINE` pragma...

nobrakal: I think this is because we are working outside the module itself, but I find a bit messy to have these tests included directly in `src/Algebra/Graph.hs`...

nobrakal:
> Much better than what we discussed some time ago :)

Indeed ^^

> Shall we also test that we always rewrite everything to use AdjacencyIntMap for Int graphs?

It would be great, but these tests are failing because the rewrited part seems to not be inlined... For example with:

```Haskell
eqG, eqInt :: Graph Int -> Graph Int -> Bool
eqG = (==)
eqInt x y =  adjacencyIntMap x == adjacencyIntMap y

inspect $ 'eqG === 'eqInt
```

I have

```
test/Algebra/Graph/Test/RewriteRules.hs:106:1: eqG === eqInt failed:
    LHS:    
        eqG :: Graph Int -> Graph Int -> Bool
        [LclIdX,
         Arity=2,
         Str=<S,1*U><S,1*U>,
         Unf=Unf{Src=InlineStable, TopLvl=True, Value=True, ConLike=True,
                 WorkFree=True, Expandable=True,
                 Guidance=ALWAYS_IF(arity=0,unsat_ok=True,boring_ok=True)
                 Tmpl= equalsInt}]
        eqG = equalsInt
            
    RHS:    
        eqInt :: Graph Int -> Graph Int -> Bool
        [LclIdX,
         Arity=2,
         Str=<S,1*U><S,1*U>,
         Unf=Unf{Src=InlineStable, TopLvl=True, Value=True, ConLike=True,
                 WorkFree=True, Expandable=True,
                 Guidance=ALWAYS_IF(arity=2,unsat_ok=True,boring_ok=False)
                 Tmpl= \ (x [Occ=Once] :: Graph Int) (y [Occ=Once] :: Graph Int) ->
                         equal
                           @ IntSet
                           $fEqIntSet
                           (($fEqGraph_go x)
                            `cast` (N:AdjacencyIntMap[0]
                                    :: Coercible AdjacencyIntMap (IntMap IntSet)))
                           (($fEqGraph_go y)
                            `cast` (N:AdjacencyIntMap[0]
                                    :: Coercible AdjacencyIntMap (IntMap IntSet)))}]
        eqInt
          = \ (x [Dmd=<S,1*U>] :: Graph Int)
              (y [Dmd=<S,1*U>] :: Graph Int) ->
              equal
                @ IntSet
                $fEqIntSet
                (($fEqGraph_go x)
                 `cast` (N:AdjacencyIntMap[0]
                         :: Coercible AdjacencyIntMap (IntMap IntSet)))
                (($fEqGraph_go y)
                 `cast` (N:AdjacencyIntMap[0]
                         :: Coercible AdjacencyIntMap (IntMap IntSet)))
```

Otherwise, `inspection-testing` can test if a function have dictionaries (ie. is not specialized) so we can test `SPECIALIZE` pragma too :)

snowleopard:
@nobrakal I am fine to limit this PR to just rewrite rules. We'll tackle specialisation later. 

However, I'm still a bit unsatisfied with how the module looks. There are 99 lines for just 11 tests. Do we have to write 5-6 lines of boilerplate code for every test?

snowleopard:
@nobrakal Yep, here is another example:

https://github.com/kcsongor/generic-lens/blob/master/test/Spec.hs

Perhaps, it's just not possible to have a less verbose version...

snowleopard:
@nobrakal Is this ready to be merged? I think apart from the verbosity issue, I don't have any other comments. We can merge and refactor later.

nobrakal:
All seems ok for me, as you said, this can always be polished later !

PR #133, opened 2018-10-25
Labelled.AdjacencyMap.replaceEdge

Avasil:
Hello, I was playing with `LabelledAdjacencyMap` and needed this function. 
Maybe it will be useful!

### DIFF
--- a/src/Algebra/Graph/Labelled/AdjacencyMap.hs
+++ b/src/Algebra/Graph/Labelled/AdjacencyMap.hs
@@ -217,10 +217,10 @@ replaceVertex u v = gmap $ \w -> if w == u then v else w
 -- @
 replaceEdge :: (Eq e, Monoid e, Ord a) => e -> a -> a -> AdjacencyMap e a -> AdjacencyMap e a
 replaceEdge e x y
-  | e == zero  = AM . createVertex . Map.alter (Just . maybe Map.empty (Map.delete y)) x . adjacencyMap
-  | otherwise  = AM . createVertex . Map.alter replace x . adjacencyMap
+  | e == zero  = AM . createVertexY . Map.alter (Just . maybe Map.empty (Map.delete y)) x . adjacencyMap
+  | otherwise  = AM . createVertexY . Map.alter replace x . adjacencyMap
     where
-      createVertex     = Map.alter (Just . fromMaybe Map.empty) y
+      createVertexY    = Map.alter (Just . fromMaybe Map.empty) y
       replace (Just m) = Just $ Map.insert y e m
       replace Nothing  = Just $ Map.singleton y e
 
--- a/src/Algebra/Graph/Labelled/AdjacencyMap.hs
+++ b/src/Algebra/Graph/Labelled/AdjacencyMap.hs
@@ -209,11 +209,18 @@ replaceVertex u v = gmap $ \w -> if w == u then v else w
 
 -- | Replace an edge from a given graph. If it doesn't exist, it will be created.
 -- Complexity: /O(log(n))/ time.
+--
+-- @
+-- replaceEdge e x y m                 == overlay (removeEdge x y m) (edge e x y)
+-- replaceEdge e2 x y (edge e1 x y)    == edge e2 x y
+-- edgeLabel x y (replaceEdge e x y m) == e
+-- @
 replaceEdge :: (Eq e, Monoid e, Ord a) => e -> a -> a -> AdjacencyMap e a -> AdjacencyMap e a
 replaceEdge e x y
-  | e == zero  = removeEdge x y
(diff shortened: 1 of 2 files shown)


test/Algebra/Graph/Test/Labelled/AdjacencyMap.hs
```diff
@@ -29,3 +29,11 @@ testLabelledAdjacencyMap = do
...
     test "Consistency of fromAdjacencyMaps" $ \xs ->
         consistent (fromAdjacencyMaps xs :: LAI)
+
+    putStrLn "\n============ Labelled.AdjacencyMap.replaceEdge ============"
+
+    test "edgeLabel x y (replaceEdge e x y m) == e" $ \(e :: Sum Int) (x :: Int) (y :: Int) m ->
+          edgeLabel x y (replaceEdge e x y m) == e
+
+    test "replaceEdge e2 x y (edge e1 x y) == edge e2 x y" $ \(e1 :: Sum Int) (e2 :: Sum Int) (x :: Int) (y :: Int) ->
+          replaceEdge e2 x y (edge e1 x y) == edge e2 x y
```
snowleopard: Let's add these test cases as examples into the comment?

Here is an example:

https://hackage.haskell.org/package/algebraic-graphs-0.2/docs/Algebra-Graph.html#v:replaceVertex

snowleopard:
@Avasil Looks great, thank you! I've just left one comment above.

I also suggest to add one more property to the comment and tests:
```
replaceEdge e x y m = overlay (removeEdge x y m) (edge e x y)
```
This property almost acts as a definition of what we want from this function. We could even use it as an implementation, but I think yours is going to be faster.

src/Algebra/Graph/Labelled/AdjacencyMap.hs
```diff
@@ -207,6 +207,23 @@ removeEdge x y = AM . Map.adjust (Map.delete y) x . adjacencyMap
...
+--
+-- @
+-- replaceEdge e x y m                 == overlay (removeEdge x y m) (edge e x y)
+-- replaceEdge e2 x y (edge e1 x y)    == edge e2 x y
+-- edgeLabel x y (replaceEdge e x y m) == e
+-- @
+replaceEdge :: (Eq e, Monoid e, Ord a) => e -> a -> a -> AdjacencyMap e a -> AdjacencyMap e a
+replaceEdge e x y
+  | e == zero  = AM . createVertex . Map.alter (Just . maybe Map.empty (Map.delete y)) x . adjacencyMap
+  | otherwise  = AM . createVertex . Map.alter replace x . adjacencyMap
```
snowleopard: I don't think you need `createVertex` here: 
* If `x` is not in the map, you'll add both `x` and `y` in the `replace Nothing` case.
* If `y` is not in the map, you'll add it in `Map.insert y e m`.

src/Algebra/Graph/Labelled/AdjacencyMap.hs
```diff
@@ -207,6 +207,23 @@ removeEdge x y = AM . Map.adjust (Map.delete y) x . adjacencyMap
...
+    where
+      createVertex     = Map.alter (Just . fromMaybe Map.empty) y
```
snowleopard: A bit more precise name for this function is `createVertexY`.

snowleopard:
> Do you think it's still going to be faster or should we benchmark it?

I think yours is still faster, but if you like, you can try to benchmark using @nobrakal's benchmarking suite:  https://github.com/haskell-perf/graphs. It would be interesting to see how these two implementations compare.

src/Algebra/Graph/Labelled/AdjacencyMap.hs
Avasil: Without `createVertex` it fails test on 
`replaceEdge e x y m == overlay (removeEdge x y m) (edge e x y)`
with input `Sum 1`, `0`, `1`, `empty`.

I think it creates edge from `x to y` in "inner" `Map` (successors of `x`) so `y` is there but there is no `y` in "outer" `Map` (all vertices)

snowleopard: Ah, you are right! This illustrates how easy it is to make a mistake when you have invariants in your data structure. Whenever I'm working on `AdjacencyMap`s I have a feeling I'm walking a mine-field. Algebraic representation is so much safer :-)

snowleopard:
@Avasil Thanks for addressing my comments. Would you like me to merge, or do you prefer to do some benchmarking first?

Avasil:
@snowleopard Let's merge it, I'll try to do it anyway but I don't have a time right now. In case it turns out to be worse we can change it. :) Thanks for fast responses!

snowleopard:
Sounds good. Merged!

PR #136, opened 2018-10-28
Refactor non-empty graphs

snowleopard:
This PR implements the following:
* Switch from `NonEmptyGraph` to `NonEmpty.Graph`.
* Add `Algebra.Graph.AdjacencyMap.NonEmpty` as a preparation for #130.


snowleopard:
@nobrakal I'm not sure why, but I see a mysterious improvement of `hasEdge: 0.64 (good)`. Perhaps, this is related to unexpected improvements you were seeing for your PRs too?

Also, as you can see, I'm doing quite a bit refactoring to non-empty graphs here. I think it would be better to use simply `Graph` as the name for non-empty graphs. One can always use qualified imports and use `NonEmpty.Graph` to distinguish from graphs that can be empty (if need be).

I think this change breaks your performance regression suite for non-empty graphs. Sorry about that!

snowleopard:
OK, everything seems to be in order.

@nobrakal Merging with `hasEdge: 0.65 (good)`. 

P.S.: This is quite a big refactoring, so it will probably create a few conflicts in existing PRs (both yours and mine), but I prefer this to be merged first.

PR #137, opened 2018-11-02
Cabal flag for inspection testing

nobrakal:
Hi,

This is related to #129 

It adds a flag to disable test of rewrite rules, namely `inspection-testing`. Disabling it also remove the dependence on the `inspection-testing` package.

The flag is on by default.

### DIFF
--- a/algebraic-graphs.cabal
+++ b/algebraic-graphs.cabal
@@ -56,7 +56,7 @@ source-repository head
     type:     git
     location: https://github.com/snowleopard/alga.git
 
-Flag inspection-testing
+flag inspection-testing
     Description: Enable test of rewrite rules with inspection-testing
 
 library
--- a/algebraic-graphs.cabal
+++ b/algebraic-graphs.cabal
@@ -56,6 +56,9 @@ source-repository head
     type:     git
     location: https://github.com/snowleopard/alga.git
 
+Flag inspection-testing
+    Description: Enable test of rewrite rules with inspection-testing
+
 library
     hs-source-dirs:     src
     exposed-modules:    Algebra.Graph,
@@ -132,7 +135,7 @@ test-suite test-alga
                         Algebra.Graph.Test.Labelled.AdjacencyMap,
                         Algebra.Graph.Test.Relation,
                         Data.Graph.Test.Typed
-    if impl(ghc >= 8.0.2)
+    if (impl(ghc >= 8.0.2) && flag(inspection-testing))
         other-modules:  Algebra.Graph.Test.RewriteRules
     build-depends:      algebraic-graphs,
                         array        >= 0.4     && < 0.6,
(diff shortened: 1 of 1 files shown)


snowleopard:
Thanks! This looks good, but I wonder if the default should be different. Most users of the library are not going to need the test suite, so forcing them to depend on an unnecessary library seems wrong.

nobrakal:
From memory Cabal is intelligent enough to not require test dependencies when building only the library :) (at least Stack do it).

snowleopard:
@nobrakal Yes, you are right, however I think running `ghcid` with default parameters will still lead to the errors that I reported in #129. 

Could you try running `ghcid` with your patch? Ideally, it would be great if `ghcid` users did not have to pass anything special when working with Alga.

nobrakal:
Mhh, it seems strange to me that `ghcid` is loading the test suite by default !

If we switch off this flag by default, anyone wanting to run the test-suite will have to set it by hand, so it is a win for a loss.

One solution to solve the issue with `ghcid` can be to add a `.ghcid` file at the root of the alga repository, where we disable the flag.

snowleopard:
> Mhh, it seems strange to me that ghcid is loading the test suite by default !

I think that's the right thing to do: whenever you change the library API, you'll immediately need to address any following breakage in the testsuite.

> One solution to solve the issue with `ghcid` can be to add a `.ghcid` file at the root of the alga repository

This may indeed be the right solution, especially, if we somehow manage to convince `ghcid` to compile the file `RewriteRules.hs` with the necessary flags so that the inspection testing works in `ghcid` too. Do you think this is possible? Maybe we don't even need any `inspection-testing` flag then, only to add some settings to `.ghcid`. I imagine that with the help of `ghcid` it would be much more productive to tweak rewrite rules until the compilation succeeds (i.e. no need to run the whole testsuite manually).

nobrakal:
>  if we somehow manage to convince ghcid to compile the file RewriteRules.hs ...

Indeed it is possible,forcing `ghci` to optimize our code :) I added a `.ghcid` file that achieve this.

So it remove the need of the flag, I have reverted these changes.

Could you try and tell me if this ok for you ?

> it would be much more productive to tweak rewrite rules until the compilation succeeds (i.e. no need to run the whole testsuite manually).

You are right, using `ghcid` saves the time of typing `stack test` at each changes :)

snowleopard:
@nobrakal I tried it and it works, thanks!

Using `ghcid` with optimisation is quite a bit slower, of course, but I can live with it. Still, do you think it's possible to enable optimisation only selectively for some modules so that we can keep `ghcid` fast? I guess we need optimisation only in `Algebra.Graph` and `RewriteRules` modules.

nobrakal:
> Still, do you think it's possible to enable optimisation only selectively for some modules so that we can keep ghcid fast?

This seems complicated, as far as I know file-per-file optimization is available only via the `GHC_OPTIONS` pragma, and simply adding it to the concerned file does not solve the problem.

I don't really know how GHC handle semi-optimized packages...

Issue #140, opened 2018-11-14
Fix Show instances


snowleopard:
While working on #128, I realised that the `Show` instances we provide do not insert parentheses. For example, showing `vertex (vertex 1)` gives `vertex vertex 1` instead of `vertex (vertex 1)`. We need to fix this.

snowleopard:
Fixed by @pyrtsa in #142.

PR #142, opened 2018-11-19
Fix Show instances to insert parens (#140)

pyrtsa:
This should fix the starter bug #140.

Not quite all `show` code paths get tested with the updated unit tests; notably `Labelled.AdjacencyMap` remains with no tests.

### DIFF
--- a/test/Algebra/Graph/Test/Generic.hs
+++ b/test/Algebra/Graph/Test/Generic.hs
@@ -122,6 +122,22 @@ testShow (Testsuite prefix (%)) = do
     test "show (1 * 2 + 3) == \"overlay (vertex 3) (edge 1 2)\"" $
           show % (1 * 2 + 3) == "overlay (vertex 3) (edge 1 2)"
 
+    test "showsPrec 11 empty \"\" == \"empty\"" $
+          (showsPrec 11 % empty) "" == "empty"
+
+    test "showsPrec 11 1 \"\" == \"(vertex 1)\"" $
+          (showsPrec 11 % 1) "" == "(vertex 1)"
+
+    test "showsPrec 11 (1 * 2) \"\" == \"(edge 1 2)\"" $
+          (showsPrec 11 % (1 * 2)) "" == "(edge 1 2)"
+
+    test "showsPrec 11 (1 * 2 * 3) \"\" == \"(edges [(1,2),(1,3),(2,3)])\"" $
+          (showsPrec 11 % (1 * 2 * 3)) "" == "(edges [(1,2),(1,3),(2,3)])"
+
+    test "showsPrec 11 (1 * 2 + 3) \"\" == \"(overlay (vertex 3) (edge 1 2))\"" $
+          (showsPrec 11 % (1 * 2 + 3)) "" == "(overlay (vertex 3) (edge 1 2))"
+
+
 testOrd :: Testsuite -> IO ()
 testOrd (Testsuite prefix (%)) = do
     putStrLn $ "\n============ " ++ prefix ++ "Ord ============"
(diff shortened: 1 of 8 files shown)


pyrtsa:
Sorry, I don't quite follow what you're trying to say. My confusion is a mixture of the following:

1.  In the code snippet following the quote "[…], but with minuses:", you didn't actually sprinkle minuses anywhere!
2.  The alga `Num` instances in general have `negate = id`, so the intuitive way to produce a vertex with a negative number doesn't work as the following example illustrates, but one needs to use e.g. `fromInteger` explicitly:
    ```
    λ> show (-1 :: AdjacencyIntMap)
    "vertex 1"
    λ> show (fromInteger (-1) :: AdjacencyIntMap)
    "vertex (-1)"
    ```
    (The above snippet is from `stack repl` using the code from this PR, which already has the result parenthesised like you asked for, i.e. not `"vertex -1"`.)
3.  However, this `Show` instance is a bit of a lie because of the said `Num` instance — there's no way `vertex (-1)` would produce the same graph, instead, you'd get `vertex 1` basically! So should we instead inject `fromInteger` into the string form? (Yuck!)
4.  This final point is just about me failing to read English: did you mean the PR still needs work on the unit tests, or the implementation, or both?

Happy to help!

snowleopard:
@pyrtsa Sorry for not explaining my suggestion properly :)

Here are the unit tests I had in mind:
```haskell
show (vertex (-1)                            ) == "vertex (-1)"
show (vertex (-1) + vertex (-2)              ) == "vertices [-1,-2]"
show (vertex (-1) * vertex (-2)              ) == "edge (-1) (-2)"
show (vertex (-1) * vertex (-2) * vertex (-3)) == "edges [(-1,-2),(-1,-3),(-2,-3)]"
show (vertex (-1) * vertex (-2) + vertex (-3)) == "overlay (vertex (-3)) (edge (-1) (-2))"
```

I think your implementation should pass these tests, but let me know if I'm wrong.

Please add these tests to `test/Algebra/Graph/Test/Generic.hs`. You can keep your `showsPrec` tests (unless some of them become redundant).

You make a good point about `-1` being interpreted as `negate (vertex 1)` instead of the more desirable `vertex (-1)`. This is a common issue, which can be resolved using `NegativeLiterals` GHC extension: https://downloads.haskell.org/~ghc/master/users-guide/glasgow_exts.html#extension-NegativeLiterals. You can in principle enable it in `test/Algebra/Graph/Test/Generic.hs` if you like, but I have no strong opinion about this.

> This final point is just about me failing to read English: did you mean the PR still needs work on the unit tests, or the implementation, or both?

I think the implementation is fine. Let's just add the above unit tests, and I'm happy to merge. Thank you!

pyrtsa:
Ok, I think I got it now.

Also needed to change the order of values in the `vertices` and `edges` lists as negative values obviously sort the other way. And fixed the bug that the tests caught in the instance for `Relation`.

test/Algebra/Graph/Test/Generic.hs
```diff
@@ -1,4 +1,4 @@
-{-# LANGUAGE GADTs, RankNTypes, ViewPatterns #-}
+{-# LANGUAGE GADTs, NegativeLiterals, RankNTypes, ViewPatterns #-}
```
snowleopard: I don't think you are actually using the `NegativeLiterals` extension.

If you'd like to use it, you'll need to switch from `show % (vertex (-1))` to `show % (-1)`. 

But if you prefer to leave all tests as they are, then I guess we should drop `NegativeLiterals` from this list.

snowleopard:
@pyrtsa Looks good, thank you!

I'm ready to merge, but see a minor comment about unused `NegativeLiterals`.

test/Algebra/Graph/Test/Generic.hs
pyrtsa: Right! I used to make use of that extension at an earlier point but your suggested tests removed the need. Fixed now!

pyrtsa:
This should do it.

snowleopard:
@pyrtsa Merged! Thanks again for the PR; I hope you'll be back with more :-)

PR #144, opened 2018-11-24
Replace .ghcid with .ghci

snowleopard:
@nobrakal I propose to change our `ghcid` setup: instead of using the `.ghcid` file, we'll use `.ghci` instead. This seems to be faster in the sense that we don't need to compile with optimisation.

As a drawback, I think the rewrite rules are no longer tested automatically.

### DIFF
--- a/.ghci
+++ b/.ghci
@@ -2,5 +2,6 @@
 
 :set -isrc
 :set -itest
+:set prompt "> "
 
-:load test/Main.hs
+:load Main Algebra.Graph.Labelled.Example.Automaton Algebra.Graph.Labelled.Example.Network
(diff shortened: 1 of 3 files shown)


snowleopard:
There is one problem I'm not sure how to fix. If I run `stack ghci --test` I get the following error:

```
Ok, 47 modules loaded.
Loaded GHCi configuration from D:\code\alga\.ghci

<no location info>: error:
    module `main:Main' is defined in multiple files: D:\code\alga\test\Main.hs
                                                     D:\code\alga\test\Main.hs
Failed, 47 modules loaded.
```

snowleopard:
I've created an issue for the above failure: #145.

PR #147, opened 2018-11-27
Minor revision

snowleopard:
@nobrakal Note that I have added an expected failure test to `RewriteRules`:

```haskell
-- TODO: Why can't we reach the desired target expression?
inspect $ 'vertices2 =/= 'verticesR
```

I'm following the following naming convention:
```haskell
-- Naming convention: we use the suffix "R" to indicate the desired outcome of
-- rewrite rules, and suffices "1", "2", etc. to indicate initial expressions.
```

### DIFF
--- a/test/Algebra/Graph/Test/RewriteRules.hs
+++ b/test/Algebra/Graph/Test/RewriteRules.hs
@@ -76,7 +76,12 @@ vertices2 = transpose . overlays . map vertex
 
 inspect $ 'vertices2 === 'vertices1
 
--- TODO: Why can't we reach the desired target expression?
+-- Note that we currently have these three tests:
+-- * vertices2 === vertices1
+-- * vertices1 === verticesR
+-- * vertices2 =/= verticesR
+-- This non-transitivity is awkward, and feels like a bug in the inspection
+-- testing library. See https://github.com/nomeata/inspection-testing/issues/23.
 inspect $ 'vertices2 =/= 'verticesR
 
 cliqueT1, cliqueTR :: [a] -> Graph a


test/Algebra/Graph/Test/RewriteRules.hs
```diff
@@ -2,12 +2,12 @@
 -----------------------------------------------------------------------------
 -- |
 -- Module     : Algebra.Graph.Test.RewriteRules
--- Copyright  : (c) Andrey Mokhov 2016-2018
+-- Copyright  : (c) Andrey Mokhov R016-R018
```
nobrakal: @snowleopard  Is this normal ?

snowleopard: Ouch :) Thanks for spotting -- that's not normal at all, but is an innacurate Find & Replace in action!

test/Algebra/Graph/Test/RewriteRules.hs
```diff
@@ -18,81 +18,76 @@ import Algebra.Graph.Internal
...
+vertices2 = transpose . overlays . map vertex
 
---- transpose . clique
-cliqueTransposed, transposeDotClique :: [a] -> Graph a
-cliqueTransposed   = fromMaybe Empty . foldr (maybeF (flip Connect) . Vertex) Nothing
-transposeDotClique = transpose . connects . map vertex
+inspect $ 'vertices2 === 'vertices1
 
-inspect $ 'cliqueTransposed === 'transposeDotClique
+-- TODO: Why can't we reach the desired target expression?
```
nobrakal: About your todo,

The CSE is replacing the body of `verticesR` with only `vertices1`:
```
        verticesR :: forall a. [a] -> Graph a
        [LclIdX,                   
         Arity=1,                  
         Str=<S,1*U>,              
         Unf=Unf{Src=<vanilla>, TopLvl=True, Value=True, ConLike=True,
                 WorkFree=True, Expandable=True,
                 Guidance=ALWAYS_IF(arity=0,unsat_ok=True,boring_ok=True)}]
        verticesR = vertices1 
```

and GHC will never inline this definition so they are declared as different (I have reported this "bug" for inspection-testing).

test/Algebra/Graph/Test/RewriteRules.hs
snowleopard: Fixed!

test/Algebra/Graph/Test/RewriteRules.hs
snowleopard: Ah, yes, thanks for clarifying. I thought that was related to the CSE issue, but couldn't quite connect all dots.

snowleopard: So, we have a strange non-transitive situation:

```haskell
inspect $ 'vertices1 === 'verticesR
inspect $ 'vertices2 === 'vertices1
inspect $ 'vertices2 =/= 'verticesR
```
I guess I'll add this non-transitivity example to your `inspection-testing` issue.

snowleopard: I've updated the comment to:
```
-- Note that we currently have these three tests:
-- * vertices2 === vertices1
-- * vertices1 === verticesR
-- * vertices2 =/= verticesR
-- This non-transitivity is awkward, and feels like a bug in the inspection
-- testing library. See https://github.com/nomeata/inspection-testing/issues/23.
```

PR #157, opened 2018-12-30
Rewrite Rules for bind

nobrakal:
Hi there,

I have:
* implemented the deforestation rules for `bind`
* rewritten `fmap` in term of `bind`
* added a test to verify that `fmap f . fmap g` was rewritten in `fmap (f.g)`

Happy Holidays :)

### DIFF
--- a/src/Algebra/Graph.hs
+++ b/src/Algebra/Graph.hs
@@ -212,10 +212,9 @@ creating our own intermediate functions for guiding rewrite rules when needed.
 instance Functor Graph where
     fmap = fmapR
 
--- This is a usual implementation of 'fmap', but with custom rewrite rules.
 fmapR :: (a -> b) -> Graph a -> Graph b
-fmapR f = foldg empty (vertex . f) overlay connect
-{-# INLINE [0] fmapR #-}
+fmapR f g = g >>= (vertex . f)
+{-# INLINE fmapR #-}
 
 instance NFData a => NFData (Graph a) where
     rnf Empty         = ()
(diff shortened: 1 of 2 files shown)


src/Algebra/Graph.hs
```diff
@@ -212,10 +212,9 @@ creating our own intermediate functions for guiding rewrite rules when needed.
 instance Functor Graph where
     fmap = fmapR
 
--- This is a usual implementation of 'fmap', but with custom rewrite rules.
 fmapR :: (a -> b) -> Graph a -> Graph b
-fmapR f = foldg empty (vertex . f) overlay connect
-{-# INLINE [0] fmapR #-}
+fmapR f g = g >>= (vertex . f)
```
snowleopard: Perhaps, use `bindR` directly to avoid the (mental) indirection?

nobrakal:
> Could you describe the motivation for this change, very briefly?

Oh, yes, I must admit it can seems a bit strange ^^. 
Very briefly: this change reflect that `fmap` can be expressed with `bind` and `bind` can be fused with `foldg`. Moreover, any function preserving the structure of the graph (like `fmap`) can be expressed with `bind`, so this is an "optimal" change, since we do not have to add any other rule for structure preserving-functions.

>  Presumably, this is a part of a grander plan which you highlighted in your blog post

I have not yet explored other fields (I did not forget about hylomorphisms and other recursion operators), so I hope, but I don't see any other direct improvement ^^. (If anyone is reading this, the concerned blog post is [here](https://blog.nyarlathotep.one/2018/12/algebraic-graphs-homomorphisms/)).

snowleopard:
Thanks @nobrakal! I've added your comment to the commit message.

PR #160, opened 2019-01-12
Make (<*>) subject to fusion

nobrakal:
Hi there,

This only change the definition of `(<*>)` in `Algebra.Graph` to an equivalent made with `bindR` (suggested by @snowleopard  ), which allow fusion with compatible functions.

### DIFF
--- a/src/Algebra/Graph.hs
+++ b/src/Algebra/Graph.hs
@@ -1165,14 +1165,10 @@ matchR :: b -> (a -> b) -> (a -> Bool) -> a -> b
 matchR e v p = \x -> if p x then v x else e
 {-# INLINE [0] matchR #-}
 
-toR :: b -> (a -> b) -> (b -> b -> b) -> (b -> b -> b) -> Graph a -> b
-toR e v o c g = foldg e v o c g
-{-# INLINE [0] toR #-}
-
 -- These rules transform functions into their buildR equivalents.
 {-# RULES
 "buildR/bindR" forall f g.
-    bindR g f = buildR (\e v o c -> foldg e (composeR (toR e v o c) f) o c g)
+    bindR g f = buildR (\e v o c -> foldg e (composeR (foldg e v o c) f) o c g)
 
 "buildR/induce" [~1] forall p g.
     induce p g = buildR (\e v o c -> foldg e (matchR e v p) o c g)
@@ -1198,8 +1194,9 @@ toR e v o c g = foldg e v o c g
 "bindR/bindR" forall c f g.
     composeR (composeR c f) g = composeR c (f.g)
 
-"toR/final" forall f.
-    composeR (toR Empty Vertex Overlay Connect) f = f
+-- Remove identity (which can appear in the rewriting of bindR)
+"foldg/id"
+    foldg Empty Vertex Overlay Connect = id
  #-}
 
 -- Eliminate remaining rewrite-only functions.
--- a/src/Algebra/Graph.hs
+++ b/src/Algebra/Graph.hs
@@ -260,8 +260,12 @@ ordIntR :: Graph Int -> Graph Int -> Ordering
 ordIntR x y = compare (toAdjacencyIntMap x) (toAdjacencyIntMap y)
 
 instance Applicative Graph where
-    pure    = Vertex
-    f <*> x = bindR f (<$> x)
+    pure  = Vertex
+    (<*>) = seqApR
+
+seqApR :: Graph (a -> b) -> Graph a -> Graph b
+seqApR f x = bindR f (<$> x)
+{-# INLINE seqApR #-}
 
 instance Monad Graph where
     return = pure
@@ -1161,10 +1165,14 @@ matchR :: b -> (a -> b) -> (a -> Bool) -> a -> b
 matchR e v p = \x -> if p x then v x else e
 {-# INLINE [0] matchR #-}
 
+toR :: b -> (a -> b) -> (b -> b -> b) -> (b -> b -> b) -> Graph a -> b
+toR e v o c g = foldg e v o c g
+{-# INLINE [0] toR #-}
+
 -- These rules transform functions into their buildR equivalents.
 {-# RULES
 "buildR/bindR" forall f g.
-    bindR g f = buildR (\e v o c -> foldg e (composeR (foldg e v o c) f) o c g)
+    bindR g f = buildR (\e v o c -> foldg e (composeR (toR e v o c) f) o c g)
 
 "buildR/induce" [~1] forall p g.
     induce p g = buildR (\e v o c -> foldg e (matchR e v p) o c g)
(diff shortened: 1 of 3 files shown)


nobrakal:
Hi,

> Please do the same change in Algebra.Graph.NonEmpty too.

Done

> If yes, perhaps, we should also add a new rewrite test?

Well your question lead me to a bug in my implementation of bind rewriting (in fact, the use of any bind included a destruction-reconstruction of the produced graph, which is not (at all) cleaver).

So I corrected it and added a test to verify that binds were well merged (directly in this PR since the changes here are pretty small).

I am working to make an other test pass for sequential application, but I did not come yet to something working...

snowleopard:
@nobrakal Thanks! OK, I'll wait until you manage to fix the second test :)

test/Algebra/Graph/Test/RewriteRules.hs
```diff
@@ -102,3 +102,15 @@ fmapFmap1 g f h = fmap h (fmap f g)
...
+
+bind2, bind2R :: (a -> Graph b) -> (b -> Graph c) -> Graph a -> Graph c
+bind2 f g x = x >>= f >>= g
+bind2R f g x = x >>= (\x -> f x >>= g)
+
+inspect $ 'bind2 === 'bind2R
+
+ovSeqAp, seqApOv :: Graph (a -> b) -> Graph (a -> b) -> Graph a -> Graph b
+ovSeqAp x y z = (overlay x y) <*> z
+seqApOv x y z = overlay (x >>= (<$> z)) (y >>= (<$> z))
```
snowleopard: Why can't we make proper left distributivity work `overlay x y <*> z == overlay (x <*> z) (y <*> z)`? 

What about the right distributivity `x <*> overlay y z == overlay (x <*> y) (x <*> z)`?

nobrakal: Oops, sorry for the lack of explanation, I forget to write them !

So I am hitting a little problem:  if I use directly
```Haskell
seqApOv x y z = overlay (x <*> z) (y <*> z)
```
`<*>` is only replaced by `seqApR` and `seqApR` is then not inlined, and I don't know why...
The only way I found to force its inlining is adding the corresponding rewrite rule, but this is not very satisfying.
Note that `<*>` is correctly inlined in `ovSeqAp`.

So I just inlined it "by hand" but this is not very beautiful ^^.

I think the problem comes from the fact that `<*>` is a class function, but I don't really know how to solve this...

snowleopard: In this case, it may be better to actually document the problem, and create an inspection test for the inequality (e.g. `=/=`). At some point this will hopefully be resolved and we'll swap to `===`.

nobrakal: This is corrected, I added both tests with an explanation.

Sadly, I am not able to make right-distributivity works right now, and in fact I don't know if it is true (from a structural point of view)

src/Algebra/Graph.hs
```diff
@@ -261,7 +261,11 @@ ordIntR x y = compare (toAdjacencyIntMap x) (toAdjacencyIntMap y)
 
 instance Applicative Graph where
     pure  = Vertex
-    (<*>) = ap
+    (<*>) = seqApR
+
+seqApR :: Graph (a -> b) -> Graph a -> Graph b
```
snowleopard: What does the `seq` prefix mean? I think `apR` would be clearer and more consistent with other such functions.

src/Algebra/Graph.hs
```diff
@@ -1189,6 +1193,10 @@ matchR e v p = \x -> if p x then v x else e
 -- their buildR form.
 "bindR/bindR" forall c f g.
     composeR (composeR c f) g = composeR c (f.g)
+
+-- Remove identity (which can appear in the rewriting of bindR)
```
snowleopard: This rule doesn't actually "remove" identity -- it rewrites it into a simpler/more efficient form.

test/Algebra/Graph/Test/RewriteRules.hs
```diff
@@ -102,3 +102,25 @@ fmapFmap1 g f h = fmap h (fmap f g)
...
+bind2R f g x = x >>= (\x -> f x >>= g)
+
+inspect $ 'bind2 === 'bind2R
+
+-- Ideally, we want this test to pass.
+-- Strangely, '<*>' in 'ovSeqApR' does not inline and makes the test fail.
+--
+-- This is corrected below, where '<*>' was inlined "by hand"
+ovSeqAp, ovSeqApR :: Graph (a -> b) -> Graph (a -> b) -> Graph a -> Graph b
+ovSeqAp  x y z = (overlay x y) <*> z
```
snowleopard: Again, I think it may be best to drop the `seq` part, as it's unclear what it stands for.

Also: the parentheses around `overlay x y` are redundant.

src/Algebra/Graph.hs
nobrakal: `seqAp` for "sequential application", but you are right, it can be confused with `seq` :p

test/Algebra/Graph/Test/RewriteRules.hs
snowleopard: Thanks! Let me think about right distributivity for a day or two... I'm curious whether it actually works or not :)

src/Algebra/Graph.hs
nobrakal: Fixed ^^

test/Algebra/Graph/Test/RewriteRules.hs
nobrakal: Ah indeed, you are right :)

test/Algebra/Graph/Test/RewriteRules.hs
```diff
@@ -102,3 +102,25 @@ fmapFmap1 g f h = fmap h (fmap f g)
...
 inspect $ 'fmapFmap1 === 'fmapFmapR
+
+bind2, bind2R :: (a -> Graph b) -> (b -> Graph c) -> Graph a -> Graph c
+bind2 f g x = x >>= f >>= g
+bind2R f g x = x >>= (\x -> f x >>= g)
+
+inspect $ 'bind2 === 'bind2R
+
+-- Ideally, we want this test to pass.
+-- Strangely, '<*>' in 'ovSeqApR' does not inline and makes the test fail.
```
snowleopard: Leftover `seq`.

src/Algebra/Graph/NonEmpty.hs
```diff
@@ -212,8 +212,8 @@ ordInt :: Graph Int -> Graph Int -> Ordering
 ordInt x y = compare (T.toAdjacencyIntMap x) (T.toAdjacencyIntMap y)
 
 instance Applicative Graph where
-    pure  = Vertex
-    (<*>) = ap
+    pure    = Vertex
+    f <*> x = f >>= (<$> x)
```
snowleopard:
```suggestion
    f <*> x = apR f x
```

Why not this? Seems simpler and less code duplication. But perhaps this makes rewriting more difficult?

nobrakal: I just not implemented all the rewrite rules system for `Graph.NonEmpty`, so `<*>` will anyway not be rewritten :p

snowleopard: Oh, sorry, my bad :)

snowleopard:
Thank you, merged!

PR #161, opened 2019-01-12
Speedup for isSubgraphOf

nobrakal:
Hi,

I just saw that in the actual `isSubGraphOf`, the second argument was converted two times in the same adjacency map in `Algebra.Graph{,.NonEmpty}`.
This PR solves this by introducing a variable. Note that the code is a bit less clear and I had to add specialized functions by hand (since I use directly `overlay` from `AdjacencyMap` or `AdjacencyIntMap` ).

### DIFF
--- a/src/Algebra/Graph.hs
+++ b/src/Algebra/Graph.hs
@@ -473,17 +473,13 @@ foldg e v o c = go
 -- isSubgraphOf x y                         ==> x <= y
 -- @
 isSubgraphOf :: Ord a => Graph a -> Graph a -> Bool
-isSubgraphOf x y = AM.overlay (toAdjacencyMap x) yAM == yAM
-  where
-    yAM = toAdjacencyMap y
+isSubgraphOf x y = AM.isSubgraphOf (toAdjacencyMap x) (toAdjacencyMap y)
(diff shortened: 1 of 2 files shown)


src/Algebra/Graph.hs
```diff
@@ -473,8 +473,17 @@ foldg e v o c = go
 -- isSubgraphOf x y                         ==> x <= y
 -- @
 isSubgraphOf :: Ord a => Graph a -> Graph a -> Bool
-isSubgraphOf x y = overlay x y == y
-{-# SPECIALISE isSubgraphOf :: Graph Int -> Graph Int -> Bool #-}
+isSubgraphOf x y = AM.overlay (toAdjacencyMap x) yAM == yAM
```
snowleopard: @nobrakal Well spotted!

Actually, we shouldn't even be using the `overlay`-based definition. The most efficient implementation is to simply call the `AM.isSubgraphOf` function after applying `toAdjacencyMap` to the arguments.

nobrakal: Haha, you are right ! I just changed it !

snowleopard:
@nobrakal Thank you, merged!

Issue #156, opened 2018-12-19
Add Bifunctor instance for labelled graphs


boggle:
Labelled graphs are parameterized in both their node and edge type.

base has Bifunctor since 4.8
(http://hackage.haskell.org/package/base-4.12.0.0/docs/Data-Bifunctor.html)

Adding a Bifunctor instance would allow implementing `emap = Bifunctor.second`

snowleopard:
@boggle Sounds good!

I'll have some time to implement this in January, but feel free to send a PR earlier.

snowleopard:
Fixed in #162.

Issue #150, opened 2018-11-28
Implement custom `shrink` for non-algebraic graph data types


snowleopard:
Right now we do not provide a custom implementation of the method `shrink` in `Arbitrary` instances of non-algebraic data types, such as `AdjacencyMap`, which makes it difficult to understand test failures, e.g. see https://travis-ci.org/snowleopard/alga/jobs/460799751:

```
============ Labelled.AdjacencyMap.replaceEdge ============
Test failure:
    replaceEdge e x y m                 == overlay (removeEdge x y m) (edge e x y)
*** Failed! Falsifiable (after 11 tests and 3 shrinks):
Sum {getSum = 0}
1
0
[...]

Issue #120, opened 2018-09-20
Improve test coverage


snowleopard:
Despite having 640+ QuickCheck properties, test coverage of some modules is not great:

| module                                 | Top Level Definitions | Alternatives | Expressions |
|----------------------------------------|-----------------------|--------------|-------------|
|                                        | %                     | %            | %           |
| Algebra.Graph                          | 70%                   | 93%          | 91%         |
| Algebra.Graph.AdjacencyIntMap          | 100%                  | 100%         | 100%        |
| Algebra.Graph.AdjacencyIntMap.Internal | 73%                   | 100%         | 95%         |
| Algebra.Graph.AdjacencyMap             | 100%                  | 100%         | 99%         |
| Algebra.Graph.AdjacencyMap.Internal    | 73%                   | 100%         | 95%         |
| Algebra.Graph.Class                    | 50%                   | 40%          | 28%         |
| Algebra.Graph.Export                   | 72%                   | 100%         | 89%         |
| Algebra.Graph.Export.Dot               | 42%                   | 100%         | 92%         |
| Algebra.Graph.Fold                     | 63%                   | 100%         | 86%         |
| Algebra.Graph.HigherKinded.Class       | -                     | -            | -           |
| Algebra.Graph.Internal                 | -                     | -            | -           |
| Algebra.Graph.NonEmpty                 | -                     | -            | -           |
| Algebra.Graph.Relation                 | 100%                  | 100%         | 100%        |
| Algebra.Graph.Relation.Internal        | 77%                   | 100%         | 91%         |
| Algebra.Graph.Relation.InternalDerived | 40%                   | -            | 84%         |
| Algebra.Graph.Relation.Symmetric       | 66%                   | -            | 87%         |
| Algebra.Graph.ToGraph                  | 100%                  | -            | 100%        |
| Data.Graph.Typed                       | 88%                   | -            | 94%         |
|                                        |                       |              |             |
| Program Coverage Total                 | 76%                   | 92%          | 90%         |

Let's improve this.

PR #162, opened 2019-01-22
Add bifunctor instance for labelled graphs

andreabedini:
I was looking for simple contributions to give to an Haskell project and I found #156, so I have given it a go. This is my first [multi character](https://github.com/supki/ldap-client/commit/824b44ac22eb68d734500e5378e2be50f315fefb) contribution in Haskell so any kind of comments or suggestions are welcome  :-)

Also let me know if this corresponds to what you had in mind.

### DIFF
--- a/src/Algebra/Graph/Labelled.hs
+++ b/src/Algebra/Graph/Labelled.hs
@@ -87,7 +87,7 @@ instance (Ord a, Num a, Dioid e) => Num (Graph e a) where
 
 #if MIN_VERSION_base(4,8,0)
 instance Bifunctor Graph where
-  bimap f g = emap f . fmap g
+  bimap f g = foldg Empty (Vertex . g) (Connect . f)
 #endif
 
 -- TODO: This is a very inefficient implementation. Find a way to construct an
--- a/src/Algebra/Graph/Labelled.hs
+++ b/src/Algebra/Graph/Labelled.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE DeriveFunctor, FlexibleInstances #-}
+{-# LANGUAGE CPP, DeriveFunctor, FlexibleInstances #-}
 -----------------------------------------------------------------------------
 -- |
 -- Module     : Algebra.Graph.Labelled
@@ -46,6 +46,9 @@ module Algebra.Graph.Labelled (
 import Prelude ()
 import Prelude.Compat
 
+#if MIN_VERSION_base(4,8,0)
+import Data.Bifunctor
+#endif
 import Data.Monoid (Any (..))
 import Data.Semigroup ((<>))
 
@@ -81,6 +84,11 @@ instance (Ord a, Num a, Dioid e) => Num (Graph e a) where
     signum      = const empty
     abs         = id
     negate      = id
+
+#if MIN_VERSION_base(4,8,0)
+instance Bifunctor Graph where
+  bimap f g = emap f . fmap g
+#endif
 
 -- TODO: This is a very inefficient implementation. Find a way to construct an
 -- adjacency map directly, without building intermediate representations for all


nobrakal:
Hi there,

Happy to see you here :D 

Your implementation seems totally correct, but leads to an unneeded destruction/reconstruction of the graph (fmap will first deconstruct the graph, change the vertices, then reconstruct it, while emap will do the same thing changing the label of the edges).

I am pretty sure this can be avoided using directly `foldg` to go through the graph structure only once.

Otherwise, I don't see any problem :)

snowleopard:
Hi @andreabedini! Thank you for the PR!

I agree with @nobrakal, your implementation can be made more efficient via `foldg`: 

```haskell
    bimap f g = foldg Empty (Vertex . g) (Connect . f)
```

I hope it compiles and passes tests, I haven't actually tried it :)

snowleopard:
Actually, speaking of tests: could you add a test that actually uses your implementation, i.e. tests that for any functions `f` and `g`, and any labelled graph `x`, we have `bimap f g x == (emap f . fmap g) x`? 

Have a look how we test `emap`: 

https://github.com/snowleopard/alga/blob/master/test/Algebra/Graph/Test/Labelled/Graph.hs#L346-L377

Our current tests are not ideal, as you can see in the `TODO`.

andreabedini:
@nobrakal @snowleopard thanks for the feedback (and for the warm welcome!). I have reimplemented bimap as @snowleopard suggested, also defined emap in terms of bimap where we have it. Tests seems to pass.

@snowleopard I'll have a look at how to improve the tests a bit later.

By the way, this got me thinking. Is there an expectation for Functor and Bifunctor instances to be "compatible"? I.e. for a type `F a b` where `F` is a `Bifunctor` and `F a` is a `Functor`. Should one expect that `fmap f . bimap g h == bimap g (f . h)`? said differently `fmap == second`?

src/Algebra/Graph/Labelled.hs
```diff
@@ -482,7 +490,12 @@ transpose = foldg empty vertex (fmap flip connect)
 -- emap g . emap h        == emap (g . h)
 -- @
 emap :: (e -> f) -> Graph e a -> Graph f a
-emap f = foldg Empty Vertex (Connect . f)
+emap f =
```
snowleopard: I find this a bit awkward. Let's keep `emap f = foldg Empty Vertex (Connect . f)` for now (at least until we need to support pre-Bifunctor `base`).

snowleopard:
> I have reimplemented bimap as @snowleopard suggested, also defined emap in terms of bimap where we have it. Tests seems to pass.

@andreabedini Nice! Well, it was expected, but it's good to have it tested :) I would however prefer to revert to the previous definition of `emap` since it's less noisy.

> Should one expect that `fmap f . bimap g h == bimap g (f . h)`?

Actually, I think this is a free theorem, i.e. it's impossible to violate this given that `bimap id id == id`. Have a look here: https://github.com/quchen/articles/blob/master/second_functor_law.md

andreabedini:
> @andreabedini Nice! Well, it was expected, but it's good to have it tested :) I would however prefer to revert to the previous definition of `emap` since it's less noisy.

Done!

> Actually, I think this is a free theorem, i.e. it's impossible to violate this law given that `bimap id id == id`. You can find some intuition here: https://github.com/quchen/articles/blob/master/second_functor_law.md

Right, I was thinking only in terms of laws but seems to be a consequence of parametricity. I appreciate the link, thanks.

snowleopard:
@andreabedini Thank you!

I think we can merge this as is, without `Bifunctor` tests, because at the moment we have generally poor test coverage for many instances, which should be addressed in #120.

Let me know what you prefer: merge as is, or wait until you implement some tests. I'm happy either way!

andreabedini:
> Let me know what you prefer: merge as is, or wait until you implement some tests. I'm happy either way!

I'd say given we have a ticket already opened for better test coverage and you are ok with this, we can merge this first (so it doesn't get hung up on me writing more tests). I have started on #150 as well and I might finish that before looking at the tests.

snowleopard:
@andreabedini Sounds great, merged! Thank you for this and future PRs :-)

PR #163, opened 2019-01-28
Implement custom shrinking for AdjacencyMap and AdjancencyIntMap

andreabedini:
This addresses issue #150.

Nothing very smart here, I consider graphs with one less vertex and one less edge. Tests pass but I haven't managed to reproduce an existing bug to see if counterexamples would get smaller.

### DIFF
--- a/test/Algebra/Graph/Test/Arbitrary.hs
+++ b/test/Algebra/Graph/Test/Arbitrary.hs
@@ -36,6 +36,7 @@ import qualified Algebra.Graph.AdjacencyIntMap       as AdjacencyIntMap
 import qualified Algebra.Graph.AdjacencyMap          as AdjacencyMap
 import qualified Algebra.Graph.NonEmpty.AdjacencyMap as NAM
 import qualified Algebra.Graph.Class                 as C
+import qualified Algebra.Graph.Fold                  as Fold
 import qualified Algebra.Graph.Labelled              as LG
 import qualified Algebra.Graph.Labelled.AdjacencyMap as LAM
 import qualified Algebra.Graph.NonEmpty              as NonEmpty
@@ -62,9 +63,19 @@ instance Arbitrary a => Arbitrary (Graph a) where
     shrink (Connect x y) = [Empty, x, y, Overlay x y]
                         ++ [Connect x' y' | (x', y') <- shrink (x, y) ]
 
--- TODO: Implement a custom shrink method.
-instance Arbitrary a => Arbitrary (Fold a) where
+instance (Eq a, Ord a, Arbitrary a) => Arbitrary (Fold a) where
     arbitrary = arbitraryGraph
+
+    shrink g = oneLessVertex ++ oneLessEdge
+      where
+         oneLessVertex =
+           let vertices = Fold.vertexList g
+           in  [ Fold.removeVertex v g | v <- vertices ]
+
+         oneLessEdge =
+           let edges = Fold.edgeList g
+           in  [ Fold.removeEdge v w g | (v, w) <- edges ]
+
 
 -- | Generate an arbitrary 'NonEmpty.Graph' value of a specified size.
 arbitraryNonEmptyGraph :: Arbitrary a => Gen (NonEmpty.Graph a)
@@ -90,9 +101,19 @@ instance Arbitrary a => Arbitrary (NonEmpty.Graph a) where
 arbitraryRelation :: (Arbitrary a, Ord a) => Gen (Relation a)
 arbitraryRelation = Relation.stars <$> arbitrary
 
--- TODO: Implement a custom shrink method.
 instance (Arbitrary a, Ord a) => Arbitrary (Relation a) where
     arbitrary = arbitraryRelation
+
+    shrink g = oneLessVertex ++ oneLessEdge
+      where
+         oneLessVertex =
+           let vertices = Relation.vertexList g
+           in  [ Relation.removeVertex v g | v <- vertices ]
+
+         oneLessEdge =
+           let edges = Relation.edgeList g
+           in  [ Relation.removeEdge v w g | (v, w) <- edges ]
+
 
 instance (Arbitrary a, Ord a) => Arbitrary (ReflexiveRelation a) where
     arbitrary = ReflexiveRelation <$> arbitraryRelation
--- a/test/Algebra/Graph/Test/Arbitrary.hs
+++ b/test/Algebra/Graph/Test/Arbitrary.hs
@@ -18,7 +18,8 @@ import Prelude ()
 import Prelude.Compat
 
 import Control.Monad
-import Data.List.NonEmpty (NonEmpty (..))
+import Data.List.NonEmpty (NonEmpty (..), toList)
+import Data.Maybe (catMaybes)
 import Data.Tree
 import Test.QuickCheck
 
@@ -110,9 +111,18 @@ instance (Arbitrary a, Ord a) => Arbitrary (PreorderRelation a) where
 arbitraryAdjacencyMap :: (Arbitrary a, Ord a) => Gen (AdjacencyMap a)
 arbitraryAdjacencyMap = AdjacencyMap.stars <$> arbitrary
 
--- TODO: Implement a custom shrink method.
 instance (Arbitrary a, Ord a) => Arbitrary (AdjacencyMap a) where
     arbitrary = arbitraryAdjacencyMap
+
+    shrink g = oneLessVertex ++ oneLessEdge
+      where
+         oneLessVertex =
+           let vertices = AdjacencyMap.vertexList g
+           in  [ AdjacencyMap.removeVertex v g | v <- vertices ]
+
+         oneLessEdge =
+           let edges = AdjacencyMap.edgeList g
+           in  [ AdjacencyMap.removeEdge v w g | (v, w) <- edges ]
 
 -- | Generate an arbitrary non-empty 'NAM.AdjacencyMap'. It is guaranteed that
 -- the resulting adjacency map is 'consistent'.
@@ -127,27 +137,54 @@ arbitraryNonEmptyAdjacencyMap = NAM.stars1 <$> nonEmpty
                 return ((x, []) :| []) -- There must be at least one vertex
             (x:xs) -> return (x :| xs)
 
--- TODO: Implement a custom shrink method.
 instance (Arbitrary a, Ord a) => Arbitrary (NAM.AdjacencyMap a) where
     arbitrary = arbitraryNonEmptyAdjacencyMap
+
+    shrink g = oneLessVertex ++ oneLessEdge
+      where
+         oneLessVertex =
+           let vertices = toList $ NAM.vertexList1 g
+           in catMaybes [ NAM.removeVertex1 v g | v <- vertices ]
+
+         oneLessEdge =
+           let edges = NAM.edgeList g
+           in  [ NAM.removeEdge v w g | (v, w) <- edges ]
 
 -- | Generate an arbitrary 'AdjacencyIntMap'. It is guaranteed that the
 -- resulting adjacency map is 'consistent'.
 arbitraryAdjacencyIntMap :: Gen AdjacencyIntMap
 arbitraryAdjacencyIntMap = AdjacencyIntMap.stars <$> arbitrary
 
--- TODO: Implement a custom shrink method.
 instance Arbitrary AdjacencyIntMap where
     arbitrary = arbitraryAdjacencyIntMap
+
+    shrink g = oneLessVertex ++ oneLessEdge
+      where
+         oneLessVertex =
+           let vertices = AdjacencyIntMap.vertexList g
+           in  [ AdjacencyIntMap.removeVertex v g | v <- vertices ]
+
+         oneLessEdge =
+           let edges = AdjacencyIntMap.edgeList g
+           in  [ AdjacencyIntMap.removeEdge v w g | (v, w) <- edges ]
 
 -- | Generate an arbitrary labelled 'LAM.AdjacencyMap'. It is guaranteed
 -- that the resulting adjacency map is 'consistent'.
 arbitraryLabelledAdjacencyMap :: (Arbitrary a, Ord a, Eq e, Arbitrary e, Monoid e) => Gen (LAM.AdjacencyMap e a)
 arbitraryLabelledAdjacencyMap = LAM.fromAdjacencyMaps <$> arbitrary
 
--- TODO: Implement a custom shrink method.
 instance (Arbitrary a, Ord a, Eq e, Arbitrary e, Monoid e) => Arbitrary (LAM.AdjacencyMap e a) where
     arbitrary = arbitraryLabelledAdjacencyMap
+
+    shrink g = oneLessVertex ++ oneLessEdge
+      where
+         oneLessVertex =
+           let vertices = LAM.vertexList g
+           in  [ LAM.removeVertex v g | v <- vertices ]
+
+         oneLessEdge =
+           let edges = LAM.edgeList g
+           in  [ LAM.removeEdge v w g | (_, v, w) <- edges ]
 
 -- | Generate an arbitrary labelled 'LAM.Graph' value of a specified size.
 arbitraryLabelledGraph :: (Arbitrary a, Arbitrary e) => Gen (LG.Graph e a)
--- a/test/Algebra/Graph/Test/Arbitrary.hs
+++ b/test/Algebra/Graph/Test/Arbitrary.hs
@@ -205,7 +205,6 @@ instance (Arbitrary a, Arbitrary e, Monoid e) => Arbitrary (LG.Graph e a) where
     shrink (LG.Connect e x y) = [LG.Empty, x, y, LG.Connect mempty x y]
                              ++ [LG.Connect e x' y' | (x', y') <- shrink (x, y) ]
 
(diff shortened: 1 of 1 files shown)


snowleopard:
@andreabedini Many thanks! This looks good to me, but there seem to be one more TODO:

```haskell
-- TODO: Implement a custom shrink method.
instance (Arbitrary a, Ord a, Eq e, Arbitrary e, Monoid e) => Arbitrary (LAM.AdjacencyMap e a) where
    arbitrary = arbitraryLabelledAdjacencyMap
```

Could you fix it too?

snowleopard:
@andreabedini Just realised that there are actually several more TODOs about the `shrink` method. Do you think you could tackle all of them in one go?

andreabedini:
> @andreabedini Just realised that there are actually several more TODOs about the `shrink` method. Do you think you could tackle all of them in one go?

I'll have a look at the other ones too 👍

snowleopard:
Awesome, thank you!

andreabedini:
Actually, I am realising I have been writing silly instances of shrink! I was under the impression that shrink was going to be applied recursively but not! therefore my implementation for `AdjacencyMap` only removes one vertex and then only removes one edge ... and it stops there.

I have cooked this up

```
betterAdjacencyMapShrink :: (Arbitrary a, Ord a) => AdjacencyMap a -> [AdjacencyMap a]                        
betterAdjacencyMapShrink =
    map (AdjacencyMap.fromAdjacencySets . Map.toList) . shrink . AdjacencyMap.adjacencyMap
```

but I am a bit surprised by `shrink` behaviour on Map, i.e.

```
λ> shrink $ Map.fromList [(1, Set.empty)] :: [Map Int (Set Int)]
[fromList [],fromList [(0,fromList [])]]
```

basically it shrinks the label itself from one to zero.

andreabedini:
Another option could be

```
shrink2 = map
    (\vs -> induce (`List.elem` vs) g)
    (List.nub $ shrinkList pure $ vertexList $ g)
```

or, more efficiently

```
shrink2' g = map
    (\vs -> induce (`Set.member` vs) g)
    (List.nub $ map Set.fromList $ shrinkList pure $ vertexList $ g)
```

andreabedini:
> @andreabedini I think reusing the `shrink` that comes with `Map` or `Set` is a good idea. It's a pity we cannot directly reuse the `Map`'s one because of the issue you mention. Perhaps, we could adapt the one from `Set`?

The one from `Set` behaves in the same way, basically any structure get shrunk along with its components (integers get shrunk too). This seems to be the general philosophy in Quickcheck (from what I can learn from the sources). I think the big issue here is that our adjacency map is not an inductive type but has invariants, so the general approach to shrink might not be appropriate. (That said I start to believe there are no "wrong" implementation of shrink, perhaps only useless ones).

> Your `shrink2` seems to be quite limited in the sense that it only produces induced subgraphs, i.e. one cannot remove an edge only. Or am I mistaken?

That is correct, my bad.

Today I came up with this

```
    shrink g = lessVertices ++ lessEdges
      where
        vss = reverse $ subsequences (AdjacencyMap.vertexList g)
        ess = reverse $ subsequences (AdjacencyMap.edgeList g)
        lessVertices = [ foldr AdjacencyMap.removeVertex         g vs | vs <- vss ]
        lessEdges    = [ foldr (uncurry AdjacencyMap.removeEdge) g es | es <- ess ]
```

basically:

1. It removes vertices and then remove edges
2. Differently from the first implementation, it doesn't delete only one vertex but all subsets of vertices (idem for edges). All subsets might be overkill, but I kinda don't like how `shrinkList pure` behaves (perhaps because I don't understand it).
3. The `reverse` there is because I want to put stronger shrinkings first (those are the vertices to remove)
4. I checked this is the behaviour of `shrink (x, y)`: shrink first `x` with `y` constant, then shrink `y` with `x` constant.

snowleopard:
@andreabedini Thanks! Looks like the generic shrink for trees doesn't work for older GHCs (CI failed). 

By the way, in the current implementation, when the number of vertices is large you will never get to the part where you are removing edges: there are just too many combinations, and I'm sure QuickCheck will just give up trying at some point.

Let me come back to one of your earlier comments:

> Actually, I am realising I have been writing silly instances of shrink! I was under the impression that shrink was going to be applied recursively but not! therefore my implementation for AdjacencyMap only removes one vertex and then only removes one edge ... and it stops there.

Is this really true? My understanding, which is confirmed by [this comment](https://stackoverflow.com/a/16970029), is that QuickCheck will try all shrinks and if any of them works, it will then proceed recursively from it.

Why don't you make an experiment with a simpler `shrink` which tries to remove one vertex or one edge? Try the `verboseCheck` option to see how shrinking goes.

(Of course removing vertices one by one is sometimes not going to work: for example, if the only way to shrink is by removing two vertices together, but I think such properties are relatively rare.)

andreabedini:
> Looks like the generic shrink for trees doesn't work for older GHCs (CI failed).

oopsie, I didn't think of older GHCs.

> By the way, in the current implementation, when the number of vertices is large you will never get to the part where you are removing edges: there are just too many combinations, and I'm sure QuickCheck will just give up trying at some point.

This must be why the shrink implementation for lists is so aggressive (as in: it generates so few sublists). 

> Is this really true? My understanding, which is confirmed by [this comment](https://stackoverflow.com/a/16970029), is that QuickCheck will try all shrinks and if any of them works, it will then proceed recursively from it.

ok, I was misunderstanding the way `shrink` is used in the process of finding a minimal counterexample. `shrink` itself is not recursive (although one might implement it in such a way).
 
> Why don't you make an experiment with a simpler `shrink` which tries to remove one vertex or one edge? Try the `verboseCheck` option to see how shrinking goes.

It doesn't go very well 😂 

```
λ> shrink empty
[empty,empty]
```

andreabedini:
> One simple property you could be checking is `forall x. isEmpty x` -- it is obviously false, but I hope we can reduce any counterexample to a single vertex.

that's likely to fail with a tiny counterexample, graphs generated by arbitrary seems to have a long-tail distribution in size.

I'm still working on this, although I am bit sort of time at the moment.

snowleopard:
@andreabedini No rush! And thanks again for working on this.

andreabedini:
In d1cd4e0 I am suggesting the following (plus variations for the other adjacency-map implementations).

```
    shrink g = lessVertices ++ lessEdges
      where
        vertices       = AdjacencyMap.vertexList g
        verticesToKeep = shrinkList pure vertices
        lessVertices   = do
          vs <- verticesToKeep
          return $ foldr AdjacencyMap.removeVertex g (vertices \\ vs)

        edges       = AdjacencyMap.edgeList g
        edgesToKeep = shrinkList pure edges
        lessEdges   = do
          es <- edgesToKeep
          return $ foldr (uncurry AdjacencyMap.removeEdge) g (edges \\ es)
```

The approach is 1) consider removing some vertices 2) consider removing some edges.

Complexity should be terrible because I am relying on `shrinkList` to shrink the vertices, but then I need the complement of the resulting set to remove the vertices one by one. At least `shrinkList pure` only generates `O(n)` lists of length `O(n)` (where `n` is the length of the input list). So it doesn't grow exponentially.

PS:

I have spent quite some time trying to shrink directly the adjacency map representation. For the edges in principle it should not be a problem (since edges are represented independently), but the implementation is tricky because it seems to require a lot of wrapping/unwrapping.

To address this I wrote the following:

```
newtype Thingy a = Thingy { unThingy :: [a] }
  deriving (Eq, Show, Functor)

instance Applicative Thingy where
  pure x = Thingy [x]
  (Thingy _ ) <*> (Thingy []) = Thingy []
  (Thingy []) <*> (Thingy _ ) = Thingy []
  (Thingy (f:fs)) <*> (Thingy (x:xs)) = Thingy $ f x : [ f' x | f' <- fs ] ++ [ f x | x <- xs ]
```

Which allows one to write `traverse (Thingy . \x -> x : shrinkList pure x) . adjacencyList g` to produce edge-shrunk graphs in the order QuickCheck expects them.

This of course will never work for vertices so I gave up on that route.

PPS: Is `Thingy` know under another name? It's neither a `ZipList` nor the monadic applicative instance for lists.

snowleopard:
@andreabedini Apologies for slow responses, this and the next weeks are very busy.

Before looking into the implementation, I'd like to double check the plan. My understanding of the plan is:

* Given a graph `g` with `n` vertices and `m` edges, generate `n + m` shrink candidates:
   - `n` candidates, each corresponding to the original graph `g` with one of the vertices removed
   - `m` candidates, each corresponding to the original graph `g` with one of the edges removed

Do you agree with the plan? Or are you suggesting something different?

Regarding `Thingy`: it's not obvious that it satisfies the `Applicative` laws. Does it?

The `(Thingy (f:fs)) <*> (Thingy (x:xs))` case looks a bit strange. I think I've come across Applicative instances where this case was essentially `f x : Thingy fs <*> Thingy xs`

andreabedini:
> @andreabedini Apologies for slow responses, this and the next weeks are very busy.

No problem at all, it wasn't a slow response for my standards :-)

> Before looking into the implementation, I'd like to double check the plan. My understanding of the plan is:
> * Given a graph `g` with `n` vertices and `m` edges, generate `n + m` shrink candidates:
> Do you agree with the plan? Or are you suggesting something different?

I am suggesting a different plan.

The plan you mention corresponds to my initial implementation but I now think that plan would produce a large number of ineffective candidates, where the "effectivity" is measured by how aggressive the shrink is.

The plan I am suggesting (and have implemented) is to use `shrinkList` to shrink the list of nodes and edges, keeping only the graph induced by what's left in the list (by removing what's not in the produced shrink, which brings my comment about complexity).

> Regarding `Thingy`: it's not obvious that it satisfies the `Applicative` laws. Does it?

I don't have a formal proof but it does according to QuickCheck.

> I think I've come across Applicative instances where this case was essentially `f x : Thingy fs <*> Thingy xs`

I think that would be ZipList?

```
instance Applicative ZipList where
    pure x = ZipList (repeat x)
    liftA2 f (ZipList xs) (ZipList ys) = ZipList (zipWith f xs ys)
```

I appreciate taking your time to look at this @snowleopard, feel free to take your time!

snowleopard:
> The plan you mention corresponds to my initial implementation but I now think that plan would produce a large number of ineffective candidates, where the "effectivity" is measured by how aggressive the shrink is.

@andreabedini This definition of "effectivity" is not very precise. Could you elaborate it? What I'd like to have as is: 1) ability to shrink counterexamples to small sizes, 2) simplicity. I think without a few concrete examples it's almost impossible to discuss how effective one shrinking technique is compared to another, so let's create a couple of simple test cases. What about starting with these?

* ``forall x. (removeVertex 42 x) == x``
* ``forall x. (removeEdge 4 2 x) == x``
* ``forall x. not (path [3, 1, 4] `isSubgraphOf` x)``

I'd like to see how the simple approach (removing an vertex, and removing an edge) works here.

> I don't have a formal proof but it does according to QuickCheck.

Cool! At first I thought that your definition of `<*>` be simplified simply to Cartesian product, i.e. to something like `Thingy fs <*> Thingy xs = [ f x | f <- fs, x <- xs ]`, but I'm probably wrong, because in your implementation leading elements are treated differently from others.

andreabedini:
> @andreabedini This definition of "effectivity" is not very precise. Could you elaborate it?

By "effective" I meant it reduces the complexity by a very limited amount. Said otherwise, for each graph `g`, `shrink g` would be a list of graphs of "complexity" very close to `g` itself (since they are only missing either one vertex or one edge).

In my reading of the [documentation](http://hackage.haskell.org/package/QuickCheck-2.12.6.1/docs/Test-QuickCheck.html#v:shrink). I thought it would be important for a good implementation of shrink to produce, in one invocation, a list of values of increasing "complexity" from a minimal value to the original value. This is how I read the following sentence (from the page linked above): 

> QuickCheck tries the shrinking candidates in the order they appear in the list, so we put more aggressive shrinking steps (such as replacing the whole tree by Nil) before smaller ones (such as recursively shrinking the subtrees).

Anyway, you make a good point:

> What I'd like to have as is: 1) ability to shrink counterexamples to small sizes, 2) simplicity.

I have to admit I might have been overthinking this and that simplicity is an important factor to consider. So important that I have found bugs in my proposed code that lead QuickCheck into a loop. I was therefore only able to test the simple strategy (removing one vertex or one edge at the time) 😞.

> * `forall x. removeVertex 42 x == x`

I won't paste the full output but it's exactly what one would expect, after finding a counterexample (any graph including the vertex 42), it starts removing vertices one by one until it reaches `vertex 42`. The result record below shows it takes 84 steps.

```
λ> quickCheckResult $ forAllShrink arbitrary shrink (\x -> removeVertex 42 x == x)
*** Failed! Falsifiable (after 43 tests and 84 shrinks):
vertex 42
Failure {numTests = 43, numDiscarded = 0, numShrinks = 84, numShrinkTries = 0, numShrinkFinal = 1, usedSeed = TFGenR 9ABF2DDED6F793A1B82E42C2B86EB51222F027C26760B8A150839801D60C2886 0 4398046511103 42 0, usedSize = 42, reason = "Falsifiable", theException = Nothing, output = "*** Failed! Falsifiable (after 43 tests and 84 shrinks):\nvertex 42\n", failingTestCase = ["vertex 42"], failingLabels = [], failingClasses = fromList []}
```

> * `forall x. removeEdge 4 2 x == x`

Same as above, the simple it reaches `edge 4 2` after 46 shrinks.

```
λ> quickCheckResult $ forAllShrink arbitrary shrink (\x -> removeEdge 4 2 x == x)
*** Failed! Falsifiable (after 24 tests and 46 shrinks):
edge 4 2
Failure {numTests = 24, numDiscarded = 0, numShrinks = 46, numShrinkTries = 42, numShrinkFinal = 3, usedSeed = TFGenR 974D0E1D8A9FE7A0196F01C17C74D85A660ACD515F6B4059F895BBC4ABCD2492 0 2147483520 31 0, usedSize = 23, reason = "Falsifiable", theException = Nothing, output = "*** Failed! Falsifiable (after 24 tests and 46 shrinks):\nedge 4 2\n", failingTestCase = ["edge 4 2"], failingLabels = [], failingClasses = fromList []}
```

> * `` forall x. not (path [3, 1, 4] `isSubgraphOf` x) ``

This is more interesting as the counterexamples are rare enough that the probability to find one in 100 tests is quite small. Nevertheless, when a counterexample is found, everything goes smoothly and quickcheck converges to `edges [(1,4),(3,1)]` in 165 steps.

```
λ> quickCheckResult $ forAllShrink arbitrary shrink (\x -> not (path [3, 1, 4] `isSubgraphOf` x))
*** Failed! Falsifiable (after 84 tests and 165 shrinks):
edges [(1,4),(3,1)]
Failure {numTests = 84, numDiscarded = 0, numShrinks = 165, numShrinkTries = 243, numShrinkFinal = 5, usedSeed = TFGenR 9E0B8A4105B05E9BF998BE11C5B2682FC8780F066688995334D3CE9A68D7A81B 0 140737488355327 47 0, usedSize = 83, reason = "Falsifiable", theException = Nothing, output = "*** Failed! Falsifiable (after 84 tests and 165 shrinks):\nedges [(1,4),(3,1)]\n", failingTestCase = ["edges [(1,4),(3,1)]"], failingLabels = [], failingClasses = fromList []}
```

If this looks good enough (it does to me), I'll revert this PR to the simple one-vertex-one-edge strategy.

> Cool! At first I thought that your definition of `<*>` be simplified simply to Cartesian product, i.e. to something like `Thingy fs <*> Thingy xs = [ f x | f <- fs, x <- xs ]`, but I'm probably wrong, because in your implementation leading elements are treated differently from others.

I am hoping to have time to think about this a big more and maybe explain on a blog post how I came to it.

PS: apologies if I abused of your patience and congrats on the selective paper! very interesting.

snowleopard:
@andreabedini Thanks for running the experiments! It looks like the simple implementation worked great. It produced very simple counterexamples for all three test properties, even though the counterexamples that were found initially were pretty big. Let's to ahead with this simple solution.

> PS: apologies if I abused of your patience and congrats on the selective paper! very interesting.

No problem at all, I was actually quite busy exactly because of the paper on selective functors :) I'm glad you came across it, and found it interesting!

andreabedini:
@snowleopard I have implemented the shrinks for 

- AdjacencyMap
- AdjacencyIntMap
- Labelled.AdjacencyMap
- NonEmpty.AdjacencyMap

using the simple remove one-vertex one-edge strategy we discussed. 

I didn't implement shrink for `Doc s` because `Algebra.Graph.Export` doesn't export any way to deconstruct a value of that type (as far as I understand).

snowleopard:
@andreabedini Looks great to me, many thanks! 

I'm happy to merge as is, but we could also use the same implementations for `Fold` and `Relation`. 

Let me know if you prefer this PR to be merged or would like to do anything else.

andreabedini:
@snowleopard I added an shrink implementation for Fold and Relation. It's worth noticing that I had to add a `Eq a, Ord a` constraint to `Arbitrary (Fold a)`.

snowleopard:
Thanks again @andreabedini!  

Looking at the final implementation, I can see quite a bit of code duplication which can be avoided by implementing something like `shrinkViaToGraph :: ToGraph g => ...` and then using it in various `Arbitrary` instances. This is not critical, but I thought you might want to give it a try. If not, I'm happy to merge as is.

snowleopard:
Actually, I take it back. `removeVertex` is not part of the `ToGraph` type class. It looks like we might need to add another type class for this purpose :)

Merged!

PR #164, opened 2019-02-04
add Applicative and Monad instances for Label

subttle:


snowleopard:
Thanks @subttle! While the instances look reasonable, it's not obvious to me that Applicative and Monad laws hold. Could you give some intuition about why you chose this particular implementation, and why it satisfies the expected laws?

subttle:
Hi, thank you for considering this PR! While it is not a substitute for a proof, I did use quick check 
```

snowleopard:
@subttle Strange: I received an email notification with your comment about using QuickCheck, but it doesn't seem to be here. Have you deleted it, or is it a GitHub glitch?

Regarding using QuickCheck: this requires an `Eq Label` instance, but we don't have one. Which one did you use?

snowleopard:
Thanks @subttle! One simple thing to do to at least resolve the question about the validity of the Applicative instance is to use `(<*>) = ap` as the implementation. This is required by Applicative-Monad laws anyway, plus makes the implementation simpler.

PR #167, opened 2019-02-13
Adds instance up to Dioid for Minimum.

dalaing:


snowleopard:
@dalaing This is merged, too, thanks again! No idea how I missed this instance :)

PR #168, opened 2019-02-13
Give Graph Bifunctor on GHC 7.8 too

gwils:
Replace CPP with a conditional `bifunctors` dependency

### DIFF
--- a/algebraic-graphs.cabal
+++ b/algebraic-graphs.cabal
@@ -107,6 +107,8 @@ library
                         mtl         >= 2.1     && < 2.3
     if !impl(ghc >= 8.0)
         build-depends:  semigroups  >= 0.18.3  && < 0.18.4
+    if !impl(ghc >= 7.10)
+        build-depends:  bifunctors  >= 5       && < 5.6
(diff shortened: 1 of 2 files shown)


snowleopard:
@gwils This looks good, thanks you! I was actually wondering if I need to continue to support GHC 7.8, but your PR means you still care about it...

gwils:
I don't use 7.8, but I find it difficult to work with APIs which are conditional on GHC version like this. This PR was a very easy change for me to make that could save some theoretical 7.8 user from that pain :)

I would be equally as happy with you dropping 7.8 support instead of merging this PR.

snowleopard:
I see, I'll keep this in mind. Merged now!

dalaing:
Would there be interest in a `Bifoldable` instance as well?

snowleopard:
@dalaing No, because Foldable instances are in general problematic, see the top comment here: https://hackage.haskell.org/package/algebraic-graphs-0.3/docs/Algebra-Graph-ToGraph.html

dalaing:
Ah, yeah - fair enough :)

PR #171, opened 2019-02-18
Implement sparsification into King-Launchbury graph representation

snowleopard:


snowleopard:
@nobrakal Strangely, it looks like I somehow broken your rewrite rules. I guess `OverloadedLists` could be the reason?

snowleopard:
Indeed, the rewrite rules got fixed by getting rid of `OverloadedLists`.

PR #177, opened 2019-03-16
Allow project to be developed using Nix

srid:
Some Haskell developers like myself use **Nix** to manage their environment. This change makes it easier to use Nix to get the project up and running. The `nix-shell --run ghcid` command for example will pull the dependencies (including ghc itself!), and launch `ghcid` which is extremely handy for iterative compilation feedback from edits.

I have also increased the version (referring to a yet to be published version) in default.nix so that nix does not accidentally pick the published version from hackage.

Ref:
* https://nixos.org/nixpkgs/manual/#how-to-create-a-development-environment
* https://twitter.com/puffnfresh/status/990154797494943744?lang=en

### DIFF
--- a/algebraic-graphs.cabal
+++ b/algebraic-graphs.cabal
@@ -1,5 +1,5 @@
 name:          algebraic-graphs
-version:       0.3
+version:       0.4
 synopsis:      A library for algebraic graph construction and transformation
 license:       MIT
 license-file:  LICENSE
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,1 @@
+(import <nixpkgs> { }).haskellPackages.developPackage { root = ./.; }


snowleopard:
Thanks @srid! I'm unsure if it's worth committing this file to the repository, because it's completely generic, i.e. has no relation to the library itself. What about adding it to `.gitignore` instead? It would allow you and other Nix users build the library with Nix by just creating this file.

Bumping the version to `0.4` is a good idea, let's do that.

srid:
Sure, that's fine (the `default.nix` can grow depending on custom dependencies used though).

snowleopard:
> the default.nix can grow depending on custom dependencies used though

Thanks, I see. In this case I'll be happy to add it!

srid:
Yea. Here's an example, where my app is using your library but using a local checkout (with changes) with tests disabled:

```nix
let
  pkgs = import <nixpkgs> { };
  H = pkgs.haskellPackages;
  dontCheck = pkgs.haskell.lib.dontCheck;
in H.developPackage {
  root = ./.;
  overrides = self: super: {
    algebraic-graphs = dontCheck (H.callCabal2nix "algebraic-graphs" dep/algebraic-graphs {});
  };
}
```

I could also have it point to a fork (instead of local directory) by using `fetchFromGithub` ([example here](https://github.com/jwiegley/nix-config/blob/master/overlays/10-haskell.nix#L60-L66)).

All of this is only useful if a dev is using Nix for the Haskell workflow.

snowleopard:
My worry is that I won't be able to maintain any sophisticated Nix scripts, because I'm not a Nix user. It's probably better to not have any Nix scripts at all than to have a Nix script that's occasionally out-of-date.

PR #178, opened 2019-03-16
Derive Generic for AdjacencyMap

srid:


src/Algebra/Graph/NonEmpty/AdjacencyMap/Internal.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE DeriveGeneric #-}
```
snowleopard: I'm not sure, but it seems to me that here we derive `Generic` using `GeneralisedNewtypeDeriving`, which is enabled by default. Could you please check?

snowleopard:
@srid Thank you! Shall we do this for non-empty adjacency maps too?

src/Algebra/Graph/NonEmpty/AdjacencyMap/Internal.hs
srid: FWIW here's what I get without `DeriveGeneric`:

```
src/Algebra/Graph/AdjacencyMap/Internal.hs:136:51-57: error:
    • Can't make a derived instance of ‘Generic (AdjacencyMap a)’
        (even with cunning GeneralizedNewtypeDeriving):
        You need DeriveGeneric to derive an instance for this class
    • In the newtype declaration for ‘AdjacencyMap’
    |
136 |     adjacencyMap :: Map a (Set a) } deriving (Eq, Generic)
    |                                                   ^^^^^^^

```

srid:
@snowleopard I've added Generic instances for the other two adjacency maps.

src/Algebra/Graph/NonEmpty/AdjacencyMap/Internal.hs
snowleopard: Aha, I see, thanks for checking!

snowleopard:
@srid Great, thank you. Merged!

@nobrakal It looks like the performance suite is broken again. Could you please have a look when you have time?

nobrakal:
@snowleopard  Indeed the new `0.4` version prevented the build. This is corrected :)

snowleopard:
Thanks @nobrakal!

PR #188, opened 2019-04-13
Single Source Shortest Path

adithyaov:
A generic implementation of the single source shortest path algorithm as described in the paper **Semiring frameworks and algorithms for shortest-distance problems** by **Mehryar Mohri**. I believe there is still work to be done, but I wanted to get the conversation started on this.


snowleopard:
Apologies for taking so long to review. Busy times!

adithyaov:
Hi, It would be great if this could be reviewed quickly. I am planning to write a test suite once the initial review is done. As far as I've checked, the algorithm seems to be working.

adithyaov:
Hi, Could you also mention some guidelines for coding, documentation, naming, etc? I agree that the code is a little hard to understand. Maybe I should document it better to make the job easy for the reviewers.

src/Algebra/Graph/Labelled/AdjacencyMap/Algorithm/Dijkstra.hs
```diff
@@ -0,0 +1,109 @@
+module Algebra.Graph.Labelled.AdjacencyMap.Algorithm.Dijkstra where
+
+import Data.Map.Strict (Map, (!))
+import Algebra.Graph.Label (Semiring(..), (<+>), zero)
+import qualified Data.Map.Strict as Map
+import Algebra.Graph.Labelled.AdjacencyMap.Internal (AdjacencyMap(..))
+
+-- | Psuedocode for Generic-Single-Source-Shortest-Distance (G,s)
```
snowleopard: No need to provide pseudocode of a classic algorithm in Haddock, better link to the Wikipedia and the paper you referred to previously. Users need a high-level description and examples.

The text you have is still useful, so don't delete it, but move to a separate standalone comment: it is useful for *developers*, not users of the library.

src/Algebra/Graph/Labelled/AdjacencyMap/Algorithm/Dijkstra.hs
```diff
@@ -0,0 +1,109 @@
...
+-- w[e] is the weight of the edge e
+--
+-- Brief:
+-- We use a queue q to maintain the set of vertices whose leaving edges are to be relaxed. q is initialized to [source].
+-- We maintain 2 attributes, namely, `d` and `r`.
+-- d[p] an estimate of the shortest distance from s to p
+-- r[p] is the total weight added to d[q] since the last time p was extracted for q.
+-- We update d and r till the queue is empty.
+
+dijkstra ::
```
snowleopard: The documentation misses examples with corresponding tests, see an example here:

https://hackage.haskell.org/package/algebraic-graphs-0.3/docs/Algebra-Graph-AdjacencyMap-Algorithm.html#v:dfs

Also, what are the time/memory complexity of your implementation?

src/Algebra/Graph/Labelled/AdjacencyMap/Algorithm/Dijkstra.hs
```diff
@@ -0,0 +1,109 @@
...
+     (Ord a, Eq b, Semiring b) => a -> AdjacencyMap b a -> Map a b
```
snowleopard: We use `e` for the type of edge labels in the rest of the codebase.

src/Algebra/Graph/Labelled/AdjacencyMap/Algorithm/Dijkstra.hs
```diff
@@ -0,0 +1,109 @@
...
+dijkstra src adjMap = fst3 fDRQ
+  where
+    iDRQ = initialize src adjMap
+    mDRQ = recursiveUpdateByQ adjMap iDRQ
+    fDRQ = finalize src mDRQ
+    fst3 (x, _, _) = x
+
+initialize ::
```
snowleopard: Please add comments, explaining how this and other functions work.

snowleopard:
@adithyaov Thanks, I've added some comments -- mostly regarding the current state of the documentation, examples and tests. 

> Could you also mention some guidelines for coding, documentation, naming, etc? 

If you just look at the code in the rest of the project, you'll see that every exported function comes with a lot of examples and tests. Please do the same for your module.

> Strictly speaking, This is not Dijkstra. This is a generic single source shortest path algorithm! I should probably rename the module, I'd like some suggestions on this.

Hmm, but the plan was to implement the classic Dijkstra algorithm. It would be a good start because everyone knows what it is, and there are existing implementations which can be used for benchmarking. I'd prefer to stick to Dijkstra. Implementing it on top of a semiring shouldn't be very hard.

adithyaov:
> Hmm, but the plan was to implement the classic Dijkstra algorithm.

Hi, Thanks for the review. Please note the following details.

The only difference between Dijkstra and GSSSD (Generic-Single-Source-Shortest-Distance) algorithm is that Dijkstra uses a priority queue whereas GSSSD uses a normal queue.

Even the paper quotes that **Dijkstra is a special case of the Generic-Single-Source-Shortest-Distance algorithm where the semiring `K` is taken as the tropical semiring `T` and where the queue discipline is based on the shortest-first order.** (Tropical semiring is (R+ ∪ {∞}, min, +, ∞, 0))

I could implement the Dijkstra algorithm and choose to use either min-heap or max-heap based on the arity of the semiring, which would allow the algorithm to work on generic semirings (not only Tropical). Please let me know your thoughts on this.

src/Algebra/Graph/Labelled/AdjacencyMap/Algorithm/Dijkstra.hs
```diff
@@ -0,0 +1,109 @@
...
+            cDRQ = (d, newR, rest)
+
+updateDRQByEdgeMap ::
+     (Ord a, Eq b, Semiring b)
+  => b
+  -> Map a b
+  -> (Map a b, Map a b, [a])
+  -> (Map a b, Map a b, [a])
+updateDRQByEdgeMap r' edgeMap pDRQ =
+  foldr (\k dRQ -> updateDRQByEdge r' (k, edgeMap ! k) dRQ) pDRQ $
```
nobrakal: Why using `Map.keys` and then doing a `foldr`? Can not this be done by a single traversal using `Map.foldrWithKey`?

src/Algebra/Graph/Labelled/AdjacencyMap/Algorithm/Dijkstra.hs
```diff
@@ -0,0 +1,109 @@
...
+updateDRQByEdge r' (k, w) dRQ@(d, r, queue) =
+  if pDK /= nDK
+    then (Map.insert k nDK d, Map.insert k nRK r, nQ)
+    else dRQ
+  where
+    pDK = d ! k
+    nDK = pDK <+> (r' <.> w)
+    nRK = (r ! k) <+> (r' <.> w)
+    nQ =
+      if k `elem` queue
```
nobrakal: Here too, I think this can be done with a single traversal of the queue, but I don't know if it really worth the change.

snowleopard:
> The only difference between Dijkstra and GSSSD (Generic-Single-Source-Shortest-Distance) algorithm is that Dijkstra uses a priority queue whereas GSSSD uses a normal queue.

Doesn't it mean that GSSSD has different time complexity?

This brings me to my question in the code that you didn't answer:

> Also, what are the time/memory complexity of your implementation?

src/Algebra/Graph/Labelled/AdjacencyMap/Algorithm/Dijkstra.hs
adithyaov: > Can not this be done by a single traversal using Map.foldrWithKey?

Yes, I'll change that. It is better to use `Map.foldrWithKey`.

adithyaov:
Hi, I will resolve the discussions, but I might have to change the algorithm completely. As mentioned before this isn't strictly Dijkstra.

The time complexity for Dijkstra is `O(E + V * log(V))`

The time complexity for GSSSD is `O(V + E * (max v ∈ V : N(v)) + ∑ (v ∈ V : N(v)))`
N(v) is the number of times a vertex `v` is inserted back into the queue.
Note: I've simplified the time complexity from that mentioned in the paper, taking T⊕ = O(1) and T⊗ = O(1), etc.

**The time complexity is Dijkstra is less than that of GSSSD as we use a priority heap instead.**

I believe I should create a new pull request with the Dijkstra algorithm instead.

src/Algebra/Graph/Labelled/AdjacencyMap/Algorithm/Dijkstra.hs
adithyaov: Thank you for the review! I'll look into it and get back to you.

adithyaov:
@snowleopard 
#194 Please go through this pull request and let me know what you think. This is exactly Dijkstra.

adithyaov:
This PR has become rather messy. Closing it as of now.

PR #190, opened 2019-04-21
Update CI to test with GHC 8.6.4

snowleopard:


snowleopard:
@nobrakal Looks like I've broken the performance suite, again :( Could you please fix it?

PR #174, opened 2019-02-19
Added functionality to the 'SymmetricRelation' module.

bolt12:
### Functionalites (based on `Algebra.Graph.Relation`):
- Basic graph construction primitives
- Relations on graphs
- Graph properties
- Standard families of graphs
- Graph transformation
- Relational operations (only `compose` yet)
These new functions respect the undirected graph axiom (commutativity of
'connect').

Adjusted the implementation of the `SymmetricRelation` `Graph` instance
to respect the axiom.

### Problems: 

`Num` instance for `SymmetricRelation` was derived so the
axiom was not guaranteed for `(*)`, so I added a custom instance
implementation on `Algebra.Graph.Relation.Symmetric` which the compiler
gives an error for duplicate declaration of the `Num` instance. I
already deleted the `deriving Num` from the `Relation.InternalDerived`
module but the compiler still gives the error.

PR #191, opened 2019-04-22
Symmetric relation fixes

snowleopard:
A minor revision after #174.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## 0.4
 
+* #174: Add `Symmetric.Relation`.
 * #143: Allow newer QuickCheck.
 * #171: Implement sparsification for King-Launchbury graph representation.
 * #178: Derive `Generic` for adjacency maps.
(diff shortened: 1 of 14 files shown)


snowleopard:
@bolt12 Just to let you know: I've done some further work on symmetric relations -- you might be interested to have a look!

bolt12:
Alright! Very interesting changes, why do you said that about the `ToGraph` instance I didn't understand what you mean.

snowleopard:
> why do you said that about the `ToGraph` instance I didn't understand what you mean.

@bolt12 This `ToGraph` instance seems to be incompatible with others because of the way we treat symmetric edges. In particular, [we expect](http://hackage.haskell.org/package/algebraic-graphs-0.3/docs/Algebra-Graph-ToGraph.html#v:edgeSet) that the following property holds:

```haskell
edgeSet == Algebra.Graph.AdjacencyMap.edgeSet . toAdjacencyMap
```

But it doesn't (and we don't test it in the testsuite).

In fact, maybe we need a separate type class `Algebra.Graph.ToGraph.Undirected` or something similar for working with undirected graphs, since they are quite different from directed graphs in terms of the typical operations we apply to them.

bolt12:
Ohh I didn't had that in mind when I wrote that instance.. my bad for not
understanding that it had to obey that property.. If I have time I'll think
of an option for this particular case and add it to the next contribution!

snowleopard:
No problem! Maybe it's worth looking around for other existing APIs for working with undirected graphs.

bolt12:
I will! Probably will fit nicely with the idea of adding support for more
functions like `isConnected`, `nodeDegree`, etc..

PR #192, opened 2019-04-22
Fix the benchmarking suite

nobrakal:
Hi,

These commits fix the benchmarking suite. The fail was due to the changes made in `.travis.yml`.
I changed accordingly the script in my repository ( https://github.com/nobrakal/benchAlgaPr ), don't hesitate to pull if you use it locally :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -104,7 +104,7 @@ install:
   - if [ "$DOBENCH" != "" ]; then
         echo "Downloading performance regression suite...";
         pwd;
-        curl https://raw.githubusercontent.com/nobrakal/benchAlgaPr/debug/compare.sh -o compare.sh;
+        curl https://raw.githubusercontent.com/nobrakal/benchAlgaPr/master/compare.sh -o compare.sh;
         chmod +x compare.sh;
     fi
 ##### Performance regression suite, added manually #####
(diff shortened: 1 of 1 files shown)


snowleopard:
@nobrakal Thanks! Looks like it currently times out though...

nobrakal:
Yes, this is due to the update of GHC, which I think made the dependencies to be rebuilt. If you just resubmit de job, it should pass now (and with caching, will pass for next builds).

I will try to refactor the whole benchmarking library during the summer (to use the full power of Criterion and drop some uneeded dependencies). This should, I hope, reduce the build and running time :)

snowleopard:
Hmm, I restarted Travis, but it still timed out.

snowleopard:
@nobrakal Or did you mean something else by "resubmit the job"?

nobrakal:
That is strange, I will investigate.

nobrakal:
Well, I was able to have a success on my fork, so I guess try to clear the travis cache and resubmit the job (maybe two times will be required).
Anyway, I think there are several optimizations to do to this benchmarking suite, I hope I will be able to simplify all of this :)

snowleopard:
@nobrakal Thanks, it worked! :)

PR #194, opened 2019-05-01
Made the new version of Dijkstra Algorithm.

adithyaov:
This is a response to the feedback given to the PR for Generic single source shortest distance. This version is Dijkstra and the running time is similar to that of Dijkstra (O(V + E * log (V)) as we used a priority Heap.

The usage should be like follows:
`dijkstra a am :: DijkstraForDistance Int Int`

There are 2 types alias' defined, DijkstraForDistance and DijkstraForCapacity. One should use these types to get the algorithm working as we used a heap.

I am unable to deduce the heap policy from the semiring. Hence, for any semiring `X` that dijkstra should work on, there needs to be a new type alias `DijkstraForX`.
Maybe I can deduce the heap policy from the semiring using DataKinds or reworking the heap but that seems like overkill.

Please let me know if this kind of usage seems fine. I'll comment and write up the documentation as soon as possible.

### DIFF
--- a/src/Algebra/Graph/Labelled/AdjacencyMap/Algorithm/Dijkstra.hs
+++ b/src/Algebra/Graph/Labelled/AdjacencyMap/Algorithm/Dijkstra.hs
@@ -1,0 +1,100 @@
+module Algebra.Graph.Labelled.AdjacencyMap.Algorithm.Dijkstra where
+
+import Algebra.Graph.Label (Capacity, Distance, Semiring(..), (<+>), zero)
+import Algebra.Graph.Labelled.AdjacencyMap.Internal (AdjacencyMap(..))
+import Control.Monad (sequence_, when)
+import Control.Monad.Trans.State.Strict (State, execState, get, gets, put)
+import Data.Heap (FstMaxPolicy, FstMinPolicy, Heap, HeapItem(..))
+import qualified Data.Heap as Heap
+import Data.Map (Map, (!))
+import qualified Data.Map.Strict as Map
+
+type DijkstraForDistance a b = DijkstraState FstMinPolicy a (Distance b)
+
+type DijkstraForCapacity a b = DijkstraState FstMaxPolicy a (Capacity b)
+
+dijkstra ::
+     (Eq e, Ord a, HeapItem p (e, a), Semiring e)
+  => a
+  -> AdjacencyMap e a
+  -> DijkstraState p a e
+dijkstra a am =
+  flip execState (DijkstraState Heap.empty Map.empty Map.empty) $ do
+    initialize a am
+    exploreVertices am
+
+data DijkstraState p a e = DijkstraState
+  { heap :: Heap p (e, a)
+  , distance :: Map a e
+  , path :: Map a (Maybe a)
+  }
+
+instance (Show e, Show a) => Show (DijkstraState p a e) where
+  show (DijkstraState heap' distance' path') =
+    "distance: " ++ show distance' ++ "\n" ++ "path: " ++ show path'
+
+putHeap :: Heap p (e, a) -> State (DijkstraState p a e) ()
+putHeap h = do
+  (DijkstraState _ distance' path') <- get
+  put $ DijkstraState h distance' path'
+
+initialize ::
+     (Ord a, HeapItem p (e, a), Semiring e)
+  => a
+  -> AdjacencyMap e a
+  -> State (DijkstraState p a e) ()
+initialize a am =
+  put $
+  DijkstraState
+    (Heap.singleton (one, a))
+    (Map.insert a one . Map.map (const zero) $ adjacencyMap am)
+    (Map.map (const Nothing) $ adjacencyMap am)
+
+exploreVertices ::
+     (Ord a, Eq e, HeapItem p (e, a), Semiring e)
+  => AdjacencyMap e a
+  -> State (DijkstraState p a e) ()
+exploreVertices am = do
+  curHeap <- gets heap
+  case Heap.view curHeap of
+    Nothing -> return ()
+    (Just ((_, v), t)) -> do
+      putHeap t
+      exploreVertex v am
+      exploreVertices am
+
+exploreVertex ::
+     (Ord a, Semiring e, Eq e, HeapItem p (e, a))
+  => a
+  -> AdjacencyMap e a
+  -> State (DijkstraState p a e) ()
+exploreVertex v1 am =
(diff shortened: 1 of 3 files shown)


adithyaov:
The problem for this implementation is that the time complexity of Dijkstra is O(|V| * |E|) as the current heap library does not provide a way to `delete` or `modify` a value. The way I handle a `modify` is to `insert` after a `filter` operation. Is there a better way to achieve this?

adithyaov:
I wonder why Travis CI is failing for ghc-7.10.3 C Linux, ghc-7.10.4 C Linux, and ghc-8.6.4 C Mac. Is there some problem with the `heap` module used?
I tried reading the Job log but that is not getting me anywhere.
The only dependency for `heap` is `base >= 4.6 && < 5` which is satisfied in all the versions of ghc mentioned above according to [version history](https://gitlab.haskell.org/ghc/ghc/wikis/commentary/libraries/version-history)

snowleopard:
@adithyaov Thanks for further experiments! 

In general, I'd prefer to avoid adding any new dependencies to Alga, precisely because of various compatibility and maintenance issues, which you seem to be hitting with older GHC versions.

But what bothers me even more is that you say:

> The problem for this implementation is that the time complexity of Dijkstra is O(|V| * |E|) 

But that's terribly slow. Even the most basic Dijkstra implementation that does not use any heaps runs in just O(|V|^2). Why do you need to depend on `heap` then?

adithyaov:
> I'd prefer to avoid adding any new dependencies to Alga

Should I try creating an implementation of priority queue which depends on the semiring (G, <.>, <+>, 0, 1) using the (G, 0, <+>) monoid?

> But that's terribly slow. Even the most basic Dijkstra implementation that does not use any heaps runs in just O(|V|^2). Why do you need to depend on heap then?

That is because of the heap implementation. The package `pqueue` provides an O(log|N|) deletion method. Using that package with the same code will result in O(|V|) + O(E * log|V|).

adithyaov:
I've significantly improved the time complexity (I had a redundant function which was not required).
The current time complexity is `O(|V| + |E| * log|V|)`

I need to address the problem of avoiding heap dependency.
I can try to create an implementation of Heap.
If I need to, could you let me know how the module structure should look like?

src/Algebra/Graph/Labelled/AdjacencyMap/Algorithm/Dijkstra.hs
```diff
@@ -0,0 +1,106 @@
...
+  -> AdjacencyMap e a
+  -> DijkstraState p a e
+dijkstra a am =
+  flip execState (DijkstraState Heap.empty Map.empty Map.empty Map.empty) $ do
+    initialize a am
+    exploreVertices am
+
+data DijkstraState p a e = DijkstraState
+  { heap :: Heap p (e, a)
+  , explored :: Map a Bool
```
adithyaov: This is only used for further optimization. The time complexity would remain the same even if this was removed.

snowleopard:
> Should I try creating an implementation of a priority queue?

@adithyaov Another option would be to create a separate package with some graph algorithms, which would depend both on `algebraic-graphs` and on `heap`. 

For now, however, feel free to experiment in this PR (or in multiple PRs).

What I find useful personally is to start with the simplest possible implementation that you can write, then write examples/documentation/tests/benchmarks, **and only afterwards start to optimise** by switching to more advanced data structures, etc. In this way you always have a simple baseline implementation to compare against. Not all optimisations which sound promising in theory actually give you much benefit in practice.

src/Algebra/Graph/Labelled/AdjacencyMap/Algorithm/Dijkstra.hs
snowleopard: See my comment below about avoiding premature optimisation. I would start with the simplest implementation first, because it's easier to implement correctly, easier to review, easier to integrate into the library, etc.

adithyaov:
Hi, Thanks for the reviews!

One thing that bothers me is that the only way to use the current algorithm is to use the type alias' `DijkstraForDistance` and `DijkstraForCapacity`. The reason these exist is that I want to restrict the user to use the proper priority heap depending on the semigroup.

Is there a way to force a priority heap based on the semigroup? I tried a few implementations but it's not getting me anywhere.

And about the CI failing. I'll probably try different packages that provide the heap data structure.
Maybe I should create a class, say `PriorityQDijkstraGeneric` in the Dijkstra module which would contain all the basic operations for better abstraction?
Or as you said, making a separate package seems like a good idea too!

> Maybe I should create a class, say `PriorityQDijkstraGeneric` in the Dijkstra module which would contain all the basic operations for better abstraction?

I think this library should have a standard for this! Algorithms use wide variety of Data Structures and for implementing algorithms one should probably follow some standards.
Maybe there should be a module `Data.Graph.StructureStandards` which provides type classes such as `PriorityHeap`, `RedBlackTree`, etc and one can choose to use any library out there but should create instances of these classes.

snowleopard:
@adithyaov Please see my comment above about simplicity. Why don't you start with an implementation that works only for distances, and uses a simple priority queue based on `Data.Set`?

adithyaov:
I wonder why this error occures in ghc-7.8.4 C Linux,
```
    Couldn't match expected type ‘[Control.Monad.Trans.State.Strict.StateT
                                     (DijkstraState a (Distance e))
                                     Data.Functor.Identity.Identity
                                     a0]’
                with actual type ‘Map a (State (DijkstraState a (Distance e)) ())’
```
I believe there is some problem with usage of `sequence_`. It should work on any `Traversable` element.

adithyaov:
In the current Dijkstra algorithm (simplest version), I'm using the State Monad. How do you suggest I write examples? Currently, I'm using some random notation mentioned below,
```
old |> State before the operation
operation
new |> State after the operation
```
Is there a better notation I could use?

snowleopard:
> In the current Dijkstra algorithm (simplest version), I'm using the State Monad. How do you suggest I write examples? Currently, I'm using some random notation mentioned below,

@adithyaov The top-level interface should just use plain functions. There is no need to expose the State monad to the user, who simply wants to get a mapping from vertices to distances.

Also, we don't need to do anything about paths in the simple version: paths will be handled as just another semiring!

> I believe there is some problem with the usage of `sequence_`. 

Maybe you need to define your own `sequence_` or just figure out how to avoid using it.

adithyaov:
Another PR will address this with a cleaner implementation.

PR #197, opened 2019-05-12
Simplify adjacency map conversions using Coercible

snowleopard:


snowleopard:
@bolt12 You may be interested to have a look how `coerce` can help us get rid of many calls to conversion functions like `fromSymmetric`.

bolt12:
I'm going to have a look!

bolt12:
@snowleopard It appears that the coerce calls on the symmetric is already done, right? I'm sorry that took this long for me to come up to this issue.

snowleopard:
@bolt12 Yes, I already did this, as soon as I realised we can use `coerce` here.

PR #199, opened 2019-05-15
Basic Acyclic AdjacencyIntMap

adithyaov:
The idea is that there should be no edge from a higher valued vertex to a lower valued vertex. This is the basic edge rule. If the rule breaks there will be a compilation error.
This is the most basic implementation and is open to constructive debate.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@ cabal.sandbox.config
 cabal.project.local
 ghcid.txt
 .vscode/
+*.swp
+*.swo
--- a/algebraic-graphs.cabal
+++ b/algebraic-graphs.cabal
@@ -68,6 +68,7 @@ source-repository head
 library
     hs-source-dirs:     src
     exposed-modules:    Algebra.Graph,
+                        Algebra.Graph.Acyclic.AdjacencyIntMap,
                         Algebra.Graph.AdjacencyIntMap,
                         Algebra.Graph.AdjacencyIntMap.Algorithm,
                         Algebra.Graph.AdjacencyIntMap.Internal,
--- a/src/Algebra/Graph/Acyclic/AdjacencyIntMap.hs
+++ b/src/Algebra/Graph/Acyclic/AdjacencyIntMap.hs
@@ -7,40 +7,32 @@
 
 module Algebra.Graph.Acyclic.AdjacencyIntMap where
 
-import Data.IntMap.Strict (IntMap)
-import qualified Data.IntMap.Strict as IntMap
-import Data.IntSet (IntSet)
-import qualified Data.IntSet as IntSet
+import qualified Algebra.Graph.NonEmpty.AdjacencyMap as NAM
 import Data.Proxy (Proxy(..))
 import GHC.TypeLits
 
 data Edge (f :: Nat) (t :: Nat) =
   Edge
 
-newtype AdjacencyIntMap (edgesFollowRule :: Bool) =
-  AM (IntMap IntSet)
+newtype AdjacencyIntMap =
+  AM (NAM.AdjacencyMap Int)
   deriving (Show)
 
-empty :: AdjacencyIntMap 'True
-empty = AM IntMap.empty
+singleton :: Int -> AdjacencyIntMap
+singleton x = AM $ NAM.vertex x
 
-singleton :: Int -> AdjacencyIntMap 'True
-singleton x = AM $ IntMap.singleton x IntSet.empty
-
-addEdge ::
-     forall f t. (KnownNat f, KnownNat t)
+edge ::
+     forall f t. (KnownNat f, KnownNat t, (f <=? t) ~ 'True)
   => Edge f t
-  -> AdjacencyIntMap 'True
-  -> AdjacencyIntMap (f <=? t)
-addEdge _ (AM m) =
-  AM $ IntMap.insertWith IntSet.union fI (IntSet.singleton tI) m
+  -> AdjacencyIntMap
+edge _ = AM $ NAM.edge fI tI
   where
     fI = fromIntegral . natVal $ (Proxy :: Proxy f)
     tI = fromIntegral . natVal $ (Proxy :: Proxy t)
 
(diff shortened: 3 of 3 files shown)


snowleopard:
Thanks! Here is some feedback:

I think it may be more productive to follow the approach from `NonEmpty.AdjacencyMap` where we simply wrap an existing implementation of adjacency maps into a `newtype`:

https://github.com/snowleopard/alga/blob/master/src/Algebra/Graph/NonEmpty/AdjacencyMap/Internal.hs#L112-L121

This approach allows to reuse a lot of code via type-safe coercions, for example:

https://github.com/snowleopard/alga/blob/master/src/Algebra/Graph/NonEmpty/AdjacencyMap.hs#L200

We also modify the consistency invariant to make sure the graph is non-empty:

https://github.com/snowleopard/alga/blob/master/src/Algebra/Graph/NonEmpty/AdjacencyMap/Internal.hs#L165-L166

In your case you'll use `isAcyclic` instead of `not . isEmpty`.

Also, I don't think we'll be using low-level interfaces like `addEdge` for creating such graphs. I'd prefer to start by adding something like `scc` which is guaranteed to return an acyclic graph. 

Finally, let's first start with a parametric version `AdjacencyMap`, not with an `Int` variant which is just a performance optimisation.

snowleopard:
On a second thought, we might use your type-safe `Edge` idea when implementing a graph construction primitive like `edges`, so let's keep it!

adithyaov:
> I think it may be more productive to follow the approach from NonEmpty.AdjacencyMap where we simply wrap an existing implementation of adjacency maps into a `newtype`.

Agreed! I'll change the implementation.

> Also, I don't think we'll be using low-level interfaces like `addEdge`.

Agreed, I wanted to take feedback on the idea before proceeding further. I believe I can make safe `connect` and `overlay` as well.

> Finally, let's first start with a parametric version AdjacencyMap, not with an Int variant which is just a performance optimization.

I've used the Int variant because it is easier to type check (using the Nat data kind and type level comparision) and convert it from type level to Integer datatype (KnownNat).

Is it possible to promote **any** data type and somehow derieve a type level comparision function for it? 

Parametrization could be possible if we define the AdjacencyMap like such: 
```
newtype AdjacencyIntMap [some phantom types for acyclic checks] a = ...
```
The above design looks more like vertex labeled Graphs.

Is it OK if i keep opening PRs for feedback? Or should I opt for Draft PRs?

snowleopard:
> I've used the Int variant because it is easier to type check (using the Nat data kind and type level comparison) and convert it from type level to Integer datatype (KnownNat).

Note that it's perfectly fine if some functions can only produce `AdjacencyMap Int` as a result. For example, `edges` defined on top of your `Edge` idea could return `AdjacencyMap Int`, but `scc` could return `AdjacencyMap a`.

> Is it possible to promote any data type and somehow derive a type level comparison function for it?

Good question! I don't know the answer. It would be interesting to see if there is a way to promote `Ord` to the type level.

Regarding phantom types: I suggest we look at this option a bit later. It's definitely possible to use phantom types to get rid of a lot of very similar graph types (non-empty, labelled, acyclic, etc.) but it's a separate issue which I suggest to consider at the end of the project.

adithyaov:
> Note that it's perfectly fine if some functions can only produce AdjacencyMap Int as a result. For example, edges defined on top of your Edge idea could return AdjacencyMap Int, but scc could return AdjacencyMap a.

Could you explain how this can be possible with the implementation in this PR? I don't this is possible from the kind of implementation that I provide in the pull request. I'll probably have to change it.

> Regarding phantom types: I suggest we look at this option a bit later.

In this implementation, I am using phantom types though. I believe any kind of type-level programming uses phantom types in some way?

@anfelor has actually suggested that an Acyclic graph could look like the following, ([https://gist.github.com/adithyaov/f87b5b496fd88ef91cfe438dfaf3a955](https://gist.github.com/adithyaov/f87b5b496fd88ef91cfe438dfaf3a955))
```
data Acyclic n a = Acyclic (AdjacencyMap a) (Vector n a)
append :: a -> [Ordinal n] -> Acyclic n a -> Acyclic (S n) a
```
One could probably take this further, I'll experiment with this and get back to you!

snowleopard:
> In this implementation, I am using phantom types though.

Ah, indeed, I didn't notice the phantom parameter at first. 

In this case let's start with a version that only uses `scc` and is just a newtype wrapper around the existing `AdjacencyMap a` data type.

> Could you explain how this can be possible with the implementation in this PR? 

Couldn't you just use `<=?` as a constraint, e.g. something like this?

```haskell
addEdge ::
     forall f t. (KnownNat f, KnownNat t, (f <=? t) ~ True)
  => Edge f t
  -> AdjacencyMap a
  -> AdjacencyMap a
```

adithyaov:
> Couldn't you just use <=? as a constraint, e.g. something like this?

Ah, Interesting!
```
addEdge ::
     forall f t. (KnownNat f, KnownNat t, (f <=? t) ~ True)
  => Edge f t
  -> AdjacencyMap a
  -> AdjacencyMap a
```
In the above, how can one encode a generic type `a` to a natural number? `Edge f t` does not have a meaning if `f` and `t` dont correspond to some vertices. One can restrict the generic type `a` to have an instance of `Integral` (to use `toInteger`). This just does not feel correct.

I think it would be better to use sized Vector and Ordinals for defining an Acyclic Adjacency map.

snowleopard:
Oh, sorry, I meant:

```haskell
addEdge ::
     forall f t. (KnownNat f, KnownNat t, (f <=? t) ~ True)
  => Edge f t
  -> AdjacencyMap Int
  -> AdjacencyMap Int
```

snowleopard:
@adithyaov Are you planning on reviving this PR? If not, please close it. 

In principle, I'd love to see something like `edge` or `edges` for creating acyclic graphs using ordered edges.

adithyaov:
> In principle, I'd love to see something like edge or edges for creating acyclic graphs using ordered edges.

@snowleopard I'll experiment a bit and get back to you.

PR #200, opened 2019-05-20
Acyclic monad

adithyaov:
The first implementation of Acyclic Monad. The implementation is in its initial stage and yet to be tested.
The only way to build an Acyclic Graph is by using the `dag` function.  The most important function is
`unsafeToAcyclic`. It is unsafe as it goes into an infinite loop is the input graph has cycles.
The algorithm recursively defines vertices and builds the graph on top of them. I should probably rename the module.

Constructive criticism is highly appreciated :-)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@ cabal.sandbox.config
 cabal.project.local
 ghcid.txt
 .vscode/
+*.swp
+*.swo
--- a/algebraic-graphs.cabal
+++ b/algebraic-graphs.cabal
@@ -106,7 +106,8 @@ library
                         containers  >= 0.5.5.1 && < 0.8,
                         deepseq     >= 1.3.0.1 && < 1.5,
                         mtl         >= 2.1     && < 2.3,
-                        transformers
+                        indexed-extras,
+                        indexed
     if !impl(ghc >= 8.0)
         build-depends:  semigroups  >= 0.18.2  && < 0.18.4
     default-language:   Haskell2010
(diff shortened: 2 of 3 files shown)


adithyaov:
I can still see an unsafe methods to create the graph. For eg.
```
dags :: StateT (DAG a) (State (DAG a) ()
dags = do
  v1 <- lift singleton "v1"
  put $ edgeTo [v1] "v"
  return ()

d1 = do
  x <- execStateT dags Nil
  ...
```
Here `x` will result in an incomplete DAG as we are using vertices defined in one monadic computation in another.
Please ignore this for now, this can be eliminated by wrapping State into a new type and hiding it from the world.

snowleopard:
This PR doesn't seem to match the requirements from the wiki: 

https://github.com/snowleopard/alga/wiki/Acyclic-graphs#requirements

Do you disagree with the requirements? If yes, let's discuss them first, not the implementation.

adithyaov:
> This PR doesn't seem to match the requirements from the wiki:

I think one can replace DAG' with AdjacencyMap. It probably won't make a difference as primitives of construction are the same (singleton, edgeTo).
I wanted feedback on the construction method. This has no type checking, the safety is because of the method of construction itself.

snowleopard:
> I think one can replace DAG' with AdjacencyMap.

Maybe, and guessing what exactly you mean makes this PR difficult to review. It's much more productive to review a PR which follows an agreed interface.

adithyaov:
This experiment was not so successful.

PR #203, opened 2019-05-22
Partial order based Acyclic Graph

adithyaov:
The following implementation is based on partial orders. I've added a few required functions such as `scc`, `topSort` and an auxillary function `incrementOrd`.

The most important function is `unsafeToAcyclic` with results in an Acyclic graph after a partial order is applied based on the topological ordering. If the input graph has cycles, it results in an infinite loop.


adithyaov:
There is currently 1 problem with this implementation. If vertices are defined as `(Int, a)`. `(2, "x")` and `(3, "x")` are considered different vertices. I'll solve this as soon as possible, please ignore this for now.

adithyaov:
With the partial order `(Int, a)`, `(7, "x")` and `(8, "x")` are treated as different vertices. One can force them to be treated as same vertices by defining a custom `Ord` and modifying overlay and connect to retain only one of to the two. Please ignore this problem for now.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,112 @@
+module Algebra.Graph.Acyclic.AdjacencyMap where
+
+import qualified Algebra.Graph.AdjacencyMap as AM
+import qualified Algebra.Graph.AdjacencyMap.Algorithm as AMA
+import qualified Algebra.Graph.AdjacencyMap.Internal as AMI
```
snowleopard: For simplicity, I think we should use `AM` for qualifying all these modules, because they all refer to the same `AdjacencyMap` data type and have no name clashes.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,112 @@
...
+     (Ord a, Ord b)
+  => AdjacencyMap a
+  -> AdjacencyMap b
+  -> AdjacencyMap (Either a b)
+connectD (AAM a) (AAM b) = AAM (AM.connect (AM.gmap Left a) (AM.gmap Right b))
+
+transitiveClosure :: Ord a => AdjacencyMap a -> AdjacencyMap a
+transitiveClosure (AAM x) = AAM (AM.transitiveClosure x)
+
+overlay :: (Ord a) => AdjacencyMap a -> AdjacencyMap a -> AdjacencyMap a
```
snowleopard: This can produce cyclic graphs, so shouldn't be here. There should be no functions that allow the user to create `Acyclic.AdjacencyMap` that is not `consistent`.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,112 @@
...
+overlay (AAM x) (AAM y) = AAM (AM.overlay x y)
+
+connect :: (Ord a) => AdjacencyMap a -> AdjacencyMap a -> AdjacencyMap a
```
snowleopard: What does this function do?

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,112 @@
...
+  Map.unionsWith Set.union $
+  AM.adjacencyMap x :
+  AM.adjacencyMap y :
+  [ Map.fromSet
+      (\k -> Set.filter (< k) . Map.keysSet $ AM.adjacencyMap y)
+      (Map.keysSet $ AM.adjacencyMap x)
+  ]
+
+-- reverse and Map.keys results in an ascending list
+topSort :: (Ord a) => AdjacencyMap a -> [a]
```
snowleopard: This is incorrect: you need to call the `AM.topSort` function, because the topological sort does not necessarily have anything to do with the order on keys. Try to test this function using `isTopSort` and you'll see that the test will fail. 

As a general comment, we should test all functions.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,112 @@
...
+topSort (AAM (AMI.AM m)) = reverse $ Map.keys m
+
+-- scc
+scc :: (Ord a) => AM.AdjacencyMap a -> AdjacencyMapP (NonEmpty.AdjacencyMap a)
```
snowleopard: This doesn't match the requirements: the result should be `AdjacencyMap (NonEmpty.AdjacencyMap a)`.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,112 @@
...
+import qualified Data.Set as Set
+
+newtype AdjacencyMap a = AAM
+  { aam :: AM.AdjacencyMap a
+  } deriving (Show)
+
+consistent :: Ord a => AdjacencyMap a -> Bool
+consistent (AAM m) = AMI.consistent m && AMA.isAcyclic m
+
+type AdjacencyMapP a = AdjacencyMap (Int, a)
```
snowleopard: Why do we need this?

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,112 @@
...
+
+-- reverse and Map.keys results in an ascending list
+topSort :: (Ord a) => AdjacencyMap a -> [a]
+topSort (AAM (AMI.AM m)) = reverse $ Map.keys m
+
+-- scc
+scc :: (Ord a) => AM.AdjacencyMap a -> AdjacencyMapP (NonEmpty.AdjacencyMap a)
+scc am = unsafeToAcyclic $ AMA.scc am
+
+unsafeToAcyclic :: (Ord a) => AM.AdjacencyMap a -> AdjacencyMapP a
```
snowleopard: Is this function supposed to rely on partial order? It doesn't match the required API.

As for an unsafe way to create an acyclic graph from a cyclic graph, we can just use `AAM`, i.e. the constructor itself.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
snowleopard: `scc` is a very simple function, something like this:

```haskell
scc m = AAM (AM.scc m)
```

snowleopard:
@adithyaov I left a few comments in the review. While this is a good start, the partial order story does not match the requirements.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
adithyaov: Oh! I'm sorry, will change that ASAP!

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
adithyaov: There is an invariant that I have in mind:
**Invariant:** The edges in the graph only depend on the order on its vertices. There can only be edges from a higher ordered vertex to a lower ordered vertex. All the functions in this PR are made keeping that in mind.
Keeping the invariant in mind it would result in a correct topological ordering.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
adithyaov: It wouldn't be that simple as some edges might break the invariant if the default order is taken into account.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
adithyaov: To impose a different order on the vertices this is required.
The graph is always constructed keeping the order of vertices in mind.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
adithyaov: This function is not for the partial order API. That is yet to be created.

All the vertices in `AM.AdjacencyMap` need not follow the proper ordering as required by the Invariant.
This function is used to induce a new ordering to the vertices so that the invariant is not broken.

adithyaov:
@snowleopard All the functions I have created are with keeping a specific invariant in mind.
Keeping the invariant true for all acyclic graphs can have many advantages.
For example, adding edges is easier and no matter what kind of edges one adds, it always results in an acyclic graph. One need not check for cycles all the time.
Adding edges if there is only a wrapper becomes more complicated as you will need to check for cycles and what not.
We can have more helper functions on `Acyclic.AdjacencyMapP` to give users more freedom. For Eg, we can add ways to increase/decrease the order of vertices without breaking the invariant. etc.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
adithyaov: Ok, will do that!

adithyaov:
> the partial order story does not match the requirements.

That has not been created yet, will do it today!

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
adithyaov: Ok, will change that!

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
snowleopard: Which requirement are you trying to achieve using this extra complexity?

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
snowleopard: Here is an example: `1 * 2 + 2 * 1`. Both `1*2` and `2*1` are `consistent`, but the result isn't.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
snowleopard: But we use `consistent` to capture all our invariants, and at the moment it doesn't include this extra complexity. As I mentioned, if you use the function `AM.topSort` you do not need to rely on any extra invariants: `consistent` provides enough guarantees. Why do you need this extra complexity, which requirement does it help to achieve?

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
snowleopard: Can you give an example? The very definition of `scc` guarantees absence of cycles.

Have a look at the documentation:

https://hackage.haskell.org/package/algebraic-graphs-0.4/docs/Algebra-Graph-AdjacencyMap-Algorithm.html#v:scc

We even have the property `isAcyclic . scc == const True` which is tested in the testsuite. So the result is guaranteed to be acyclic, we do not need any additional requirements.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
snowleopard: I'm afraid I'm lost. I don't understand why any extra invariants are needed on the ordering of vertices.

snowleopard:
> Keeping the invariant true for all acyclic graphs can have many advantages.

But we can meet all the requirements from the wiki without this extra invariant. 

> For example, adding edges is easier and no matter what kind of edges one adds, it always results in an acyclic graph. One need not check for cycles all the time.

But there is no requirement to support addition of edges. And I suggest not to jump into this territory now: please focus on the simple, achievable goal for now, which is captured in the wiki, and in the GSoC project description. It's simple and useful!

adithyaov:
I have a few questions about the partial orders written down in the wiki. There still seems a way to create cyclic graphs with an improper definition of `partialOrder`.

snowleopard:
> There still seems a way to create cyclic graphs with an improper definition of `partialOrder`.

Yes, this is indeed the case, but it's probably OK. We will state the requirements: the user of this function must provide a valid **strict partial order**, which satisfies the axioms of **irreflexivity** and **transitivity**.

I've updated the wiki and answered your questions: https://github.com/snowleopard/alga/wiki/Acyclic-graphs#construct-from-an-algebraic-graph-and-a-partial-order-high-priority.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
adithyaov: For simplicity in adding the edges in the future. But maybe I'm thinking too far ahead. I think it's better to build on simple building blocks as you suggested :-)

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
adithyaov: Keeping the invariant in mind an edge could be formed from a lower ordered vertex to a higher one. In that sense, it breaks the invariant. Does not form any cycles though. I've changed it now.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
adithyaov: Adding edges in the near future. But I've changed it now!

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
adithyaov: I apologize, my bad! Maybe I was thinking far too ahead making things complicated.

adithyaov:
I apologize for all the messy versions of the code but I think having the invariant suggested will serve as a very powerful assumption (in a good way) to build the API on. But as you suggest I think it's best if we proceed step by step :-)

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,58 @@
+module Algebra.Graph.Acyclic.AdjacencyMap where
```
snowleopard: We need to add an export list -- at least the constructor will have to be hidden.

algebraic-graphs.cabal
```diff
@@ -104,7 +105,8 @@ library
                         base-compat >= 0.9.1   && < 0.11,
                         containers  >= 0.5.5.1 && < 0.8,
                         deepseq     >= 1.3.0.1 && < 1.5,
-                        mtl         >= 2.1     && < 2.3
+                        mtl         >= 2.1     && < 2.3,
+                        transformers
```
snowleopard: It doesn't look like we actually need this package.

snowleopard:
@adithyaov Great, thank you! We are getting close to a version that can be merged :) Here are the next steps that need to be done:

* Please bring back `topSort`. 
* Add documentation and examples -- see how this is done in other modules.
* Add tests. This will probably force you to introduce an `Internal` module.
* Please also address [this new requirement for `box`](https://github.com/snowleopard/alga/wiki/Acyclic-graphs#construct-via-cartesian-product-medium-priority) in this PR. It should be straightforward.

> Also, `Algebra.Graph` does not export `toAdjacencyMap`. Is this intentional?

I think we don't export it simply to keep the API of this module as small as possible. Note that the function `toAdjacencyMap` can be obtained from the generic graph conversion module `Algebra.Graph.ToGraph`.

algebraic-graphs.cabal
adithyaov: Oh no, I've needed it previously but not anymore. I'll remove it in the next commit.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
adithyaov: Yup! I wanted to do all of that in the end. Will do it in the next commit.

adithyaov:
I've added the documentation. Please let me know your thoughts. I'll add the tests in a day or two.
Is it ok if I resolve previous conversations?

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,181 @@
...
+  } deriving (Show)
+
+-- | Check if the internal graph representation is consistent,
+-- i.e. that all edges refer to existing vertices and the graph
+-- is acyclic. It should be impossible to create an inconsistent 
+-- adjacency map.
+--
+-- @
+-- consistent 'empty'         == True
+-- consistent (1 + 2)         == True
```
snowleopard: Alas, we cannot implement a `Num` instance for `Acyclic.AdjacencyMap`, so we cannot use operators in the examples. All examples must be valid Haskell, and we will need to tested them as is (or with minimum possible modifications).

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,181 @@
...
+-- consistent (1 * 2 + 2 * 3) == True
+-- consistent (1 * 2 + 2 * 1) == False (Violates cyclic property)
```
snowleopard: We should not be able to create an acyclic graph via the public API, so there should be no `False` examples.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,181 @@
...
+-- @
+vertices :: (Ord a) => [a] -> AdjacencyMap a
+vertices xs = AAM $ AM.vertices xs
+
+-- | Perform a disjoint overlay of two different acyclic graphs.
+-- Complexity: /O((n + m) * log(n))/ time and /O(n + m)/ memory.
+--
+-- @
+-- overlayD empty empty             == 'empty'
+-- overlayD (1 * 2 + 1 * 3) (1 * 2) == (Left 1) * (Left 2)
```
snowleopard: If you add a function like `edgeList` to the API, you will be able to rewrite such examples as

```haskell
edgeList (overlayD (1 * 2 + 1 * 3) (1 * 2)) = [(Left 1, Left 2), ...]
```

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,181 @@
...
+scc = AAM . AM.scc
+
+-- | Compute the /topological sort/ of a graph.
+-- 
+-- @
+-- topSort (1)             == [1]
+-- topSort (1 * 2 + 3 * 1) == [3,1,2]
+-- @
+topSort :: (Ord a) => AdjacencyMap a -> [a]
+topSort (AAM am) = Typed.topSort (Typed.fromAdjacencyMap am)
```
snowleopard: Why are you going via the `Typed` module? I think you should just do `fromJust . AM.topSort`.

snowleopard: Sorry, my above comment is wrong. Let's keep your current implementation, because going via `AM.topSort` leads to unnecessary `isTopSort` check.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,181 @@
...
+-- Complexity: /O(s1 * s2)/ time, memory and size, where /s1/ and /s2/
+-- are the sizes of the given graphs.
+--
+-- @
+-- box (1 * 2) (3 * 4) == (1, 3) * (1, 4) + (2, 3) * (2, 4) 
+--                      + (1, 3) * (2, 3) + (1, 4) * (2, 4)
+-- box (1 + 2) (3 + 4) == (1, 3) + (1, 4) + (2, 3) + (2, 4)
+-- @
+box :: (Ord a, Ord b) => AdjacencyMap a -> AdjacencyMap b -> AdjacencyMap (a, b)
+box (AAM x) (AAM y) = AAM $ AM.overlay (fO $ xy y x) (fO $ yx x y)
```
snowleopard: Why do you reimplement this function instead of simply calling to `AM.box`?

snowleopard: Ah! Answering my own question: because there is no `AM.box` yet! I'll try to add it today, so that you can simplify this function.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,181 @@
...
+  where
+    xy q = AM.gmap (\x -> AM.gmap (x, ) q)
+    yx p = AM.gmap (\y -> AM.gmap (, y) p)
+    fO =
+      Map.foldrWithKey (\k -> AM.overlay . Set.foldr AM.overlay k) AM.empty .
+      AM.adjacencyMap
+
+type PartialOrder a = a -> a -> Bool
```
snowleopard: This needs quite a bit of documentation explaining the axioms of strict partial orders, etc.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,181 @@
...
+
+-- | Constructs an acyclic graph from any graph based on
+-- a strict partial order to produce an acyclic graph.
```
snowleopard: Please expand this comment: explain that we only keep the edge `(x, y)` where `x` precedes `y` in the partial order.

snowleopard:
> I've added the documentation. Please let me know your thoughts. I'll add the tests in a day or two.

@adithyaov Thank you! I've added some more feedback. In particular, I think we can't use `+` and `*` in the examples, unless we introduce some clever operators `+` and `*` based on your ideas of using type-level comparison, so that we can rule out expressions like `2*1`. Do you think you can do this?

> Is it ok if I resolve previous conversations?

I think I've already resolved them all.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
snowleopard: See also [my comment below](https://github.com/snowleopard/alga/pull/203#issuecomment-495928972): perhaps, we could implement operators `+` and `*` that work on `AdjacencyMap Int`.

adithyaov:
> I think we can't use + and * in the examples, unless we introduce some clever operators + and * based on your ideas of using type-level comparison, so that we can rule out expressions like 2*1. Do you think you can do this?

Yes, I will give it a try. 

At the beginning of the document I have mentioned the following,
```
In the following code examples, the arthemetic is used to represent
the graph. There is no instance of 'Num' on the AdjacencyMap so it
used to only to act as a representation of the graph.
```

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
adithyaov: Ah! Will do that.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
adithyaov: I've mentioned at the beginning of the file that the arithmetic is only for displaying the graph structure since there no instance of 'Num'. But maybe as you said it's possible to create a safe 'Num' instance.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
adithyaov: Yes, that makes sense!

snowleopard:
@adithyaov Actually, I just realised that we can create a simple `Num` instance by filtering out all edges that violate `(<) :: PartialOrder Int`, so `2 * 1` would be equal to `1 + 2`.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
snowleopard: Also, it's probably sufficient to use a simpler `overlayD 1 (2 * 3)` as an example to avoid a long list of edges.

adithyaov:
That is what I had in mind. But I'll also explore the type-safe alternative and get back to you.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
adithyaov: I'm still exploring a type-safe version of  Num but I created a version where we filter out invalid edges.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
adithyaov: Please let me know if I need to add/change anything else.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
adithyaov: Please let me know if I need to add/change anything else.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,260 @@
...
+  where
+    aMF = AM.AM . Map.mapWithKey (\k -> Set.filter (k `o`)) . AM.adjacencyMap
+
+-- | __Note:__ this does not satisfy the usual ring laws; see 
+-- 'AdjacencyMap' for more details.
+instance (Ord a, Num a) => Num (AdjacencyMap a) where
+  fromInteger = AAM . fromInteger
+  (AAM (AM.AM x)) + (AAM (AM.AM y)) =
+    AAM . AM.AM $
+    Map.unionWithKey (\k p q -> Set.filter (> k) $ Set.union p q) x y
```
adithyaov: Should I mention that by default the partial order is (<) in the instance 'Num'?
ie. an edge (x, y) is possible if and only if 'x < y == True'

adithyaov:
I can't believe I wrote `writed` in the commit there. My bad!

adithyaov:
I am still yet to make tests for `scc`.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
snowleopard: @adithyaov I've added `AM.box`, see #206.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,373 @@
...
+  , connectD
+  , vertices
+  -- * Additional functions on acyclic graphs
+  , box
+  , topSort
+  , transitiveClosure
+  -- * Acyclic graph construction methods
+  , scc
+  , fromGraph
+  ) where
```
adithyaov: I will change the export structure in the end (before merge).

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
snowleopard: I think you can drop Asymmetry, because it follows from the other two properties.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,360 @@
...
+
+-- | The sorted list of edges of a graph.
+-- Complexity: /O(n + m)/ time and /O(m)/ memory.
+--
+-- @
+-- edgeList 'empty'      == []
+-- edgeList ('vertex' x) == []
+-- edgeList (1 * 2)      == [(1,2)]
+-- edgeList (2 * 1)      == []
+-- @
```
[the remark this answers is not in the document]
adithyaov: I'm sorry I did not understand. I've aligned all the examples wrt `==`.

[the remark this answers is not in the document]
snowleopard: Here I'm referring to the alignment in the generated documentation, not in the source.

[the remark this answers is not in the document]
snowleopard: Nice alignment in the documentation is more important for me than the alignment in sources :)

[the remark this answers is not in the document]
adithyaov: Ah! I understand what you meant. Just ran `haddock`.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,402 @@
...
+--
+-- @
+-- transpose 'empty'       == 'empty'
+-- transpose ('vertex' x)  == 'vertex' x
+-- transpose (1 * 2)     == (2 * 1)
+-- transpose . transpose == id
+-- 'edgeList' . transpose  == 'Data.List.sort' . 'map' 'Data.Tuple.swap' . 'edgeList'
+-- @
+transpose :: Ord a => AdjacencyMap a -> AdjacencyMap a
+transpose = AAM . AM.transpose . aam
```
snowleopard: Just a hint: you can use type-safe zero-cost coercions to avoid wrapping and unwrapping:

```haskell
transpose = coerce AM.transpose
```

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,405 @@
...
+--   > a 'R' b and b 'R' c => a 'R' c (Transitive)
+-- Some examples of a Strict Partial Order are 
+-- __\<__ and __\>__.
+type PartialOrder a = a -> a -> Bool
+
+-- A helper function.
+-- This is not efficient but can be used till an efficient version
+-- of induce is included in the main library.
+induceE :: (Ord a) => ((a, a) -> Bool) -> AdjacencyMap a -> AdjacencyMap a
+induceE p a = rmER p eL a
```
adithyaov: Hopefully, there will be an `induceE` in the `Graph.AdjacencyMap` module for me to use.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,418 @@
+module Algebra.Graph.Acyclic.AdjacencyMap (
+  -- * Data types and type aliases
+  PartialOrder,
+
+  -- * Data structure
```
snowleopard: This section seems unnecessary. Also, I think `AdjacencyMap` should go first in documentation, as it is more important.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,418 @@
...
+  AdjacencyMap,
+
+  -- * Internal Checks
+  consistent,
```
snowleopard: Please be consistent with other modules, where this function comes last in the "Miscellaneous" section.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,418 @@
...
+import Algebra.Graph (Graph, foldg)
+import qualified Algebra.Graph.AdjacencyMap as AM
+import qualified Algebra.Graph.AdjacencyMap.Algorithm as AM
+import qualified Algebra.Graph.NonEmpty.AdjacencyMap as NonEmpty
+import qualified Data.Graph.Typed as Typed
+import Data.Set (Set)
+import Data.Coerce (coerce)
+
+{-| The 'AdjacencyMap' data type represents an acyclic graph by
+wrapping around 'Algebra.Graph.AdjacencyMap.AdjacencyMap'. The
```
snowleopard: Implementation should not be a part of the documentation. We might change the implementation later.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,418 @@
...
+idea is that the user of the library should not be given a way
+to produce a cyclic graph.
+
+We define a 'Num' instance as a convenient notation for working
+with graphs:
+
+  > edgeList   0           == []
```
snowleopard: Please swap the order in these examples: start with `vertexList`.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,418 @@
...
+  > edgeList   (1 + 2)     == []
+  > vertexList (1 + 2)     == [1,2]
+  > edgeList   (1 * 2)     == [(1,2)]
+  > vertexList (1 * 2)     == [1,2]
+  > edgeList   (1 + 2 * 3) == [(2,3)]
+  > vertexList (1 + 2 * 3) == [1,2,3]
+  > edgeList   (1 * 2 + 3) == [(1,2)]
+  > vertexList (1 * 2 + 3) == [1,2,3]
+
+The default partial order for the 'Num' instance to restrict
```
snowleopard: There is no need to talk about partial orders here -- not everyone knows what they are, and they are not essential for understanding how `Num` instance works. Please simplify this comment. Just say that we only keep edges that are aligned with the order of integers.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,418 @@
...
+
+-- | Check if the internal graph representation is consistent,
+-- i.e. that all edges refer to existing vertices and the graph
+-- is acyclic. It should be impossible to create an inconsistent 
+-- adjacency map.
+--
+-- @
+-- consistent 'empty'           == True
+-- consistent (1 + 2)         == True
+-- consistent (1 * 2 + 2 * 3) == True
```
snowleopard: These examples do not agree with the testsuite. Just list here everything that you test. 

In general, all examples should be tested.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,418 @@
...
+-- @
+consistent :: Ord a => AdjacencyMap a -> Bool
+consistent (AAM m) = AM.consistent m && AM.isAcyclic m
+
+-- | Construct the /empty acyclic graph/.
+-- Complexity: /O(1)/ time and memory.
+empty :: AdjacencyMap a
```
snowleopard: Please keep as many examples from `Algebra.Graph.AdjacencyMap` as possible in the acyclic API. In this case, you can keep all of them!

http://hackage.haskell.org/package/algebraic-graphs-0.4/docs/Algebra-Graph-AdjacencyMap.html#v:empty

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,418 @@
...
+
+-- | Construct the graph comprising a given list of isolated vertices.
+-- Complexity: /O(L * log(L))/ time and /O(L)/ memory, where /L/ is
+-- the length of the given list.
+--
+-- @
+-- vertices []        == 'empty'
+-- vertices [x]       == 'vertex' x
+-- vertices [1, 2, 3] == 1 + 2 + 3
+-- @
```
snowleopard: Please don't diverge from standard examples used in other modules. This makes it harder to reuse tests, and also introduces unnecessary inconsistency between modules.

http://hackage.haskell.org/package/algebraic-graphs-0.4/docs/Algebra-Graph-AdjacencyMap.html#v:vertices

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,418 @@
...
+-- | Perform a disjoint overlay of two different acyclic graphs.
+-- Complexity: /O((n + m) * log(n))/ time and /O(n + m)/ memory.
+--
+-- @
+-- edgeList (overlayD empty empty)             == []
+-- edgeList (overlayD (1 * 2 + 1 * 3) (1 * 2)) == [(Left 1,Left 2) 
+--                                                ,(Left 1,Left 3)
+--                                                ,(Right 1,Right 2)]
+-- @
+overlayD ::
```
snowleopard: I had a look at some standard libraries, and it seems more common to use prefix "disjoint" rather than suffix "D". So, let's rename these functions to `disjointOverlay` and `disjointConnect`.

snowleopard: Actually, an even better option would be to use "union" and "join" instead, which are standard terms from graph theory:

http://mathworld.wolfram.com/GraphUnion.html
http://mathworld.wolfram.com/GraphJoin.html

What do you think?

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,418 @@
...
+
+-- | Perform a disjoint overlay of two different acyclic graphs.
+-- Complexity: /O((n + m) * log(n))/ time and /O(n + m)/ memory.
+--
+-- @
+-- edgeList (overlayD empty empty)             == []
+-- edgeList (overlayD (1 * 2 + 1 * 3) (1 * 2)) == [(Left 1,Left 2) 
+--                                                ,(Left 1,Left 3)
+--                                                ,(Right 1,Right 2)]
+-- @
```
snowleopard: Can you think of some interesting algebraic properties? They give much better intuition than just ad hoc examples. Look at `overlay` as an example -- plenty of interesting properties, and I'm sure this operation has some interesting properties too.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,418 @@
...
+--                                        ,(Left 1,Right 2)
+--                                        ,(Left 2,Right 1)
+--                                        ,(Left 2,Right 2)]
+-- edgeList (connectD (1 * 2) (1 * 2)) == [(Left 1,Left 2) 
+--                                        ,(Left 1,Right 1)
+--                                        ,(Left 1,Right 2) 
+--                                        ,(Left 2,Right 1) 
+--                                        ,(Left 2,Right 2) 
+--                                        ,(Right 1,Right 2)]
+-- @
```
snowleopard: These examples are quite complex (lots of details). Could you simplify them? Even better -- find some algebraic properties!

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,418 @@
...
+-- scc ('vertex' x) == 'vertex' (NonEmpty.'NonEmpty.vertex' x)
+-- @
+scc :: (Ord a) => AM.AdjacencyMap a -> AdjacencyMap (NonEmpty.AdjacencyMap a)
+scc = coerce AM.scc
+
+-- | Compute the /topological sort/ of a graph.
+-- 
+-- @
+-- topSort (1)             == [1]
+-- topSort (1 * 2 + 3 * 1) == [3,1,2]
```
snowleopard: This seems to be a confusing example, because the `Num` instance is supposed to keep only the edge `(1, 2)`. Why do we get this order and not, say, `[1,2,3]`? By accident? The reader will surely be wondering about that.

test/Algebra/Graph/Test/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,235 @@
...
+import Algebra.Graph ()
+import Data.List.NonEmpty hiding (transpose)
+
+import qualified Algebra.Graph.AdjacencyMap as AM
+import qualified Algebra.Graph.NonEmpty.AdjacencyMap as NonEmpty
+import qualified Data.List as List
+import qualified Data.Set as Set
+
+-- t :: Testsuite
+-- t = testsuite "AcyclicAdjacencyMap." empty
```
snowleopard: Let's not commit commented out code.

test/Algebra/Graph/Test/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,235 @@
...
+-- Testsuite for "Algebra.Graph.Acyclic.AdjacencyMap".
+-------------------------------------------------------------
+
+module Algebra.Graph.Test.Acyclic.AdjacencyMap (
+  testAcyclicAdjacencyMap 
+  ) where
+
+import Algebra.Graph.Acyclic.AdjacencyMap
+import Algebra.Graph.Test
+import Algebra.Graph ()
```
snowleopard: Why do we need this import?

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
adithyaov: Since we are distinguishing between which `AdjacencyMap` the vertex came from, using `disjointUnion` makes more sense. `union` and `join` are similar to `overlay` and `connect`.

test/Algebra/Graph/Test/Acyclic/AdjacencyMap.hs
adithyaov: I think I am (was) using the `Num` instance of `Graph` somewhere. I'll check and remove it if I can.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
adithyaov: Ah! I've used this initially but changed the test suite to use arbitrary graph instances.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
snowleopard: To me `disjointUnion` feels redundant, because graph union is disjoint by definition. I guess we better stick to `disjointOverlay` to avoid confusing terms. It's fairly easy to understand the difference between `overlay` and `disjointOverlay`.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
snowleopard: Just to avoid a misunderstanding: better change these examples to match the current tests. It's very important to test invariants using `Arbitrary` instance.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,449 @@
+module Algebra.Graph.Acyclic.AdjacencyMap (
+  -- * Data structure
+  AdjacencyMap, aam,
```
snowleopard: If we need to export `aam` then we should give it a proper name, e.g. `fromAcyclic`, and add docs/tests. See, for example:

https://github.com/snowleopard/alga/blob/07fa7cdfc5358a39fbf9bff47568d2fedf67f203/src/Algebra/Graph/Relation/Symmetric.hs#L98-L106

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,449 @@
...
+
+  -- * Basic graph construction primitives
+  empty, vertex, disjointOverlay, disjointConnect, vertices,
+
+  -- * Graph properties
+  isEmpty, edgeList, vertexList, edgeSet, vertexSet,
```
snowleopard: Please stick to the same export order as in other modules, for consistency of the documentation: 

https://hackage.haskell.org/package/algebraic-graphs-0.4/docs/Algebra-Graph-AdjacencyMap.html#g:4

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
adithyaov: Exporting it isn't necessary now. I'll remove it. Maybe I'll add it later if required.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,449 @@
...
+should act as additive and multiplicative identities, and 'negate' as
+additive inverse. Nevertheless, overloading 'fromInteger', '+' and '*'
+is very convenient when working with algebraic graphs; we hope that
+in future Haskell's Prelude will provide a more fine-grained class
+hierarchy for algebraic structures, which we would be able to utilise
+without violating any laws.
+-}
+newtype AdjacencyMap a = AAM
+  { aam :: AM.AdjacencyMap a
+  } deriving (Show, Eq, Ord)
```
snowleopard: I don't think we can derive the `Show` instance, because it would use graph construction primitives like `edge`, which are not available in the API here. We'll need to figure out something else, and it looks like a non-trivial task.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,449 @@
...
+
+-- | Check if the internal graph representation is consistent,
+-- i.e. that all edges refer to existing vertices and the graph
+-- is acyclic. It should be impossible to create an inconsistent 
+-- adjacency map.
+--
+-- @
+-- consistent 'empty'                 == True
+-- consistent (vertex x)            == True
+-- consistent (disjointOverlay x y)        == True
```
snowleopard: The alignment is off. Also, only `empty` has quotes -- I think all primitives should be quoted. The general rule is: quote all functions apart from the function that is being defined.

adithyaov: Oh! That's because I've replaced `overlayD` with `disjointOverlay`. It should be off in `disjointConnect` too. My bad, I'll fix it!

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,449 @@
...
+vertices = coerce AM.vertices
+
+-- | Perform a disjoint overlay of two different acyclic graphs.
+-- Complexity: /O((n + m) * log(n))/ time and /O(n + m)/ memory.
+--
+-- @
+-- 'isEmpty' (disjointOverlay x y)             == 'isEmpty' x && 'isEmpty' y
+-- 'hasVertex' (Left z) (disjointOverlay x y)  == 'hasVertex' z x
+-- 'hasVertex' (Right z) (disjointOverlay x y) == 'hasVertex' z y
+-- 'vertexCount' (disjointOverlay x y)         >= 'vertexCount' x
```
snowleopard: This inequality is probably unnecessary, because the equality property below is much stronger.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,449 @@
...
+-- 'vertexCount' (disjointOverlay x y)         == 'vertexCount' x + 'vertexCount' y
+-- 'edgeCount'   (disjointOverlay x y)         >= 'edgeCount' x
```
snowleopard: Same here.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,449 @@
...
+-- 'edgeCount'   (disjointOverlay x y)         == 'edgeCount' x + 'edgeCount' y
+-- 'vertexCount' (disjointOverlay 1 2)         == 2
+-- 'edgeCount'   (disjointOverlay 1 2)         == 0
```
snowleopard: These two special cases are also unnecessary -- they are covered by the equality properties above.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,449 @@
...
+-- 'hasVertex'   (Left z) (disjointConnect x y)  == 'hasVertex' z x
+-- 'hasVertex'   (Right z) (disjointConnect x y) == 'hasVertex' z y
+-- 'vertexCount' (disjointConnect x y)           >= 'vertexCount' x
+-- 'vertexCount' (disjointConnect x y)           == 'vertexCount' x + 'vertexCount' y
+-- 'edgeCount'   (disjointConnect x y)           >= 'edgeCount' x
+-- 'edgeCount'   (disjointConnect x y)           >= 'edgeCount' y
+-- 'edgeCount'   (disjointConnect x y)           >= 'vertexCount' x * 'vertexCount' y
+-- 'edgeCount'   (disjointConnect x y)           == 'vertexCount' x * 'vertexCount' y + 'edgeCount' x + 'edgeCount' y
+-- 'vertexCount' (disjointConnect 1 2)           == 2
+-- 'edgeCount'   (disjointConnect 1 2)           == 1
```
snowleopard: As in `disjointOverlay`, some properties here seem to be redundant. This is not too critical: feel free to leave everything as is -- I'll do a minor revision after merging this PR anyway.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
adithyaov: I cannot write the `API` for `Acyclic.AdjacencyMap` is there isn't a `Show` instance. I am not using the `API` anyway, so should I remove the `API`?

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,466 @@
+module Algebra.Graph.Acyclic.AdjacencyMap (
```
nobrakal: From memory, all files of `src/algebra` have a header with some documentation (used after by Haddock), maybe it's worth adding one for this file (for example: https://github.com/snowleopard/alga/blob/master/src/Algebra/Graph.hs ).

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
nobrakal: Maybe changing the comment to say that there is two axioms, and the antisymmetry follows from them can be a good idea. Right now, it feels strange to says "there is three axioms", and having after just two of them explained.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -0,0 +1,466 @@
...
+-- induce (/= x)          == 'removeVertex' x
+-- induce p . induce q    == induce (\x -> p x && q x)
+-- @
+induce :: (a -> Bool) -> AdjacencyMap a -> AdjacencyMap a
+induce = coerce AM.induce
+
+-- Helper function, not to be exported.
+-- Induce a subgraph from AM.AdjacencyList removing edges not
+-- following the given predicate.
+induceEAM ::
```
nobrakal: Maybe this function can be optimized to use directly the adjacency map structure, with something like 
```Haskell
Map.mapWithKey (\x -> Set.filter (\y -> p (x,y))) m
```

adithyaov: Unfortunately, I cannot work directly on the `Map` since `AM` from `AdjacencyMap` isn't exported. Until a version of `induce` that works on edges is not provided this can be used as an alternative.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
adithyaov: Done!

snowleopard:
Many thanks @adithyaov, merged!

Please add yourself to `AUTHORS.md` (if you'd like to be listed there).

PR #204, opened 2019-05-22
Derive Generic and NFData for Algebra.Graph and Algebra.Graph.Labelled

wismill:
Similar work as #178


snowleopard:
Many thanks @wismill!

PR #207, opened 2019-05-31
Add bipartite adjacency maps

vasalf:


src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,24 @@
...
+import qualified Data.Map.Strict as Map
+import qualified Data.Set        as Set
+
+data AdjacencyMap a b = BAM {
+    leftToRight :: Map.Map a (Set.Set b),
+    rightToLeft :: Map.Map b (Set.Set a)
+} deriving Eq
+
+consistent :: (Ord a, Ord b) => AdjacencyMap a b -> Bool
+consistent (BAM lr rl) = oneWayConstistent lr rl && oneWayConstistent rl lr
```
snowleopard: The current invariant seems too complicated. Couldn't we just do this?

```haskell
consistent (BAM lr rl) = internalEdgeList lr == sort (map swap $ internalList rl)
```

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,24 @@
...
+
+internalEdgeList :: Map.Map a (Set.Set b) -> [(a, b)]
+internalEdgeList lr = [ (u, v) | (u, vs) <- Map.toAscList lr, v <- Set.toAscList vs ]
+
+hasEdge :: (Ord a, Ord b) => Map.Map a (Set.Set b) -> a -> b -> Bool
```
snowleopard: It would be best to make this function similar to the API of other modules:

```haskell
hasEdge :: (Ord a, Ord b) => a -> b -> AdjacencyMap a b -> Bool
```

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
vasalf: @snowleopard, good point, it looks much easier to understand this way!

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
vasalf: @snowleopard It was first designed to be used in `consistent`. Now that we don't need it to check consistency, I will change the signature.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,165 @@
...
+-- consistent ('fromGraph' ('Algebra.Graph.vertex' x)) == True
+-- consistent ('fromGraph' ('Algebra.Graph.edge' x y)) == True
+-- consistent ('fromGraph' ('Algebra.Graph.edges' xs)) == True
+-- consistent ('fromGraph' g)          == True
+-- consistent ('fromAdjacencyMap' am)  == True
+-- @
+consistent :: (Ord a, Ord b) => AdjacencyMap a b -> Bool
+consistent (BAM lr rl) = internalEdgeList lr == sort (map swap $ internalEdgeList rl)
+
+-- | The list of edges of a bipartite adjacency map
```
snowleopard: The convention in the library is to not give Haddock comments to functions that are not exported -- it makes it easier to immediately see whether a function is exported or not, so please change `-- |` to just `--`.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,165 @@
...
+import Data.Tuple  (swap)
+
+import qualified Algebra.Graph              as G
+import qualified Algebra.Graph.AdjacencyMap as AM
+
+import qualified Data.Map.Strict as Map
+import qualified Data.Set        as Set
+
+-- | The 'Bipartite.AdjacencyMap' data type represents an __undirected__
+-- bipartite graph by two maps of vertices into its adjacency sets. The two
```
snowleopard: into its -> into their

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,165 @@
...
+--  __Note:__ even if the identifiers and their types for two vertices of
+-- different parts are equal, these vertices are considered to be different.
+-- See the examples for more details.
+data AdjacencyMap a b = BAM {
+    -- | The /adjacency map/ of the left part of the graph: each vertex is
+    -- associated with a set of its neighbours. Complexity: /O(1)/ time and
+    -- memory
+    --
+    -- @
+    -- leftToRight ('fromGraph' 'Algebra.Graph.empty')                                          == Map.'Map.empty'
```
snowleopard: I think it may be better to follow the naming convention in the rest of the library, where such functions are typically named `adjacencyMap`. In this case we could call them `leftAdjacencyMap` and `rightAdjacencyMap`.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,165 @@
...
+    -- rightToLeft ('fromGraph' ('Algebra.Graph.edge' (Left 1) (Right 1)))                      == Map.'Map.singleton' 1 (Set.'Set.singleton' 1)
+    -- rightToLeft ('fromGraph' ('Algebra.Graph.edges' [(Left 1, Right 1), (Left 1, Right 2)])) == Map.'Map.fromAscList' [(1, Set.'Set.singleton' 1), (2, Set.'Set.singleton' 1)]
+    -- @
+    rightToLeft :: Map.Map b (Set.Set a)
+} deriving (Eq, Show)
+
+-- | Check that the internal graph representation is consistent, i.e. that all
+-- edges that are present in the 'leftToRight' map are present in the
+-- 'rightToLeft' map.
+-- Complexity: /O(m log m)/
```
snowleopard: Let's drop complexity here. This function is not supposed to be performance-critical, so I'd say we don't want to promise any performance guarantees.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,165 @@
...
+--
+-- @
+-- consistent ('fromGraph' 'Algebra.Graph.empty')      == True
+-- consistent ('fromGraph' ('Algebra.Graph.vertex' x)) == True
+-- consistent ('fromGraph' ('Algebra.Graph.edge' x y)) == True
+-- consistent ('fromGraph' ('Algebra.Graph.edges' xs)) == True
+-- consistent ('fromGraph' g)          == True
```
snowleopard: This property covers all of the properties above, so I suggest to delete them.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
snowleopard: Same for `addReverseEdges`.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,165 @@
...
+-- edges to make the graph undirected and removing all edges inside one part.
+-- Complexity: /O(m log n)/
+--
+-- @
+-- 'leftToRight' (fromAdjacencyMap ('Algebra.Graph.AdjacencyMap.empty'))                                        == Map.'Map.empty'
+-- 'leftToRight' (fromAdjacencyMap ('Algebra.Graph.AdjacencyMap.edge' (Left 1) (Right 2)))                      == Map.'Map.singleton' 1 (Set.'Set.singleton' 2)
+-- 'leftToRight' (fromAdjacencyMap ('Algebra.Graph.AdjacencyMap.edge' (Left 1) (Left 2)))                       == Map.'Map.empty'
+-- 'rightToLeft' (fromAdjacencyMap ('Algebra.Graph.AdjacencyMap.edges' [(Left 1, Right 1), (Left 1, Right 2)])) == Map.'Map.fromAscList' [(1, Set.'Set.singleton' 1), (2, Set.'Set.singleton' 1)]
+-- @
+fromAdjacencyMap :: (Ord a, Ord b) => AM.AdjacencyMap (Either a b) -> AdjacencyMap a b
```
snowleopard: After checking other modules, I think this should be named `toBipartite` for consistency.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,165 @@
...
+-- | Constructs an 'Algrebra.Graph.AdjacencyMap' from a bipartite
+-- 'AdjacencyMap'.
+-- Complexity: /O(m log n)/
+--
+-- @
+-- toAdjacencyMap (fromGraph 'Algebra.Graph.empty')                     == 'Algebra.Graph.AdjacencyMap.empty'
+-- toAdjacencyMap (fromGraph ('Algebra.Graph.vertex' $ Left 1))         == 'Algebra.Graph.AdjacencyMap.vertex' 1
+-- toAdjacencyMap (fromGraph ('Algebra.Graph.edge' (Left 1) (Right 2))) == 'Algebra.Graph.AdjacencyMap.edges' [(Left 1, Right 2), (Right 2, Left 1)]
+-- @
+toAdjacencyMap :: (Ord a, Ord b) => AdjacencyMap a b -> AM.AdjacencyMap (Either a b)
```
snowleopard: After checking other modules, I think this should be named `fromBipartite` for consistency.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,165 @@
...
+-- type arguments for this datatype stand for the identifiers of the vertices
+-- of each part.
+--
+--  __Note:__ even if the identifiers and their types for two vertices of
+-- different parts are equal, these vertices are considered to be different.
+-- See the examples for more details.
+data AdjacencyMap a b = BAM {
+    -- | The /adjacency map/ of the left part of the graph: each vertex is
+    -- associated with a set of its neighbours. Complexity: /O(1)/ time and
+    -- memory
```
snowleopard: Missing dot (.) at the end of a sentence. Same issue in a couple of other places.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,165 @@
...
+hasEdge u v (BAM m _) = ((Set.member v) <$> (u `Map.lookup` m)) == Just True
+
+-- | Adds all edges needed to make the graph undirected.
+-- Complexity: /O(m log n)/
+--
+-- /Note: this function is for internal use only./
+addReverseEdges :: (Ord a, Ord b) => AM.AdjacencyMap (Either a b) -> AM.AdjacencyMap (Either a b)
+addReverseEdges m = AM.overlay m $ AM.edges [ (v, u) | (u, v) <- AM.edgeList m ]
+
+-- | Constructs a bipartite 'Bipartite.AdjacencyMap' from
```
snowleopard: I think the Haddock link is broken and should actually be `'AdjacencyMap'`? Please check generated documentation.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,165 @@
...
+-- 'Algebra.Graph.AdjacencyMap' with given part identifiers, adding all needed
```
snowleopard: Here please use double quotes to link to the module instead (otherwise it may be confusing to see two equally named links).

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,165 @@
...
+-- different parts are equal, these vertices are considered to be different.
+-- See the examples for more details.
+data AdjacencyMap a b = BAM {
+    -- | The /adjacency map/ of the left part of the graph: each vertex is
+    -- associated with a set of its neighbours. Complexity: /O(1)/ time and
+    -- memory
+    --
+    -- @
+    -- leftToRight ('fromGraph' 'Algebra.Graph.empty')                                          == Map.'Map.empty'
+    -- leftToRight ('fromGraph' ('Algebra.Graph.vertex' (Left 1)))                              == Map.'Map.singleton' 1 Set.'Set.empty'
```
snowleopard: The examples based on `fromGraph` conversions are quite difficult to read. Could you add basic graph construction primitives to this PR? I suggest adding at least `empty`, `leftVertex`, `rightVertex`, `edge` and `edges`. Then all examples in this module would be much nicer.

snowleopard:
@vasalf Many thanks for the PR! I've left a bunch of comments, please have a look.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
vasalf: I replaced them with the brand new primitives.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
vasalf: Yes, it was broken. Fixed now.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
vasalf: There are all of them now.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,431 @@
...
+-- Complexity: /O(n)/ time.
+--
+-- @
+-- edgeCount 'empty'      == 0
+-- edgeCount ('vertex' x) == 0
+-- edgeCount ('edge' x y) == 1
+-- edgeCount . 'edges'    == 'length' . 'Data.List.nub'
+-- @
+edgeCount :: AdjacencyMap a b -> Int
+edgeCount = Map.foldr (+) 0 . fmap Set.size . leftAdjacencyMap
```
nobrakal: I am not sure that deforestation is implemented for `Map`... Maybe `Map.foldr ((+) . Set.size) 0` will be more efficient

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,431 @@
...
+-- License    : MIT (see the file LICENSE)
+-- Maintainer : andrey.mokhov@gmail.com
+-- Stability  : experimental
+--
+-- __Alga__ is a library for algebraic construction and manipulation of graphs
+-- in Haskell. See <https://github.com/snowleopard/alga-paper this paper> for
+-- the motivation behind the library, the underlying theory, and
+-- implementation details.
+--
+-- This module defines the 'Bipartite.AdjacencyMap' for bipartite graphs data
```
snowleopard: This Haddock link is most likely broken.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,431 @@
...
+    -- | The /adjacency map/ of the left part of the graph: each vertex is
+    -- associated with a set of its neighbours. Complexity: /O(1)/ time and
+    -- memory.
+    --
+    -- @
+    -- leftAdjacencyMap 'empty'                    == Map.'Map.empty'
+    -- leftAdjacencyMap ('leftVertex' 1)           == Map.'Map.singleton' 1 Set.'Set.empty'
+    -- leftAdjacencyMap ('rightVertex' 1)          == Map.'Map.empty'
+    -- leftAdjacencyMap ('edge' 1 2)               == Map.'Map.singleton' 1 (Set.'Set.singleton' 2)
+    -- leftAdjacencyMap ('edge' 1 1)               == Map.'Map.singleton' 1 (Set.'Set.singleton' 1)
```
snowleopard: I suggest to first give the example `edge 1 1`, and then follow it with `edge 1 'a'` to demonstrate that the types of vertices in the two parts can be different.

snowleopard: Same in `rightAdjacencyMap`.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,431 @@
...
+-- consistent ('edges' xs)       == True
+-- consistent ('fromGraph' g)    == True
+-- consistent ('toBipartite' am) == True
+-- @
+consistent :: (Ord a, Ord b) => AdjacencyMap a b -> Bool
+consistent (BAM lr rl) = internalEdgeList lr == sort (map swap $ internalEdgeList rl)
+
+-- The list of edges of a bipartite adjacency map.
+--
+-- /Note: this function is for internal use only./
```
snowleopard: This comment is redundant: all non-exported functions are internal (and one can immediately see that this function is not exported, because it doesn't have a Haddock comment, by convention).

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,431 @@
...
+internalEdgeList lr = [ (u, v) | (u, vs) <- Map.toAscList lr, v <- Set.toAscList vs ]
+
+-- | Check if a graph contains a given edge.
+-- Complexity: /O(log(n))/ time.
+--
+-- @
+-- hasEdge x y 'empty'                                == False
+-- hasEdge x y ('edge' x y)                           == True
+-- hasEdge 2 3 ('edge' 1 2)                           == False
+-- hasEdge x y ('overlay' g ('edge' x y))               == True
```
snowleopard: Let's use `x`-`z` in examples (instead of `g`).

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,431 @@
...
+-- hasEdge 1 2 ('fromGraph' ('Algebra.Graph.edge' (Left 1) (Left 2))) == False
+-- @
+hasEdge :: (Ord a, Ord b) => a -> b -> AdjacencyMap a b -> Bool
+hasEdge u v (BAM m _) = ((Set.member v) <$> (u `Map.lookup` m)) == Just True
```
snowleopard: Redundant parentheses in `(Set.member v)`.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,431 @@
...
+-- @
+-- hasRightVertex x 'empty'           == False
+-- hasRightVertex x ('rightVertex' x) == True
+-- hasRightVertex x ('leftVertex' x)  == False
+-- hasRightVertex 1 ('rightVertex' 2) == False
+-- @
+hasRightVertex :: Ord b => b -> AdjacencyMap a b -> Bool
+hasRightVertex y (BAM _ rl) = y `Map.member` rl
+
+-- | Check if a graph contains a given vertex in the given part.
```
snowleopard: Can be shortened to just "Check if a graph contains a given vertex".

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,431 @@
...
+-- @
+isEmpty :: AdjacencyMap a b -> Bool
+isEmpty (BAM lr rl) = Map.null lr && Map.null rl
+
+-- | The number of vertices in the left part in a graph.
+-- Complexity: /O(1)/ time.
+--
+-- @
+-- leftVertexCount 'empty'         == 0
+-- leftVertexCount 'leftVertex' x  == 1
```
snowleopard: I don't think this compiles -- there are missing parentheses. Same below and in `rightVertexCount` examples.

snowleopard: And in `vertexCount`.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,431 @@
...
+rightVertexCount = Map.size . rightAdjacencyMap
+
+-- | The number of vertices in a graph.
+-- Complexity: /O(1)/ time.
+--
+-- @
+-- vertexCount 'empty'         == 0
+-- vertexCount 'leftVertex' x  == 1
+-- vertexCount 'rightVertex' x == 1
+-- vertexCount g             == 'leftVertexCount' g + 'rightVertexCount' g
```
snowleopard: `g` => `x`

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,431 @@
...
+-- edgeCount 'empty'      == 0
+-- edgeCount ('vertex' x) == 0
+-- edgeCount ('edge' x y) == 1
+-- edgeCount . 'edges'    == 'length' . 'Data.List.nub'
+-- @
+edgeCount :: AdjacencyMap a b -> Int
+edgeCount = Map.foldr (+) 0 . fmap Set.size . leftAdjacencyMap
+
+-- Adds all edges needed to make the graph undirected.
+-- Complexity: /O(m log n)/.
```
snowleopard: Let's use `log(n)` consistently.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,431 @@
...
+--
+-- /Note: this function is for internal use only./
+addReverseEdges :: (Ord a, Ord b) => AM.AdjacencyMap (Either a b) -> AM.AdjacencyMap (Either a b)
```
snowleopard: I think this function is called `symmetricClosure` in other modules.

snowleopard: Perhaps, you could even use `symmetricClosure` directly?

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,431 @@
...
+    -- @
+    -- rightAdjacencyMap 'empty'                    == Map.'Map.empty'
+    -- rightAdjacencyMap ('leftVertex' 1)           == Map.'Map.empty'
+    -- rightAdjacencyMap ('rightVertex' 1)          == Map.'Map.singleton' 2 Set.'Set.empty'
+    -- rightAdjacencyMap ('edge' 1 2)               == Map.'Map.singleton' 2 (Set.'Set.singleton' 1)
+    -- rightAdjacencyMap ('edge' 1 1)               == Map.'Map.singleton' 1 (Set.'Set.singleton' 1)
+    -- rightAdjacencyMap ('edges' [(1, 1), (1, 2)]) == Map.'Map.fromAscList' [(1, Set.'Set.singleton' 1), (2, Set.'Set.singleton' 1)]
+    -- @
+    rightAdjacencyMap :: Map.Map b (Set.Set a)
+} deriving (Eq, Show, Generic)
```
snowleopard: You shouldn't use derived `Show` instance here. Please have a look how `Show` is implemented for `AdjacencyMap` and try to follow it here.

snowleopard: Also, using derived `Eq` is probably suboptimal: we can implement `Eq` by comparing only one of the two maps.

snowleopard: Oh wait, my second comment is probably wrong :)

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,431 @@
...
+-- @
+-- 'leftAdjacencyMap' (fromGraph ('Algebra.Graph.empty'))                                         == Map.'Map.empty'
+-- 'leftAdjacencyMap' (fromGraph ('Algebra.Graph.edge' (Left 1) (Right 2)))                       == Map.'Map.singleton' 1 (Set.'Set.singleton' 2)
+-- 'leftAdjacencyMap' (fromGraph ('Algebra.Graph.edge' (Left 1) (Left 2)))                        == Map.'Map.empty'
+-- 'rightAdjacencyMap' (fromGraph ('Algebra.Graph.edges' [(Left 1, Right 1), (Left 1, Right 2)])) == Map.'Map.fromAscList' [(1, Set.'Set.singleton' 1), (2, Set.'Set.singleton' 1)]
+-- @
+fromGraph :: (Ord a, Ord b) => G.Graph (Either a b) -> AdjacencyMap a b
+fromGraph = toBipartite . (G.foldg AM.empty AM.vertex AM.overlay AM.connect)
+
+-- | Constructs the /empty graph/.
```
snowleopard: Please reorder all function definitions in this module according to the export list. This helps navigating it.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,431 @@
...
+       , AM.vertices $ Left  <$> Map.keys lr
+       , AM.vertices $ Right <$> Map.keys rl ]
+
+-- | Constructs a bipartite 'AdjacencyMap' from a 'Algebra.Graph.Graph' with
+-- given part identifiers, adding all needed edges to make the graph undirected
+-- and removing all edges inside one part.
+-- Complexity: /O(m log n)/.
+--
+-- @
+-- 'leftAdjacencyMap' (fromGraph ('Algebra.Graph.empty'))                                         == Map.'Map.empty'
```
snowleopard: I think these examples can be simplified as follows:

```haskell
fromGraph (Algebra.Graph.'Algebra.Graph.empty') == 'empty'
fromGraph (Algebra.Graph.'Algebra.Graph.edge' (Left 1) (Right 2))) == 'edge' 1 2
...
```

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,431 @@
...
+addReverseEdges m = AM.overlay m $ AM.edges [ (v, u) | (u, v) <- AM.edgeList m ]
+
+-- | Constructs a bipartite 'AdjacencyMap' from "Algebra.Graph.AdjacencyMap"
+-- with given part identifiers, adding all needed edges to make the graph
+-- undirected and removing all edges inside one part.
+-- Complexity: /O(m log n)/.
+--
+-- @
+-- 'toBipartite ('Algebra.Graph.AdjacencyMap.empty')                                                           == 'empty'
+-- 'leftAdjacencyMap' (toBipartite ('Algebra.Graph.AdjacencyMap.edge' (Left 1) (Right 2)))                       == Map.'Map.singleton' 1 (Set.'Set.singleton' 2)
```
snowleopard: See this comment for how to simplify such examples: https://github.com/snowleopard/alga/pull/207/files#r295996443

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,431 @@
...
+empty = BAM Map.empty Map.empty
+
+-- | Constructs the bipartite graph comprising /a single isolated vertex/ in
+-- the left part.
+-- Complexity: /O(1)/ time and memory.
+--
+-- @
+-- 'leftAdjacencyMap' (leftVertex x)  == Map.'Map.singleton' x Set.'Set.empty'
+-- 'rightAdjacencyMap' (leftVertex x) == Map.'Map.empty'
+-- 'hasEdge' x y (leftVertex x)       == False
```
snowleopard: `leftVertex x` => `leftVertex z` for greater generality.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
snowleopard: Also, let's use "Construct" instead of "Constructs" throughout the module.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,431 @@
...
+vertex (Right y) = rightVertex y
+
+-- | Constructs the bipartite graph comprising /a single edge/.
+-- Complexity: /O(1)/ time and memory.
+--
+-- @
+-- 'leftAdjacencyMap' (edge x y)  == Map.'Map.singleton' x (Set.'Set.singleton' y)
+-- 'rightAdjacencyMap' (edge x y) == Map.'Map.singleton' y (Set.'Set.singleton' x)
+-- 'hasEdge' x y (edge x y)       == True
+-- 'hasEdge' y x (edge x y)       == (x == y)
```
snowleopard: I find this example a bit difficult to understand. Perhaps, it's worth splitting it into two concrete ones, say `x=y=1`, and `x=1` and `y=2`.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,431 @@
...
+-- 'hasVertex' z (overlay x y) == 'hasVertex' z x || 'hasVertex' z y
+-- 'vertexCount' (overlay x y) >= 'vertexCount' x
+-- 'vertexCount' (overlay x y) <= 'vertexCount' x + 'vertexCount' y
+-- 'edgeCount'   (overlay x y) >= 'edgeCount' x
+-- 'edgeCount'   (overlay x y) <= 'edgeCount' x   + 'edgeCount' y
+-- @
+overlay :: (Ord a, Ord b) => AdjacencyMap a b -> AdjacencyMap a b -> AdjacencyMap a b
+overlay (BAM lr1 rl1) (BAM lr2 rl2) = BAM (Map.unionWith Set.union lr1 lr2) (Map.unionWith Set.union rl1 rl2)
+
+-- | /Connect/ two bipartite graphs, not adding the inappropriate edges. This
```
snowleopard: Make it clearer: "not adding the inappropriate edges" => "not adding edges between vertices in the same part".

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,431 @@
...
+-- is an associative operation with the identity 'empty', which distributes
+-- over 'overlay' and obeys the decomposition axion.
```
snowleopard: "axion" => "axiom"

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,431 @@
...
+-- 'vertexCount' (overlay x y) >= 'vertexCount' x
+-- 'vertexCount' (overlay x y) <= 'vertexCount' x + 'vertexCount' y
+-- 'edgeCount'   (overlay x y) >= 'edgeCount' x
+-- 'edgeCount'   (overlay x y) <= 'edgeCount' x   + 'edgeCount' y
+-- @
+overlay :: (Ord a, Ord b) => AdjacencyMap a b -> AdjacencyMap a b -> AdjacencyMap a b
+overlay (BAM lr1 rl1) (BAM lr2 rl2) = BAM (Map.unionWith Set.union lr1 lr2) (Map.unionWith Set.union rl1 rl2)
+
+-- | /Connect/ two bipartite graphs, not adding the inappropriate edges. This
+-- is an associative operation with the identity 'empty', which distributes
```
snowleopard: I guess it's also commutative.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,431 @@
...
+-- over 'overlay' and obeys the decomposition axion.
+-- Complexity: /O((n + m) * log(n))/ time and /O(n + m)/ memory. Note that the
+-- number of edges in the resulting graph is quadratic with respect to the
+-- number of vertices in the arguments: /m = O(m1 + m2 + l1 * r2 + l2 * r1)/.
+--
+-- @
+-- 'isEmpty'     (connect x y) == 'isEmpty'   x   && 'isEmpty'   y
```
snowleopard: Add a couple of concrete examples, as this function is quite tricky.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,431 @@
...
+-- 'vertexCount' (connect x y) <= 'vertexCount' x + 'vertexCount' y
+-- 'edgeCount'   (connect x y) >= 'edgeCount' x
+-- 'edgeCount'   (connect x y) >= 'edgeCount' y
+-- 'edgeCount'   (connect x y) >= 'leftVertexCount' x * 'rightVertexCount' y
+-- 'edgeCount'   (connect x y) <= 'leftVertexCount' x * 'rightVertexCount' y + 'rightVertexCount' x * 'leftVertexCount' y + 'edgeCount' x + 'edgeCount' y
+-- @
+connect :: (Ord a, Ord b) => AdjacencyMap a b -> AdjacencyMap a b -> AdjacencyMap a b
+connect (BAM lr1 rl1) (BAM lr2 rl2) = BAM (Map.unionsWith Set.union [ lr1, lr2,
+                                                                      Map.fromSet (const $ Map.keysSet rl2) (Map.keysSet lr1),
+                                                                      Map.fromSet (const $ Map.keysSet rl1) (Map.keysSet lr2)]) $
```
snowleopard: Please try not to exceed 80 characters per line, if possible. Long lines are difficult to review.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,431 @@
...
+-- length of two lists.
+--
+-- @
+-- vertices [] []                               == 'empty'
+-- vertices [x] []                              == 'leftVertex' x
+-- vertices [] [x]                              == 'rightVertex' x
+-- 'hasLeftVertex' x . ('flip' ('const' [])) vertices == 'elem' x
+-- 'hasRightVertex' x . 'const' [] vertices         == 'elem' x
+-- @
+vertices :: (Ord a, Ord b) => [a] -> [b] -> AdjacencyMap a b
```
snowleopard: I think it may be better to follow the approach taken with `vertex`: create `leftVertices`, `rightVertices` and let `vertices` take a list of `Either`s as input. This will make the API closer to non-bipartite graphs.

test/Algebra/Graph/Test/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,433 @@
...
+module Algebra.Graph.Test.Bipartite.AdjacencyMap (
+    -- * Testsuite
+    testBipartiteAdjacencyMap
+    ) where
+
+import Algebra.Graph.Bipartite.AdjacencyMap
+import Algebra.Graph.Test
+
+import qualified Algebra.Graph                       as G
+import qualified Algebra.Graph.AdjacencyMap          as AM
```
snowleopard: Whitespace.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
vasalf: Well, they claim in the docs that each of those `Map.*` functions work in linear time, so deforestation **is** implemented for strict maps. But still `fmap` has some overhead over `(.)`. In the following example `sumPlusOne'` works 1.5 times faster.

```Haskell
module Main where

import qualified Data.Map.Strict as Map

inputMap :: Map.Map Int Int
inputMap = Map.fromAscList [ (i, i) | i <- [1..1000000] ]

sumPlusOne :: Map.Map Int Int -> Int
sumPlusOne = Map.foldr (+) 0 . fmap ((+) 1)

sumPlusOne' :: Map.Map Int Int -> Int
sumPlusOne' = Map.foldr ((+) . ((+) 1)) 0

main :: IO ()
-- main = putStrLn $ show $ sumPlusOne inputMap
main = putStrLn $ show $ sumPlusOne' inputMap
```

I will change this to `Map.foldr ((+) . Set.size) 0`.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
vasalf: Whoops, forgot about it.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
vasalf: We can compare only `leftAdjacencyMap`s and `keysSet`s of the `rigthAdjacencyMap`s, this would be a little bit faster.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
vasalf: OK, but maybe we would better rename the current version and keep it? Something like `fromPartLists`?
It seems ridiculus to do `overlay (leftVertices ls) (rightVertices rs)` or even `vertices ((Left <$> ls) ++ (Right <$> rs))` in case one needs it. And I needed it in some examples.

snowleopard: OK, I agree, let's keep this function as is. Don't worry about adding others for now.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
snowleopard: Thanks for benchmarking, interesting result!

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
vasalf: Yup, I just didn't notice it when I tried to search for something like this.

test/Main.hs
```diff
@@ -26,16 +27,17 @@ main :: IO ()
...
+    go "Export"                testExport
+    go "Graph"                 testGraph
+    go "Internal"              testInternal
+    go "LabelledAdjacencyMap"  testLabelledAdjacencyMap
+    go "LabelledGraph"         testLabelledGraph
+    go "NonEmptyAdjacencyMap"  testNonEmptyAdjacencyMap
+    go "NonEmptyGraph"         testNonEmptyGraph
+    go "Relation"              testRelation
+    go "SymmetricRelation"     testSymmetricRelation
+    go "BipartiteAdjacencyMap" testBipartiteAdjacencyMap
```
adithyaov: Maybe put this after line 32 following the alphabetical order?

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,433 @@
...
+-- License    : MIT (see the file LICENSE)
+-- Maintainer : andrey.mokhov@gmail.com
+-- Stability  : experimental
+--
+-- __Alga__ is a library for algebraic construction and manipulation of graphs
+-- in Haskell. See <https://github.com/snowleopard/alga-paper this paper> for
+-- the motivation behind the library, the underlying theory, and
+-- implementation details.
+--
+-- This module defines the 'AdjacencyMap' for bipartite graphs data type and
```
adithyaov: I think you meant `This module defines the 'AdjacencyMap' data type for bipartite graphs and`

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,666 @@
...
+-- | Swap parts of a given graph.
+-- Complexity: /O(1)/ time and memory.
+--
+-- @
+-- swap 'empty'        == 'empty'
+-- swap . 'leftVertex' == rightVertex
+-- swap . 'vertices'   == flip 'vertices'
+-- swap ('edge' 1 "a") == 'edge' "a" 1
+-- swap . 'edges'      == 'edges' . map Data.Tuple.'Data.Tuple.swap'
+-- @
```
snowleopard: One of the nicest properties is missing: `swap . swap == id` :)

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,666 @@
...
+-- edges that are present in the 'leftAdjacencyMap' are present in the
+-- 'rightAdjacencyMap' map.
+--
+-- @
+-- consistent 'empty'            == True
+-- consistent ('vertex' x)       == True
+-- consistent ('edge' x y)       == True
+-- consistent ('edges' xs)       == True
+-- consistent ('fromGraph' g)    == True
+-- consistent ('toBipartite' am) == True
```
snowleopard: Check consistency of `swap`.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,668 @@
...
+
+    * Absorption and saturation of 'connect':
+
+        > x * y + x + y == x * y
+        >     x * x * x == x * x
+
+When specifying the time and memory complexity of graph algorithms, /n/ and /m/
+will denote the number of vertices and edges in the graph, respectively. In
+addition, /l/ and /r/ will denote the number of vertices in the left and in the
+right part of graph, respectively.
```
snowleopard: Just a note: I wonder if by adding `swap` into the algebra we could get some new interesting axioms and theorems. Something to think about!

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,668 @@
...
+    -- @
+    leftAdjacencyMap :: Map.Map a (Set.Set b),
+
+    -- | The inverse map for 'leftAdjacencyMap'. Complexity: /O(1)/ time and memory.
+    --
+    -- @
+    -- rightAdjacencyMap 'empty'                    == Map.'Map.empty'
+    -- rightAdjacencyMap ('leftVertex' 1)           == Map.'Map.empty'
+    -- rightAdjacencyMap ('rightVertex' 1)          == Map.'Map.singleton' 2 Set.'Set.empty'
+    -- rightAdjacencyMap ('edge' 1 1)               == Map.'Map.singleton' 1 (Set.'Set.singleton' "a")
```
snowleopard: The above two lines seem wrong. Are they in the testsuite?

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,668 @@
...
+instance (Ord a, Ord b, Num b) => Num (AdjacencyMap a b) where
+    fromInteger = rightVertex . fromInteger
+    (+)         = overlay
+    (*)         = connect
+    signum      = const empty
+    abs         = id
+    negate      = id
+
+instance (Ord a, Ord b) => Eq (AdjacencyMap a b) where
+    (BAM lr1 rl1) == (BAM lr2 rl2) = (lr1 == lr2) && (Map.keysSet rl1 == Map.keysSet rl2)
```
snowleopard: This is a tricky definition. It would be nice to test that it actually produces the same result as the trivial implementation based on comparing both maps.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,668 @@
...
+-- | Construct the graph comprising two given lists of isolated vertices for
+-- each part.
+-- Complexity: /O(L * log(L))/ time and /O(L)/ memory, where /L/ is the total
+-- length of two lists.
+--
+-- @
+-- vertices [] []                               == 'empty'
+-- vertices [x] []                              == 'leftVertex' x
+-- vertices [] [x]                              == 'rightVertex' x
+-- 'hasLeftVertex' x . ('flip' ('const' [])) vertices == 'elem' x
```
snowleopard: This and the next examples seem too complex, I suggest simplifying them by dropping point-free notation:

```haskell
hasLeftVertex  x (vertices ys zs) == elem x ys
hasRightVertex x (vertices ys zs) == elem x zs
```

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,668 @@
...
+-- Complexity: /O(m log(n))/.
+--
+-- @
+-- toBipartite (Algebra.Graph.AdjacencyMap.'Algebra.Graph.AdjacencyMap.empty')                     == 'empty'
+-- toBipartite (Algebra.Graph.AdjacencyMap.'Algebra.Graph.AdjacencyMap.edge' (Left 1) (Right 1))   == 'edge' 1 1
+-- toBipartite (Algebra.Graph.AdjacencyMap.'Algebra.Graph.AdjacencyMap.edge' (Left 1) (Left 1))    == 'empty'
+-- toBipartite (Algebra.Graph.AdjacencyMap.'Algebra.Graph.AdjacencyMap.edge' (Left 1) (Right "a")) == 'edge' 1 "a"
+-- @
+toBipartite :: (Ord a, Ord b) => AM.AdjacencyMap (Either a b) -> AdjacencyMap a b
+toBipartite m = BAM (Map.fromAscList [ (u, setRights vs) | (Left  u, vs) <- Map.toAscList (AM.adjacencyMap $ AM.symmetricClosure m)])
```
snowleopard: Please shorten the lines to simplify reviewing.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,668 @@
...
+-- 'AdjacencyMap'.
+-- Complexity: /O(m log(n))/.
+--
+-- @
+-- fromBipartite 'empty'          == 'Algebra.Graph.AdjacencyMap.empty'
+-- fromBipartite ('leftVertex' 1) == 'Algebra.Graph.AdjacencyMap.vertex' (Left 1)
+-- fromBipartite ('edge' 1 2)     == 'Algebra.Graph.AdjacencyMap.edges' [(Left 1, Right 2), (Right 2, Left 1)]
+-- @
+fromBipartite :: (Ord a, Ord b) => AdjacencyMap a b -> AM.AdjacencyMap (Either a b)
+fromBipartite (BAM lr rl) = AM.overlays $
```
snowleopard: I think you could use `fromAdjacencySets` for faster implementation:

https://hackage.haskell.org/package/algebraic-graphs-0.4/docs/Algebra-Graph-AdjacencyMap.html#v:fromAdjacencySets

test/Algebra/Graph/Test/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,620 @@
...
+    test "AM.consistent (fromBipartite x)                                                == True" $ \x ->
+        AM.consistent $ fromBipartite (x :: BAII)
+    test "fromBipartite (toBipartite (AM.biclique (map Left [1..x]) (map Right [1..y]))) == <correct result>" $ \x y ->
+        (fromBipartite $ toBipartite $ AM.biclique (map Left [1..(x :: Int)]) (map Right [1..(y :: Int)]))
+         == AM.overlay (AM.biclique (map Left [1..(x :: Int)]) (map Right [1..(y :: Int)]))
+                       (AM.biclique (map Right [1..(y :: Int)]) (map Left [1..(x :: Int)]))
+
+    putStrLn "\n============ Bipartite.AdjacencyMap.hasEdge ============"
+    test "hasEdge x y (fromGraph empty)                                 == False" $ \x y ->
+        not $ hasEdge (x :: Int) (y :: Int) $ fromGraph G.empty
```
snowleopard: It's typically a bit easier to read if these annotations are on the lambda, i.e. `\(x :: Int) (y :: Int) ->...`.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,668 @@
...
+The 'Show' instance is defined using basic graph construction primitives:
+
+@
+show (empty)                       == "empty"
+show 1                             == "rightVertex 1"
+show ('swap' 2)                      == "leftVertex 2"
+show (1 + 2)                       == "vertices [] [1,2]"
+show ('swap' (1 + 2))                == "vertices [1,2] []"
+show (('swap' 1) * 2)                == "edge 1 2"
+show (('swap' 1) * 2 * ('swap' 3))     == "edges [(1,2),(3,2)]"
```
snowleopard: Note that functions bind more tightly than operators, so some of the above parentheses are redundant (also in tests).

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
vasalf: Whoops, this seems to be a copy-paste error. Yes, now they are.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -0,0 +1,667 @@
...
+
+import qualified Algebra.Graph              as G
+import qualified Algebra.Graph.AdjacencyMap as AM
+
+import qualified Data.Map.Strict as Map
+import qualified Data.Set        as Set
+import qualified Data.Tuple
+
+{-| The 'Bipartite.AdjacencyMap' data type represents an __undirected__
+bipartite graph by two maps of vertices into their adjacency sets. The two type
```
adithyaov: I was told by @snowleopard  that it's best to leave out the implementation details from the documentation as the implementation might change later.

snowleopard: Good point!

Note that it's not only about the implementation that might change in future, but also about unnecessary information in the documentation, which takes up the time of the reader.

vasalf: Thanks for the comment.

I added this in order to show some difference between `Algebra.Graph.Bipartite.AdjacencyMap` and `Algebra.Graph.Bipartite` (if added in future). Now I see that there is no such difference in the documentation for `Algebra.Graph.AdjacencyMap`, so I removed it.

snowleopard:
@vasalf There is a test failure:

```
============ Eq (Bipartite.AdjacencyMap a b) ============
Test failure:
    (x == y) == ((leftAdjacencyMap x == rightAdjacencyMap x) && (leftAdjacencyMap y == rightAdjacencyMap y))
*** Failed! Falsified (after 3 tests):
empty
vertices [-2] [-2]
```

The full log is here: https://travis-ci.org/snowleopard/alga/jobs/550349421

Please investigate.

vasalf:
> @vasalf There is a test failure:
> 
> ```
> ============ Eq (Bipartite.AdjacencyMap a b) ============
> Test failure:
>     (x == y) == ((leftAdjacencyMap x == rightAdjacencyMap x) && (leftAdjacencyMap y == rightAdjacencyMap y))
> *** Failed! Falsified (after 3 tests):
> empty
> vertices [-2] [-2]
> ```
> 
> The full log is here: https://travis-ci.org/snowleopard/alga/jobs/550349421
> 
> Please investigate.

@snowleopard Thanks for pointing it out! It turned out that the test was wrong. I fixed it. Hope it will pass this time (it passes on my machine, but we know that means nothing). 

I also fixed a small bug in `Show` instance.

snowleopard:
Thanks, merged!

PR #208, opened 2019-06-01
Drop Internal modules

snowleopard:
It appears we no longer need `Internal` modules for data types like `AdjacencyIntMap`, so we drop them.


adithyaov:
Maybe having Internal modules is a good thing, even if they solve no purpose. We never know when we might need to use the internal constructures of the datatypes.

snowleopard:
@nobrakal I bumped the library version to `0.5` and it looks like this has broken the performance suite again. Sorry about that. 

Could you please fix it? It would be nice if we could automate this somehow.

PR #209, opened 2019-06-04
Add induceJust

### LINKED ISSUE #202: Add induceJust
From http://hackage.haskell.org/package/witherable-0.3

Given that we can't currently have a graph with a separate key type (https://github.com/snowleopard/alga/issues/141), I'm constructing a graph of keys, and then I want to map over the graph looking up the value of the key, to get a graph of values.

However, lookup is naturally `a -> Maybe b`, an so even with a `Functor` instance (https://github.com/snowleopard/alga/issues/201) I'd still have a `Graph (Maybe value)`.

Having `filter`/`mapMaybe`/`catMaybes` seems like an fine solution. We could either add them as simple functions or use the `Filterable` class.

Avasil:
Fixes #202

### DIFF
--- a/src/Algebra/Graph.hs
+++ b/src/Algebra/Graph.hs
@@ -41,7 +41,7 @@ module Algebra.Graph (
 
     -- * Graph transformation
     removeVertex, removeEdge, replaceVertex, mergeVertices, splitVertex,
-    transpose, induce, simplify, sparsify, sparsifyKL,
+    transpose, induce, induceJust, simplify, sparsify, sparsifyKL,
 
     -- * Graph composition
     compose, box,
@@ -55,7 +55,7 @@ import Control.DeepSeq
 import Control.Monad (MonadPlus (..))
 import Control.Monad.State (runState, get, put)
 import Data.Foldable (toList)
-import Data.Maybe (fromMaybe)
+import Data.Maybe (fromMaybe, maybe)
 import Data.Semigroup ((<>))
 import Data.Tree
 import GHC.Generics
@@ -1009,6 +1009,22 @@ induce p = foldg Empty (\x -> if p x then Vertex x else Empty) (k Overlay) (k Co
     k _ Empty y     = y
     k f x     y     = f x y
 {-# INLINE [1] induce #-}
+
+-- | Construct the /induced subgraph/ of a given graph by removing the 
+-- vertices that are Nothing.
+-- Complexity: /O(s)/ time, memory and size.
+-- @
+-- induceJust ('vertex' (Nothing :: Maybe Int))             == 'empty'
+-- induceJust (gmap Just x)                                 == x
+-- induceJust ('connect' (gmap Just x) ('vertex' Nothing))  == x
+-- @
+induceJust :: Graph (Maybe a) -> Graph a
+induceJust = foldg Empty (maybe Empty Vertex) (k Overlay) (k Connect)
+  where
+    k _ x     Empty = x -- Constant folding to get rid of Empty leaves
+    k _ Empty y     = y
+    k f x     y     = f x y
+{-# INLINE [1] induceJust #-}
 
 -- | Simplify a graph expression. Semantically, this is the identity function,
 -- but it simplifies a given expression according to the laws of the algebra.
--- a/src/Algebra/Graph.hs
+++ b/src/Algebra/Graph.hs
@@ -1014,9 +1014,9 @@ induce p = foldg Empty (\x -> if p x then Vertex x else Empty) (k Overlay) (k Co
 -- vertices that are Nothing.
 -- Complexity: /O(s)/ time, memory and size.
 -- @
--- induceJust ('vertex' (Nothing :: Maybe Int))             == 'empty'
--- induceJust (gmap Just x)                                 == x
--- induceJust ('connect' (gmap Just x) ('vertex' Nothing))  == x
+-- induceJust ('vertex' 'Nothing')                            == 'empty'
+-- induceJust (gmap Just x)                                   == x
+-- induceJust ('connect' (gmap Just x) ('vertex' 'Nothing'))  == x
 -- @
 induceJust :: Graph (Maybe a) -> Graph a
 induceJust = foldg Empty (maybe Empty Vertex) (k Overlay) (k Connect)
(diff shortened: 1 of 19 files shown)


src/Algebra/Graph.hs
```diff
@@ -1010,6 +1010,22 @@ induce p = foldg Empty (\x -> if p x then Vertex x else Empty) (k Overlay) (k Co
     k f x     y     = f x y
 {-# INLINE [1] induce #-}
 
+-- | Construct the /induced subgraph/ of a given graph by removing the 
+-- vertices that are Nothing.
```
snowleopard: Let's add single quotes around `Nothing` to make it stand out from the rest of the text.

(Also in other modules.)

src/Algebra/Graph.hs
```diff
@@ -1010,6 +1010,22 @@ induce p = foldg Empty (\x -> if p x then Vertex x else Empty) (k Overlay) (k Co
...
+-- Complexity: /O(s)/ time, memory and size.
+-- @
+-- induceJust ('vertex' (Nothing :: Maybe Int))             == 'empty'
```
snowleopard: I suggest to omit the type signature in this example, to make it a bit easier to read.

src/Algebra/Graph/AdjacencyMap.hs
```diff
@@ -754,6 +755,20 @@ gmap f = AM . Map.map (Set.map f) . Map.mapKeysWith Set.union f . adjacencyMap
...
+-- @
+-- induceJust ('vertex' (Nothing :: Maybe Int))             == 'empty'
+-- induceJust (gmap Just x)                                 == x
+-- induceJust ('connect' (gmap Just x) ('vertex' Nothing))  == x
+-- @
+induceJust :: Ord a => AdjacencyMap (Maybe a) -> AdjacencyMap a
+induceJust = AM . Map.map catMaybesSet . catMaybesMap . adjacencyMap
+    where 
+      catMaybesSet = Set.map Maybe.fromJust . Set.delete Nothing
+      catMaybesMap = Map.mapKeys Maybe.fromJust . Map.delete Nothing
```
snowleopard: Can we avoid using partial functions like `fromJust`? Perhaps, you could use `Map.traverseMaybeWithKey`? Or would it be much slower?

(Also in other modules.)

test/Algebra/Graph/Test/Generic.hs
```diff
@@ -1515,6 +1515,16 @@ testInduce (prefix, API{..}) = do
     test "isSubgraphOf (induce p x) x == True" $ \(apply -> p) x ->
           isSubgraphOf (induce p x) x == True
 
+testInduceJust :: Testsuite g Ord -> IO ()
```
snowleopard: I think we should include this test into `testTransformations` above. Then it will be called from all testsuites.

snowleopard:
@Avasil Looks great, many thanks! I've added a few minor comments.

test/Algebra/Graph/Test/Generic.hs
Avasil: I wanted to do it initially but I realized that I can't implement `induceJust` for `AdjacencyIntMap`. Is there any workaround?

snowleopard: Oh, good point. In this case let's stick to your current approach.

snowleopard:
By the way, we could also add `induceJust1` to `NonEmpty.*` modules, which has a cute type signature:

```haskell
induceJust1 :: Graph (Maybe a) -> Maybe (Graph a)
```

Feel free to ignore this if you don't have enough time.

src/Algebra/Graph/AdjacencyMap.hs
Avasil: I just tried and unfortunately we can't do it because it doesn't provide a way to change `key`: 
```haskell
traverseMaybeWithKey :: Applicative f => (k -> a -> f (Maybe b)) -> Map k a -> f (Map k b)
```
It's too bad because this function is `O(n)` instead of `O(n * log(n))` like `mapKeys`

`Map.fromJust` raises red flags but it should be safe because there can be only one `Nothing` key, right?

snowleopard: I think we could use `Set.mapMonotonic` and `Map.mapKeysMonotonic` to get O(n) complexity.

snowleopard: This will rely on the existing `Ord` instance for `Maybe`, which preserves the order in `Just`s.

Avasil: Awesome, seems like it works. :)

Avasil:
@snowleopard Thanks for the review! I implemented `NonEmpty` versions and other comments
It seems like the build is failing now https://api.travis-ci.org/v3/job/541821370/log.txt

Mostly in `RewriteRules` - is it nondeterministic test, or could it be because of my changes?

snowleopard:
Hmm, I'm not sure what's going on with rewrite rules. @nobrakal Could you please have a look?

nobrakal:
That is really strange, I just ran the test suite, and it compiled normally on my machine (with `stack test`). The failing from Travis seems pretty strange too... I will investigate further.

Speaking of rewrite rules, at first glance, it seems that `induceJust` could benefit from the same kind of rule than for `induce` (namely the `"buildR/induce"` rule). Even if this is a not-so-beautiful rule, it feels strange to have one for `induce` and not for `induceJust`.

Ps: I just noticed that the bench suite was broken (again...), I really need to fix this once and for all !

snowleopard:
I'll try to fix the CI here: https://github.com/snowleopard/alga/pull/210

snowleopard:
@Avasil Would you like to be listed in https://github.com/snowleopard/alga/blob/master/AUTHORS.md? If yes, please add yourself to the list.

algebraic-graphs.cabal
```diff
@@ -147,7 +147,7 @@ test-suite test-alga
     if !impl(ghc >= 8.0)
         build-depends:  semigroups   >= 0.18.2  && < 0.18.4
     if impl(ghc >= 8.0.2)
-        build-depends:  inspection-testing >= 0.4 && < 0.5
+        build-depends:  inspection-testing >= 0.4 && < 0.4.2
```
snowleopard: Has the rebase gone wrong? This shouldn't be here.

Avasil: Probably, I didn't change it (and it's already like this on master)

Avasil:
@snowleopard Thanks, I added myself!

@nobrakal Good idea, I can look into that but I'd rather do it in separate PR

algebraic-graphs.cabal
Avasil: I think I corrected it

snowleopard:
@Avasil Great! All looks good now. I'll merge once CI passes.

PR #215, opened 2019-06-14
Acyclic.Ord module created

adithyaov:


src/Algebra/Graph/Acyclic/Ord.hs
```diff
@@ -0,0 +1,164 @@
...
+-- Stability  : experimental
+--
+-- __Alga__ is a library for algebraic construction and manipulation of graphs
+-- in Haskell. See <https://github.com/snowleopard/alga-paper this paper> for the
+-- motivation behind the library, the underlying theory, and implementation details.
+--
+-- This module defines functions which guarantee Acyclic graph by
+-- establishing an invariant on the edges.
+-- An edge (x, y) exists if and only if x < y. This way no back
+-- edges are formed and hence making an acyclic graph is impossible.
```
nobrakal: I think you meant "hence making a cyclic graph is impossible" :)

adithyaov: Ah! yes, my bad.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -504,3 +504,26 @@ induceEAM p m = es m `AM.overlay` vs m
   where
     es = AM.edges . filter p . AM.edgeList
     vs = AM.vertices . AM.vertexList
+
+-- | Constructs an acyclic graph from any graph based on
+-- a order of vertices to produce an acyclic graph.
```
nobrakal: English is not my mother tongue, but maybe there is a typo here: "a order" -> "an order".

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -504,3 +504,26 @@ induceEAM p m = es m `AM.overlay` vs m
...
+-- /toAcyclicOrd (1 \* 2 + 2 \* 1) == 1 \* 2/ because
+-- /1 \< 2 == True/ and hence the edge is allowed.
+-- /2 \< 1 == False/ and hence the edge is filtered out.
+--
+-- @
+-- toAcyclicOrd (2 * 1)         == 1 + 2
+-- toAcyclicOrd (1 * 2)         == 1 * 2
+-- toAcyclicOrd (1 * 2 + 2 * 1) == 1 * 2
+-- toAcyclicOrd                 == fromGraph (<) . toGraph
+-- @
```
nobrakal: All of this seems very related to the definition of topological ordering, which is, quoting [wikipedia](https://en.wikipedia.org/wiki/Topological_sorting#Relation_to_partial_orders), a "linear extension of a partial order". Maybe saying some words about it is a good idea.

By the way, it seems that enumerating the vertices following the order constraint gives automatically a topological sorting. Having a custom `topsort` for such graphs can be cool, but I don't know if it is simple to implement with the actual module structure.

src/Algebra/Graph/Acyclic/Ord.hs
```diff
@@ -0,0 +1,164 @@
+-----------------------------------------------------------------
+-- |
+-- Module     : Algebra.Graph..Acyclic.Ord
```
nobrakal: Typo here: ".." -> "."

src/Algebra/Graph/Acyclic/Ord.hs
```diff
@@ -0,0 +1,164 @@
...
+-- 'hasVertex' z (overlay x y) == 'hasVertex' z x || 'hasVertex' z y
+-- 'vertexCount' (overlay x y) >= 'vertexCount' x
+-- 'vertexCount' (overlay x y) <= 'vertexCount' x + 'vertexCount' y
+-- 'edgeCount'   (overlay x y) >= 'edgeCount' x
+-- 'edgeCount'   (overlay x y) <= 'edgeCount' x   + 'edgeCount' y
+-- 'vertexCount' (overlay 1 2) == 2
+-- 'edgeCount'   (overlay 1 2) == 0
+-- @
+overlay :: Ord a => AdjacencyMap a -> AdjacencyMap a -> AdjacencyMap a
+overlay x y = toAcyclicOrd $ AM.overlay (fromAcyclic x) (fromAcyclic y)
```
nobrakal: This is only cosmetic, and I don't know what @snowleopard thinks of it, but there is not-so-known high-order function called [`on`](http://hackage.haskell.org/package/base-4.12.0.0/docs/Data-Function.html#v:on) in `base`, of type `(b -> b -> c) -> (a -> b) -> a -> a -> c `, which can maybe be used here to simplify the expression.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
snowleopard: @nobrakal Interesting, isn't `topSort` in this case going to be simply `vertexList`? :)

nobrakal: Yes, I think :D

src/Algebra/Graph/Acyclic/Ord.hs
adithyaov: > not-so-known 

Agreed! I was searching for something like this. Will change it!

test/Main.hs
```diff
@@ -27,6 +28,7 @@ main = do
     selected <- getArgs
     let go current = when (null selected || current `elem` selected)
     go "AcyclicAdjacencyMap"  testAcyclicAdjacencyMap
+    go "AcyclicOrd"           testAcyclicOrd
```
snowleopard: I think this should belong to the `AcyclicAdjacencyMap` testsuite to avoid having too many modules.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -504,3 +504,34 @@ induceEAM p m = es m `AM.overlay` vs m
...
+-- This will guarantee the production of an acyclic graph since
+-- no back edges are possible.
+--
+-- Topological orderings are also closely related to the concept
+-- of a linear extension of a partial order in mathematics.
+-- Please look at <https://en.wikipedia.org/wiki/Topological_sorting#Relation_to_partial_orders Relation to partial orders> for
+-- additional information. 
+-- In this case the partial order is the order of the vertices
+-- itself. And hence, the topological ordering for such graphs
+-- is simply its `vertexList`.
```
snowleopard: Please be careful with Haddock: I believe this markup syntax is not supported. Double-check the generated documentation.

snowleopard:
@adithyaov Thanks for the PR! Generally looks good, but see some unresolved comments above.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
adithyaov: Ah, my bad. I did check for the most part. I'll go through it again, in the meanwhile please let me know if you've found anything else.

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
adithyaov: Please let me know if the following changes shed more light on topSort.

snowleopard:
@adithyaov Thank you, merged!

PR #217, opened 2019-06-20
BFS in *.Algorithm modules

jitwit:
First draft. Took a stab at adding bfs search. I went off the api for the dfs procedures and used sets from containers to keep track of state. The unfoldTreeM_BF uses a sequence under the hood for the queue.

If this looks ok, I take it the next steps would be to add tests and documentation? Otherwise, at least in the case of AdjacencyIntMaps, perhaps it would be worth using Arrays (as Data.Graph does)?

Joseph


snowleopard:
@jitwit Many thanks for the PR! I'm currently travelling and have a few other PRs to review, but I'll try to come back to you soon! 

In the meanwhile, have a look at the discussion in an unfinished PR #185, which was also about BFS: perhaps, you'll find something useful there.

snowleopard:
@jitwit Many thanks for the PR! I'm currently travelling and have a few other PRs to review, but I'll try to come back to you soon! 

In the meanwhile, have a look at the discussion in an unfinished PR #185, which was also about BFS: perhaps, you'll find something useful there.

jitwit:
Thanks for the quick reply and no worries if it takes a while to get around to reviewing the changes!

I also incorporated tests for the bfs procedures. Again, I went off the respective dfs ones, but ensured the behavior is different where it should be. I also included the cases from the discussions in PR #185 , to indicate things are working as expected.

jitwit:
I ran some benchmarks with criterion on Knuth's words graph from the stanford graph base, which has 34027 edges and 5757 vertices.

[https://gist.github.com/jitwit/e19c2bc29126782831531755523d919b](https://gist.github.com/jitwit/e19c2bc29126782831531755523d919b)

I have two implementations of bfs right now, one using a `Set/IntSet` to keep track of visited vertices in the *.Algorithm modules, the other using a `STUArray s Int Bool` in the Data.Graph.Typed module.

The first group of benchmarks is done on the `Graph String`, the second on an isomorphic `Graph Int` copy of the first. 

IntMap/Sets are quite fast, so the current dfs implementation which uses them actually does better than converting to the KL representation then running Data.Graph's dfs (~20ms vs ~30ms). The bfs implementation builds a bfs forest almost as fast as `Data.Graph.dfs` builds a dfs forest called directly on an `Array Vertex [Vertex]` (~7ms vs ~5ms).

For, `AdjacencyMaps` it seems like it's still better to convert to KL representation on DFS.

The implementation using `STUArray s Int Bool` was surprisingly slow. The overhead of converting to KL representation may be high or I may be doing something silly.

I build up all the graphs and run `deepseq` on them before running the benches, which I believe is the right way to do it, but let me know if I'm doing anything wrong!

jitwit:
Compares well against fgl too!

[https://gist.github.com/jitwit/0c2d136cadc7a32f090fb98dda338bab](https://gist.github.com/jitwit/0c2d136cadc7a32f090fb98dda338bab)

jitwit:
Hi @snowleopard. Does the bfs implementation look reasonable? It uses Data.Tree's monadic unfold to build the forest.

I've added tests and documentation with complexity analysis. This implementation seems to perform better than fgl's, according to some criterion benchmarks I ran (see gists above). Thanks!

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -16,12 +18,13 @@
 module Algebra.Graph.AdjacencyIntMap.Algorithm (
     -- * Algorithms
     dfsForest, dfsForestFrom, dfs, reachable, topSort, isAcyclic,
-
+    bfsForest, bfsForestFrom, bfs,
```
snowleopard: A general comment: please put `bfs*` in front of `dfs*` throughout your PR. They are quite similar and `bfs` should probably get ahead simply lexicographically. I'm happy to reorder things too, but then I'd probably end up touching almost every line in this PR!

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -115,6 +118,56 @@ dfs vs = concatMap flatten . dfsForestFrom vs
...
+bfsForest :: AdjacencyIntMap -> Forest Int
+bfsForest g = bfsForestFrom' (vertexList g) g
+
+-- | Like 'bfsForest', but the traversal is seeded by a list of
+-- vertices.  Let /L/ be the number of seed vertices. Complexity:
+-- /O(n+(L+m)*log n)/ time and /O(n+m)/ space.
+--
+-- @
+-- 'forest' (bfsForestFrom [1,2] $ 'edge' 1 2) == 'vertices' [1,2]
+-- 'forest' (bfsForestFrom [2]   $ 'edge' 2 1) == 'vertex' 2
```
snowleopard: I don't think the examples here work, and they seem different from tests. Please bring them in sync.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -115,6 +118,56 @@ dfs vs = concatMap flatten . dfsForestFrom vs
...
+  discovered v = do new <- gets (not . IntSet.member v)
+                    when new $ modify' (IntSet.insert v)
+                    return new
+
+-- | Like 'bfsForestFrom' with the resulting forest flattened to a
+-- list of vertices. Complexity: /O(n+(L+m)*log n)/ time and /O(n+m)/
+-- space.
+-- 
+-- @ bfs [3] ('circuit' [1..5] + 'transpose' ('circuit'
+-- [1..5])) == [3,2,1,4,5]
```
snowleopard: Something is off with the example.

src/Data/Graph/Typed.hs
```diff
@@ -1,3 +1,5 @@
+{-# language LambdaCase, ViewPatterns #-}
```
snowleopard: Changes in this file seem accidental, please undo.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -115,6 +118,56 @@ dfs vs = concatMap flatten . dfsForestFrom vs
...
+    False -> bff vs
+    True -> (:) <$> unfoldTreeM_BF walk v <*> bff vs
+  walk v = (v,) <$> adjacentM v
+  adjacentM v = filterM discovered $ IntSet.toList (postIntSet v g)
+  discovered v = do new <- gets (not . IntSet.member v)
+                    when new $ modify' (IntSet.insert v)
+                    return new
+
+-- | Like 'bfsForestFrom' with the resulting forest flattened to a
+-- list of vertices. Complexity: /O(n+(L+m)*log n)/ time and /O(n+m)/
```
snowleopard: `L` is undefined.

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
```diff
@@ -53,6 +56,7 @@ import qualified Data.Set                            as Set
 --                                          , Node { rootLabel = 3
 --                                                 , subForest = [ Node { rootLabel = 4
 --                                                                      , subForest = [] }]}]
+-- 'forest' (dfsForest ('circuit' [1..5] + 'transpose' ('circuit' [1..5]))) == 'path' [1..5]
```
snowleopard: This example appears to be untested.

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
```diff
@@ -81,6 +85,20 @@ dfsForest g = dfsForestFrom (vertexList g) g
 -- @
 dfsForestFrom :: Ord a => [a] -> AdjacencyMap a -> Forest a
 dfsForestFrom vs = Typed.dfsForestFrom vs . Typed.fromAdjacencyMap
+--  = prune $ evalState (dff vs) Set.empty where
```
snowleopard: Some leftover code.

snowleopard:
@jitwit Many thanks for finishing off the PR and for benchmarking! Apologies once again for being so slow.

I like your implementation, but I left a few minor comments -- please have a look.

If I understand your benchmarks correctly, your BFS implementation is much faster than the current DFS one. In this case, shall we switch `reachable` to using BFS? Or do you think we could speed up DFS as well?

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: Yeah, thankfully the test cases were what they ought to be, and passed! I'm not sure how I got the documentation examples so wrong

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: Yeah, here too... This example is supposed to show the breadth-first-ness and how nodes are queued according to ordering from Ord. From ghci REPL:

```
> bfs [3] (circuit [1,2,3,4,5] + circuit [5,4,3,2,1])
[3,2,4,1,5]
```

as it should be and as tested!

src/Data/Graph/Typed.hs
jitwit: Sorry, from my experiments running bfs on KGL representation over STUArray s Int Bools. The overhead from converting made this slower than working over AdjacencyIntMaps and IntSets, so I removed the code.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: It's mentioned in the documentation for `bfsForestFrom` as the number of seed vertices. I should probably restate it here?

 I also wasn't sure what to call it, so I went with L... perhaps something else would be clearer?

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
jitwit: Ah, I was going to contrast the dfs-order to the similar example on bfs. Should I add a test or delete the line here?

jitwit:
Yeah that seems to be the case. I think it's more accurate to say that new bfs is much faster for `AdjacencyIntMaps`, and somewhat faster for `AdjacencyMap a`.

I ran yet more benches, and also tried reachable with bfs as well. For the example graph (n = 5757, m = 34027), bfs-reachable seems to be ~3x faster than the current reachable on `AdjacencyIntMap`, and around 50% faster for `AdjacencyMap Text`. So it could be the way to go! 

I think that breadth first search is performing better for two reasons. One is that once a node is enqueued, it can be excluded from further consideration, whereas with depth first, queued nodes are not necessarily in the dfs forest. The second reason is avoiding overhead of converting representations.

The first reason made the attempt at re-implementing depth first search on AdjacencyMap is a bit awkward, using Data.Tree's unfoldTreeM. I haven't come up a better solution than first returning a `Forest (Maybe a)`, then pruning  the resulting dfs forest. Basically, nodes are enqueued when "white" or "gray". On exit, if still "gray" a `Just a` is put in the forest, or if "black", a Nothing node is. I might be missing something obvious? The attempt is in the gist after imports.

According to previous benchmarks, that dfs attempt was faster for a good portion of graphs, but not all. eg K_1000 got horrible results.

[https://gist.github.com/jitwit/80447786c00a43d757e8d8de26549b8f](https://gist.github.com/jitwit/80447786c00a43d757e8d8de26549b8f)

In fewer words: using bfs for reachable is probably a good idea. Also, I think it may be possible to improve depth first search, but the current attempt gets mixed results!

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: switched to writing (circuit [1..5] + circuit [5,4,3,2,1]) wherever example appears

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: fixed to `forest (bfsForestFrom [2]   $ edge 1 2) == vertex 2`

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: Should the bfs code come before in the module as well? I changed the export lists for now

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
snowleopard: You can also do `circuit [1..5] + circuit [5,4..1]` which is a bit shorter.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
snowleopard: Yes, please redefine `L`. Ideally, each function should have a self-contained description.

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
snowleopard: It's a useful example, but I'd like all the examples to be tested. Please add the corresponding test!

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
snowleopard: Yes, I prefer to provide implementations in the same order as they appear in the export list. Same for `API` methods.

snowleopard:
@jitwit Sounds good, let's switch to using BFS for `reachable` then! We should be able to see some speed up on the CI performance suite immediately (https://travis-ci.org/snowleopard/alga/jobs/558141381).

I've been also thinking about making `bfs` a bit more useful for a larger proportion of users, and I think that perhaps we could omit the `concat` step, switching to `bfs :: Ord a => [a] -> AdjacencyMap a -> [[a]]`, thus returning BFS levels. It's easy to apply `concat` to the latter if need be, but going in the opposite direction is trickier (e.g. one needs to know the function `levels` from `Data.Tree`). What do you think?

snowleopard:
P.S.: We could have a look at optimising DFS separately.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: I switched to `s`, which I think makes more sense. Also restated it in both places it was missing

jitwit:
Hm! It seems `reachable` is hand rolled bench suite, as well as conversion to adjacencymap. I was perplexed to see 1.0 new vs old in the Bench job. 

 [https://github.com/haskell-perf/graphs/blob/master/bench/Alga/Graph.hs](https://github.com/haskell-perf/graphs/blob/master/bench/Alga/Graph.hs)

jitwit:
@snowleopard  I also definitely agree with returning `[[a]]` from `bfs`. I initially went `[a]` to mirror the API from Data.Graph.

I'll try to think more about optimizing `dfs`. I have a hunch that even if dfs proves hard to match performance-wise, topological sort directly on AdjacencyMaps may possibly work?

jitwit:
Maybe I should add a comment in the documentation about `concat . bfs x` returning the vertices in bfs-order?

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -113,7 +170,7 @@ dfs vs = concatMap flatten . dfsForestFrom vs
 -- 'isSubgraphOf' ('vertices' $ reachable x y) y == True
 -- @
 reachable :: Int -> AdjacencyIntMap -> [Int]
-reachable x = dfs [x]
+reachable x = concat . bfs [x]
```
snowleopard: Please update the comment, the examples, as well as the corresponding tests to match the new implementation.

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
```diff
@@ -116,7 +174,7 @@ dfs vs = concatMap flatten . dfsForestFrom vs
 -- 'isSubgraphOf' ('vertices' $ reachable x y) y == True
 -- @
 reachable :: Ord a => a -> AdjacencyMap a -> [a]
-reachable x = dfs [x]
+reachable x = concat . bfs [x]
```
snowleopard: Same here: comments/examples/tests are out-of-date.

src/Data/Graph/Typed.hs
```diff
@@ -181,3 +181,4 @@ dfs vs = concatMap flatten . dfsForestFrom vs
 -- @
 topSort :: GraphKL a -> [a]
 topSort (GraphKL g r _) = map r (KL.topSort g)
+
```
snowleopard: One more leftover bit.

test/Algebra/Graph/Test/API.hs
```diff
@@ -100,6 +100,9 @@ data API g c where
         , dfsForest                  :: forall a. c a => g a -> Forest a
         , dfsForestFrom              :: forall a. c a => [a] -> g a -> Forest a
         , dfs                        :: forall a. c a => [a] -> g a -> [a]
+        , bfsForest                  :: forall a. c a => g a -> Forest a
```
snowleopard: Please match the order elsewhere. Same below.

test/Algebra/Graph/Test/AdjacencyIntMap.hs
```diff
@@ -44,6 +44,9 @@ testAdjacencyIntMap = do
     testDfsForest            t
     testDfsForestFrom        t
     testDfs                  t
+    testBfsForest            t
```
snowleopard: Let's keep the order consistent everywhere :)

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
snowleopard: Actually, `s` is a bit awkward, since in other modules `s` stands for the `size` of algebraic expressions. Let's keep `L` -- in fact, I already use `L` to denote the length of the input list in some complexity bounds.

snowleopard:
> Maybe I should add a comment in the documentation about `concat . bfs x` returning the vertices in bfs-order?

@jitwit Yes, that would be useful! Perhaps, you can even add it as an example, using the same bidirectional circuit graph.

jitwit:
@snowleopard Hopefully I've not left any documentation/tests out of sync this time! 

Also, I hope the documentation changes are reasonable? Wikipedia seems to define level structure mentioning undirected graphs, but no mention of directed ones, so I can change it back if this goes against the standard definition.

[https://en.wikipedia.org/wiki/Level_structure#cite_note-dps-1](https://en.wikipedia.org/wiki/Level_structure#cite_note-dps-1)

test/Algebra/Graph/Test/AdjacencyIntMap.hs
jitwit: Haha, hopefully I've fixed them all by now

snowleopard:
@jitwit The performance suite has been fixed (thanks @nobrakal!) and now we have the following improvement (https://travis-ci.org/snowleopard/alga/jobs/558388854):

```
reachable: 0.78 (good)
```

That's pretty good! Note that this is benchmarking `Algebra.Graph`, so the increase in performance is masked somewhat by the time it takes to convert a `Graph` to `AdjacencyMap`.

jitwit:
Cool! Nice to see the benefits play out

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -31,6 +35,63 @@ import qualified Data.Graph.Typed   as Typed
...
+--   level structure.  Flattening the result via @'concat' . 'bfs' vs@
+--   gives an enumeration of vertices reachable from @vs@ in breadth
+--   first order.
+--
+--   Let /L/ be the number of seed vertices. Complexity:
+--   /O(n+(L+m)*log n)/ time and /O(n+m)/ space.
+-- 
+-- @
+-- bfs [3] ('circuit' [1..5] + 'circuit' [5,4..1])          == [[3],[2,4],[1,5]]
+-- 'concat' (bfs [3] $ 'circuit' [1..5] + 'circuit' [5,4..1]) == [3,2,4,1,5]
```
snowleopard: Is this always the case that vertices in a single level appear in sorted order? Probably not. In either case, I think it's useful to document this.

test/Algebra/Graph/Test/Generic.hs
```diff
@@ -1692,6 +1692,144 @@ testSimplify (prefix, API{..}) = do
...
+                               , Node { rootLabel = 4
+                                      , subForest = [ Node { rootLabel = 5
+                                                           , subForest = []}]}]}]
+          
+testBfs :: TestsuiteInt g -> IO ()
+testBfs (prefix, API{..}) = do
+    putStrLn $ "\n============ " ++ prefix ++ "bfs ============"
+    
+    test "bfs vs    $ empty                    == []" $ \vs ->
+          bfs vs      empty                    == []
```
snowleopard: You have a lot of tests here, which I think would be useful examples. Could you add them to the docs?

snowleopard:
@jitwit I think it's pretty much ready for merge, but see a couple of more comments.

P.S.: Please add your name to `AUTHORS.md` if you like to be listed.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: Yeah, for 2 and 4, one of the levels is not in sorted order. I changed the test case and examples to start with 2 to illustrate this. I also added a remark in the documentation.

test/Algebra/Graph/Test/Generic.hs
jitwit: Done! Some of the lines are awkwardly long when showing the forest. Is it handled ok? Or should all the equals signs be aligned?

jitwit:
@snowleopard Nice! Also, fingers crossed I didn't get any tests/docs out of sync. I tried to look carefully!

test/Algebra/Graph/Test/Generic.hs
snowleopard: I try to keep all `==` aligned both in Haddock and in the testsuite output, if possible. Happy to do the final polishing myself though.

snowleopard:
@jitwit Merged, thank you! If you are not fed up by the long review times, it would be great if you could look into improving DFS algorithms, as discussed above :)

jitwit:
@snowleopard nice! haha, definitely, I'll start trying some ideas for DFS

PR #218, opened 2019-06-28
Add detectParts to Bipartite.AdjacencyMap module

vasalf:


snowleopard:
Thanks! How difficult would it be to return an odd cycle in case of failure? This is occasionally useful.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -665,3 +674,120 @@ edgeSet = Set.fromAscList . edgeList
 -- @
 consistent :: (Ord a, Ord b) => AdjacencyMap a b -> Bool
 consistent (BAM lr rl) = internalEdgeList lr == sort (map Data.Tuple.swap $ internalEdgeList rl)
+
+
+data PartId = LeftPart | RightPart
```
snowleopard: Just `Part` (instead of `PartId`) would be fine.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -665,3 +674,120 @@ edgeSet = Set.fromAscList . edgeList
...
+
+        toEither :: Ord a => Map.Map a PartId -> a -> Either a a
+        toEither m v = case fromJust (v `Map.lookup` m) of
+                            LeftPart  -> Left  v
+                            RightPart -> Right v
+
+
+type PartMap a = Map.Map a (PartId, NE.NonEmpty a)
+
+detectParts' :: Ord a => AM.AdjacencyMap a -> Either [a] (AdjacencyMap a a)
```
snowleopard: Is this function as fast as `detectParts` in the `Right` case, i.e. if the graph is bipartite? If not and it is substantially slower than `detectParts`, then I think it's not worth combining the two tasks in one, and having a simpler `findOddCycle` would actually be better.

snowleopard: Another question: Is there a more general and hopefully simpler function that we could write, so that both `findParts` and `findOddCycle` could be implemented by calling it?

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -665,3 +674,120 @@ edgeSet = Set.fromAscList . edgeList
...
+-- 'Data.Maybe.isJust' (detectParts ('Algebra.Graph.AdjacencyMap.edge' x y))               == True
+-- 'Data.Maybe.isJust' (detectParts ('Algebra.Graph.AdjacencyMap.edges' [(1, 2), (1, 3)])) == True
+-- 'Data.Maybe.isJust' (detectParts ('Algebra.Graph.AdjacencyMap.clique' x))               == x < 3
+-- 'Data.Maybe.isJust' (detectParts ('Algebra.Graph.AdjacencyMap.star' x ys))              == not (x `elem` ys)
+-- 'Data.Maybe.isJust' (detectParts ('Algebra.Graph.AdjacencyMap.biclique' xs ys))         == True
+-- 'Data.Maybe.isJust' ('fromBipartite' ('toBipartite' am))       == True
+-- 'Data.Maybe.isJust' (detectParts (1 * 2 * 3))              == False
+-- 'Data.Maybe.isJust' (detectParts ((1 + 2) * (3 + 4)))      == True
+-- @
+detectParts :: Ord a => AM.AdjacencyMap a -> Maybe (AdjacencyMap a a)
```
snowleopard: How does this function behave if the answer is not unique? Is the result optimal in some sense, e.g. lexicographically smallest?

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
vasalf: I don't know any linear algorithm for finding an odd cycle other than extending the bipartiteness testing algorithm in some way like here. What do you mean by "simplier `findOddCycle`?"

snowleopard: > What do you mean by "simplier findOddCycle?"

I guess if you specialise this function to just finding the odd cycle, then it will become simpler. Or am I wrong? Are you saying that if the type signature were `findOddCycle :: Ord a => AM.AdjacencyMap a -> Maybe [a]` then the function would essentially remain the same?

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
vasalf: *Any* result for a non-empty graph is not unique, as we can always swap the parts.

I didn't think of a completely formal proof, but it seems to me that the bit string of the part identifiers for each vertex is lexicographically minimal. It seems that for a connected non-empty graph there are always exactly two unique answers. If so, then it is correct: each component is visited in its minimal vertex for the first time and this vertex is put into the left part.

snowleopard: Sounds good, thanks. Let's document this in comments, and perhaps, add an example and test that demonstrate this.

snowleopard: Perhaps, the existing examples are already sufficient, in which case you can just say "see examples" in the comment :)

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
vasalf: > the function would essentially remain the same

It really seems so to me. I'm going to have a try on making `findOddCycle` simpler anyway.

There are chances as the monadic idiom here is different. And for those usages of `Maybe` we already know that there are chances that DFS will look simpler.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
vasalf: They are not, as for now we know it, we can add something like
```
detectParts (1 * 2 + 3 * 4) = Just (edges [(1, 2), (3, 4)])
```
This seems to look better than
```
isJust (detectParts (1 * 2 + 3 * 4)) == True
``

Comment (snowleopard) [reply src/Algebra/Graph/Bipartite/AdjacencyMap.hs]: Indeed, `detectParts (1 * 2 + 3 * 4) = Just (edges [(1, 2), (3, 4)])` is much more informative.

File: src/Algebra/Graph/Bipartite/AdjacencyMap.hs [hunk 4]
```diff
@@ -650,6 +661,116 @@ vertexSet = Set.fromAscList . vertexList
...
+otherPart RightPart = LeftPart
+
+type PartMap a = Map.Map a Part
+type PartMonad a r = MaybeT (State (PartMap a)) r
+
+-- | An odd cycle.
+type OddCycle a = [a] -- TODO: Make this representation type-safe
+
+neighbours :: Ord a => a -> AM.AdjacencyMap a -> [a]
+neighbours v = Set.toAscList . fromJust . Map.lookup v . AM.adjacencyMap
```
snowleopard: Please avoid partial functions like `fromJust`.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -650,6 +661,116 @@ vertexSet = Set.fromAscList . vertexList
...
+    deriving (Show, Eq)
+
+otherPart :: Part -> Part
+otherPart LeftPart  = RightPart
+otherPart RightPart = LeftPart
+
+type PartMap a = Map.Map a Part
+type PartMonad a r = MaybeT (State (PartMap a)) r
+
+-- | An odd cycle.
```
snowleopard: I think it's worth clarifying how a cycle is represented. For example, a cycle `1 -> 2 -> 3 -> 1` is represented by `[1, 2, 3]`.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -650,6 +661,116 @@ vertexSet = Set.fromAscList . vertexList
...
+-- whether a vertex belongs to a part.
+--
+-- Complexity: /O((n + m) log(n))/ time and /O(n + m)/ memory.
+--
+-- @
+-- detectParts 'Algebra.Graph.AdjacencyMap.empty'                                      == Right 'empty'
+-- detectParts ('Algebra.Graph.AdjacencyMap.vertex' x)                                 == Right ('leftVertex' x)
+-- detectParts (1 * (2 + 3))                              == Right ('edges' [(1, 2), (1, 3)])
+-- detectParts ((1 + 3) * (2 + 4) + 6 * 5)                == Right ('swap' (1 + 3) * (2 + 4) + 5 * 6)
+-- 'Data.Either.isRight' (detectParts ('Algebra.Graph.AdjacencyMap.edge' x y))                       == True
```
snowleopard: I think it's best to avoid `isRight` examples -- better switch to more informative `== Right ...` instead.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -650,6 +661,116 @@ vertexSet = Set.fromAscList . vertexList
...
+-- 'Data.Either.isRight' (detectParts ('Algebra.Graph.AdjacencyMap.clique' x))                       == x < 3
+-- 'Data.Either.isRight' (detectParts ('Algebra.Graph.AdjacencyMap.star' x ys))                      == True
+-- 'Data.Either.isRight' (detectParts ('Algebra.Graph.AdjacencyMap.biclique' xs ys))                 == True
+-- 'Data.Either.isRight' (detectParts ('fromBipartite' ('toBipartite' am))) == True
+-- 'Data.Either.isRight' (detectParts (1 * 2 * 3))                      == False
```
snowleopard: And, similarly, instead of `isRight ... == False` better give `== Left ...` examples instead.

test/Algebra/Graph/Test/Bipartite/AdjacencyMap.hs
```diff
@@ -626,29 +628,72 @@ testBipartiteAdjacencyMap = do
...
 expectedBicliqueMap n m = Map.fromAscList [ (u, Set.fromAscList [1..m]) | u <- [1..n] ]
 
 isSorted :: Ord a => [a] -> Bool
 isSorted xs = and $ zipWith (<=) xs $ tail xs
+
+fromEither :: Either a a -> a
+fromEither (Left  x) = x
+fromEither (Right y) = y
+
+cycleGraph :: Ord a => [a] -> AM.AdjacencyMap a
```
snowleopard: Is this just `circuit`?

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
vasalf: In this particular test, it would be `Right (edge x y)` in case `x < y` and `Right (edge y x)` in case `x > y`.

vasalf: Well, in fact, this particular example is incorrect, for example, `detectParts (edge 1 1) == Left [1]`. But generally, it doesn't seem we could provide a complete definition at the right side.

vasalf: Actually, `detectParts (edge 1 1) == Left [1]` should be an example.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
vasalf: I don't see a formal property of the returned cycle we could guarantee to provide concrete examples in the docs. And even if the cycle is unique, we can't guarantee the order of vertices in it. For example, `detectParts (1 * 2 * 3)` might be equal to `Left [1, 2, 3]` or to `Left [2, 1, 3]` depending on the implementation. 
It is possible to provide some example like
```
 bimap sort id (detectParts (1 * 2 * 3)) == Left [1, 2, 3]
```
but this seems too complicated to me.

test/Algebra/Graph/Test/Bipartite/AdjacencyMap.hs
vasalf: Oh, thanks. It is not actually `circuit`, but `circuit` *is* what I really need there.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
snowleopard: Aha, let's have both examples then:
```
detectParts (edge 1 1) == Left [1]
detectParts (edge 1 2) == Right (edge 1 2)
```
Very clear and useful!

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
vasalf: ...or it could be `[2,3,1]` as well.

vasalf: It should be vice verse — [1, 2, 3] represents 1 -> 2 -> 3 -> 1.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
snowleopard: Why not provide `detectParts (1 * 2 * 3) == Left [1, 2, 3]` if that's how the current implementation works? At least it will give the user a clue how the current implementation handles this particular case.

In general, it would of course be great if you could precisely describe the odd cycle that is obtained. Isn't it just "lexicographically least odd cycle"? If not, can you give a counterexample?

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
snowleopard: > It should be vice versa — [1, 2, 3] represents 1 -> 2 -> 3 -> 1.

Yes, this formulation seems better.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
vasalf: For now, `detectParts (1 * 2 * 3) == Left [2, 3, 1]`.

There is a simple example where even the set of vertices in the returned odd cycle is not lexicographically minimal: 
```
detectParts (1 * (3 + 6) + 3 * 4 * 5 + 2 * 6 * 7) == Left [4, 5, 3]
```
But there is a cycle `[2, 6, 7]`. 

The problem is that the property "first odd cycle found by the depth-first search that goes through vertices in ascending order" doesn't seem to tell anything special at all. For instance, in the above example it goes `1 -> 3 -> 4 -> 5 -> 3` and doesn't even visit vertex 2.

vasalf: I added `detectParts (1 * 2 * 3) == Left [2, 3, 1]`.

vasalf: After discussion with @snowleopard we agreed on the following definition of optimality of the returned odd cycle:

> There exists a path that is either empty or ends in a vertex adjacent to the first vertex in the cycle, such that all vertices in (path ++ cycle) are distinct and (path ++ cycle) is lexicographically minimal among all such pairs.

For example, in `(1 * 3 * 4) + 2 * (1 + 2)` we will return the cycle `[2]` and the path is `[1]`. Note that we can't compare pairs (path, cycle), because the pair `([], [1, 3, 4])` is valid and lexicographically smaller than the pair `([1], [2])`.

So the example finally becomes `detectParts (1 * 2 * 3) == Left [1, 2, 3]`.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -32,19 +33,30 @@ module Algebra.Graph.Bipartite.AdjacencyMap (
...
+import Data.List                 (sort, (\\))
+import Data.Maybe                (fromJust)
 import GHC.Generics
 
 import qualified Algebra.Graph              as G
 import qualified Algebra.Graph.AdjacencyMap as AM
 
-import qualified Data.Map.Strict as Map
-import qualified Data.Set        as Set
+import qualified Data.Map.Strict    as Map
```
snowleopard: This looks like an accidental edit.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -650,6 +662,141 @@ vertexSet = Set.fromAscList . vertexList
 edgeSet :: (Ord a, Ord b) => AdjacencyMap a b -> Set.Set (a, b)
 edgeSet = Set.fromAscList . edgeList
 
+-- | The /circuit/ on a list of vertices.
+-- Complexity: /O(n * log(n))/ time and /O(n)/ memory.
+--
+-- @
+-- circuit []                       == 'empty'
+-- circuit [(x, y)]                 == 'edge' x y
+-- circuit [(x, y), (z, w)]         == 'connect' ('vertices' [x, z] []) ('vertices' [] [y, w])
```
snowleopard: Should the right-hand side be `biclique [x, z] [y, w]`?

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -650,6 +662,141 @@ vertexSet = Set.fromAscList . vertexList
...
+-- circuit [(1, 2), (3, 4), (5, 6)] == swap 1 * (2 + 6) + swap 3 * (2 + 4) + swap 5 * (6 + 2)
+-- circuit . 'reverse'              == 'swap' . circuit . 'map' 'Data.Tuple.swap'
```
snowleopard: That's a mind-bending property, I had to draw a picture to see what's going on. Nice :)

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -650,6 +662,141 @@ vertexSet = Set.fromAscList . vertexList
...
+-- @
+circuit :: (Ord a, Ord b) => [(a, b)] -> AdjacencyMap a b
+circuit l = overlay (edges l) (edges (zipWith (,) (map fst l) (rotateSnd l)))
+    where
+        rotateSnd :: [(a, b)] -> [b]
+        rotateSnd l = (snd (last l)):(map snd (init l))
```
snowleopard: There are some redundant parentheses.

snowleopard: Also please try to avoid partial functions such as `last` and `init`.

snowleopard: Here is a safer and terser implementation:

```haskell
circuit :: (Ord a, Ord b) => [(a, b)] -> AdjacencyMap a b
circuit es = edges $ es ++ zip (drop 1 $ cycle as) bs
  where 
    (as, bs) = unzip es
```

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -650,6 +662,141 @@ vertexSet = Set.fromAscList . vertexList
...
+-- @
+-- biclique [] [] == 'empty'
+-- biclique xs [] == 'vertices' xs []
+-- biclique [] ys == 'vertices' [] ys
+-- biclique xs ys == 'connect' ('vertices' xs []) ('vertices' [] ys)
+-- @
+biclique :: (Ord a, Ord b) => [a] -> [b] -> AdjacencyMap a b
+biclique xs ys = let sxs = sort xs
+                     sys = sort ys
+                  in BAM (Map.fromAscList $ zipWith (,) sxs $ repeat (Set.fromAscList sys))
```
snowleopard: I think `zipWith (,)` is just `zip` :)

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -650,6 +662,141 @@ vertexSet = Set.fromAscList . vertexList
...
+type PartMonad a = MaybeT (State (PartMap a)) [a]
+
+-- | An odd cycle. For example, @[1, 2, 3]@ represents the cycle 1 → 2 → 3 → 1.
+type OddCycle a = [a] -- TODO: Make this representation type-safe
+
+neighbours :: Ord a => a -> AM.AdjacencyMap a -> [a]
+neighbours v = Set.toAscList . AM.postSet v
+
+partMonad :: Ord a => AM.AdjacencyMap a -> PartMonad a
+partMonad g = asum $ map (runDfs g) $ AM.vertexList g
```
snowleopard: It's a minor tweak, but I often find list comprehensions easier to understand than layers of `map`s.

```suggestion
partMonad g = asum [ runDfs g v | v <- AM.vertexList g ]
```

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -650,6 +662,141 @@ vertexSet = Set.fromAscList . vertexList
...
+                                          then return [v]
+                                          else mzero
+
+        dfs :: Ord a => AM.AdjacencyMap a -> Part -> a -> PartMonad a
+        dfs g p v = do modify $ Map.insert v p
+                       let q = otherPart p
+                       asum $ map (action g q) $ neighbours v g
+
+        runDfs :: Ord a => AM.AdjacencyMap a -> a -> PartMonad a
+        runDfs g v = ((:) v) <$> do m <- get
```
snowleopard: There is no need to pass `g` to all these functions, since they can all see it!

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -650,6 +662,141 @@ vertexSet = Set.fromAscList . vertexList
...
+                          case v `Map.lookup` m of
+                               Nothing -> ((:) v) <$> dfs g p v
+                               Just q  -> if q /= p
+                                          then return [v]
+                                          else mzero
+
+        dfs :: Ord a => AM.AdjacencyMap a -> Part -> a -> PartMonad a
+        dfs g p v = do modify $ Map.insert v p
+                       let q = otherPart p
+                       asum $ map (action g q) $ neighbours v g
```
snowleopard: Same here: try using list comprehensions.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -650,6 +662,141 @@ vertexSet = Set.fromAscList . vertexList
...
+partMonad g = asum $ map (runDfs g) $ AM.vertexList g
+    where
+        {-# INLINE action #-}
+        action :: Ord a => AM.AdjacencyMap a -> Part -> a -> PartMonad a
+        action g p v = do m <- get
+                          case v `Map.lookup` m of
+                               Nothing -> ((:) v) <$> dfs g p v
+                               Just q  -> if q /= p
+                                          then return [v]
+                                          else mzero
```
snowleopard: Since you are already using `guard` elsewhere, perhaps you should simplify:

```haskell
Just q -> do guard (p == q)
             return [v]
```

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
vasalf: Thanks, that looks better!

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
vasalf: Good point!
After reviewing this code I replaced with `Map.fromSet`, so this is going to be outdated.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
vasalf: Sure, thanks! Much better this way.

vasalf: Just a small note: that should definitely be `guard (p /= q)`: we return `[v]` if `p /= q`, e. g. we try to put a vertex that is already assigned to some part into another part.

snowleopard: Ah, OK, it must be a case of Boolean blindness :)

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -774,25 +767,45 @@ partMonad g = asum $ map (runDfs g) $ AM.vertexList g
...
 -- 'Data.Either.isRight' (detectParts ('fromBipartite' ('toBipartite' x)))   == True
 -- @
-detectParts :: Ord a => AM.AdjacencyMap a -> Either (OddCycle a) (AdjacencyMap a a)
-detectParts g = let s = AM.symmetricClosure g
-                 in case runState (runMaybeT $ partMonad s) Map.empty of
-                         (Nothing, m) -> Right $ build m g
-                         (Just c,  _) -> Left  $ oddCycle c
+detectParts :: forall a. Ord a => AM.AdjacencyMap a -> Either (OddCycle a) (AdjacencyMap a a)
+detectParts x = case runState (runMaybeT $ dfs) Map.empty of
+                     (Nothing, m) -> Right $ toBipartiteWith (toEither m) g
```
snowleopard: It looks like this function may be doing more than it needs to: it could return `a -> Either b c` as a partition function instead of a returning an actual bipartite graph. The user could then use `toBipartiteWith` if need be.

We could also add a type synonym `type Partition a b c = a -> Either b c`.

vasalf: For the history sake, I'm copying the discussion here.

This seemed a really nice idea, but there are several problems with it, the main being that it's not clear what should be returned by `Partition` in case it's given a vertex that is not present in the graph. Some workaround could be suggested, for example,

```Haskell
type Partition a b c = a -> Maybe (Either b c)
```

But there are still problems with this, particularly that `Nothing` could be returned for a present vertex and that this one is anyway too complicated for use.

We'll keep the current implementation with `Either (OddCycle a) (AdjacencyMap a a)`.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -413,6 +426,20 @@ toBipartite m = BAM (Map.fromAscList [ (u, setRights vs) | (Left  u, vs) <- symm
...
+-- Complexity: /O(m log(n))/
+--
+-- @
+-- toBipartiteWith f Algebra.Graph.AdjacencyMap.'Algebra.Graph.AdjacencyMap.empty' == 'empty'
+-- toBipartiteWith Left  x  == 'empty'
+-- toBipartiteWith Right x  == 'empty'
+-- toBipartiteWith f        == 'toBiparitite' . Algebra.Graph.AdjacencyMap.'Algebra.Graph.AdjacencyMap.gmap' f
+-- @
+toBipartiteWith :: (Ord a, Ord b, Ord c) => (a -> Either b c) -> AM.AdjacencyMap a -> AdjacencyMap b c
+toBipartiteWith f = toBipartite . AM.gmap f
```
snowleopard: Wouldn't it be faster to implement `toBipartiteWith` directly and then say `toBipartite = toBipartiteWith id`?

vasalf: I'm sure it wouldn't, because values returned by an arbitrary function are not necessarily monotonic and distinct. In this case, in `toBipartite` we are going to lose some time on `Set.fromList` instead of `Set.fromAscList` and so on.

A direct implementation of `toBipartiteWith` in same fashion would be something like this:
```Haskell
toBipartiteWith :: (Ord a, Ord b, Ord c) => (a -> Either b c) -> AM.AdjacencyMap a -> AdjacencyMap b c
toBipartiteWith f m = BAM (Map.fromListWith Set.union lr)
                          (Map.fromListWith Set.union rl)
    where
        lr = do (u, vs) <- symmetricList
                case f u of
                    Left  w -> return (w, setRights vs)
                    Right _ -> []
        rl = do (u, vs) <- symmetricList
                case f u of
                    Left  _ -> []
                    Right w -> return (w, setLefts vs)
        setLefts  = Set.fromList . lefts  . map f . Set.toAscList
        setRights = Set.fromList . rights . map f . Set.toAscList
        symmetricList = Map.toAscList $ AM.adjacencyMap $ AM.symmetricClosure m
```

Again, I'm sure that `toBipartite = toBipartiteWith id` is not faster and it is very likely that it would be slower.

snowleopard: I see, thanks for looking into this. I agree with your conclusions.

snowleopard:
Merged. Many thanks @vasalf!

Issue #141, opened 2018-11-15
Explore alternative implementations for AdjacencyMap


snowleopard:
We currently store vertices of type `a` directly in an `AdjacencyMap a`, which may lead to poor performance if the corresponding `Ord a` instance is slow. One example, which motivated me to record this issue, is the SCC algorithm developed in #128 that produces `AdjacencyMap (NonEmpty.AdjacencyMap a)`, i.e. graphs whose vertices are graphs.

An alternative would be to define something like
```haskell
data AdjacencyMap a = AdjacencyMap
    { graph :: AdjacencyIntMap
    , value :: Int -> a
    , index :: a -> Maybe Int }
```
This is similar to `Data.Graph.graphFromEdges`:

http://hackage.haskell.org/package/containers-0.6.0.1/docs/Data-Graph.html#v:graphFromEdges

In this way, we can benefit from the high performance of the underlying `AdjacencyIntMap` yet provide a typed high-level interface `AdjacencyMap a` to the user.

michaelpj:
Perhaps more generally, it would be nice to be able to provide a bijective map between the "real" node type and a "key" type which is cheaper to store, or has a cheaper (or existent) `Ord` instance.

michaelpj:
Note that this can be a blocker for moving from `Data.Graph`, which allows the alternative key type. I think I just rediscovered why past-me made that comment ;)

Avasil:
I will create a prototype

Avasil:
@snowleopard I finally found some time to start working on this and I'm wondering about operations that combine graphs, e.g. `overlay`.

If we define `value :: Int -> a` and `index :: a -> Maybe Int` on graph itself, there could be two graphs with different mappings, right?

I guess we could combine `index` functions but I don't know what to do with `value` which seems to be total.

This is of course if I'm understanding the issue correctly which is to check if this representation of `AdjacencyMap` could improve anything and if so, rewrite API to forward to `AdjacencyIntMap`

snowleopard:
> there could be two graphs with different mappings, right?

Yes.

> I guess we could combine `index` functions but I don't know what to do with `value` which seems to be total.

Actually, I'm not entirely sure my draft above is correct. What we really need to represent here is an isomorphism between vertices `a` that appear in the graph and their integer counterparts `Int`. Do you have an idea of how we could do this both efficiently and in a compositional way to permit operations like `overlay`?

[...]

PR #219, opened 2019-07-10
Labelled acyclic graph

adithyaov:
This is a draft implementation of labeled acyclic graph required if one wants to make algorithms on acyclic graphs. There is still a lot of work to be done including writing tests.


Avasil:
> There is still a lot of work to be done including writing tests.

If you'd like to split some of it, I would be happy to lend a hand!

adithyaov:
@Avasil Of Course :-).
Some tasks that I find incomplete are written as `TODO`'s in `AdjacencyMap.hs`.
We need to write tests for the module as well.
Please ignore the `TODO` that says `Improve the Show instance`. It's too complex to achieve in a short amout of time. That should probably be done in another PR later.

Avasil:
Awesome!
Should I add things to your PR or wait after it's merged?

I will have a limited access to the internet for a couple of days (Saturday - Wednesday) so I was thinking about waiting until I come back

adithyaov:
> Should I add things to your PR or wait after it's merged?

Without adding the function `consistent` and tests, the PR probably should not be merged.

> I will have a limited access to the internet for a couple of days (Saturday - Wednesday) so I was thinking about waiting until I come back.

Ah, I see. In that case, for the time being, I'll add the `consistent` function and the tests and hopefully that is sufficient for the PR to be merged. You can continue working on the rest of the `TODO`'s or add your own once you come back.

src/Algebra/Graph/Acyclic/Labelled/AdjacencyMap.hs
```diff
@@ -0,0 +1,352 @@
...
+
+import Data.Set (Set)
+import Algebra.Graph.Label
+import Data.Function (on)
+
+import qualified Algebra.Graph.Labelled.AdjacencyMap as AM
+
+-- | Edge-labelled graphs, where the type variable @e@ stands for edge labels.
+-- For example, 'AdjacencyMap' @Bool@ @a@ is isomorphic to unlabelled acyclic graphs
+-- defined in the top-level module "Algebra.Graph.AdjacencyMap", where @False@
```
snowleopard: I guess you meant `Algebra.Graph.Acyclic.AdjacencyMap` here?

src/Algebra/Graph/Acyclic/Labelled/AdjacencyMap.hs
```diff
@@ -0,0 +1,352 @@
...
+-- Complexity: /O(1)/ time and memory.
+--
+-- @
+-- 'isEmpty'     empty == True
+-- 'hasVertex' x empty == False
+-- 'vertexCount' empty == 0
+-- 'edgeCount'   empty == 0
+-- @
+empty :: AdjacencyMap e a
+empty = AAM AM.empty
```
snowleopard: This, as well as many other functions, can be implemented via `coerce` to ensure that they are zero-cost.

adithyaov: Unfortunately, I am encountering an error when I use `coerce`,
image
I tried writing my own `Coercible` instance but I get an "Illegal instance declaration" error.
I am still trying to figure out how to solve the error.

snowleopard: See this comment for a similar issue and solution: https://github.com/snowleopard/alga/pull/220#issuecomment-510891564

adithyaov: @snowleopard Thanks for the help. I've changed the required.

adithyaov:
@jitwit I agree, this is not completely Dijkstra. But for an acyclic graph one does not need to maintain a set of nodes. One can directly relax the edges in the topological order. I think one can consider this as a simplification of Dijkstra for acyclic graphs.

> In CLSR I think they name it dag-shortest-paths for what it's worth!

Ah yes, I think I'll rename it to `shortestPath` or `dagShortestPath`. Thank you for the feedback.

test/Algebra/Graph/Test/Acyclic/Labelled/AdjacencyMap.hs
```diff
@@ -0,0 +1,237 @@
...
+import qualified Data.List as List
+import qualified Data.Set as Set
+
+type S = Sum Int
+type D = Distance Int
+
+type LAS = AdjacencyMap S   Int
+type LAD = AdjacencyMap D   Int
+
+-- TODO: Switch to using generic tests.
```
Avasil: I could start with this one but I'm not sure what you mean. Is it about adding it to `API.hs`?
Also, how should I incorporate my change to your branch?

adithyaov: > I could start with this one but I'm not sure what you mean. Is it about adding it to API.hs?

I suggest you leave that one as well. One needs to make separate generic tests cases for acyclic graphs as the acyclic API does not have edges yet (or) make a `*.Ord` module similar to `Acyclic.Ord` to make a complete API using functions from both the modules. Creating an issue regarding this would help start a discussion.

I would also suggest you leave out `TODO: Make 'skeleton :: Ord a => AdjacencyMap e a -> AAM.AdjacencyMap a'` and `TODO: Add 'fromGraph'`. In light of the recent PR#223, some functions might be removed.

`TODO: Add additional/change examples.` This would be a good place to start. This PR needs more examples and the corresponding tests to be written.

> Also, how should I incorporate my change to your branch?

If it is not too much hassel, one way to do this is to create a PR to my fork of the repository. I'll merge the PR accordingly.

Please let me know if you have any other method in mind.

snowleopard:
> @jitwit The following describes a minor issue.
> This is equivalent to the shortest path if the edges are distances, If the edges are capacities then this is
equivalent to finding the max capacity and hence I feel giving the name `shortestPath` is not semantically correct. If possible could you please suggest some generic name?

Indeed! In fact, we could probably use the same algorithm for finding the longest path too, which would be even more confusing :)

Perhaps, we could call it `optimumPath`? I would then match with the `Optimum` data type:

https://hackage.haskell.org/package/algebraic-graphs-0.4/docs/Algebra-Graph-Label.html#t:Optimum

We could also look at how such generic path-finding algorithms are called in the literature.

test/Algebra/Graph/Test/Acyclic/Labelled/AdjacencyMap.hs
snowleopard: @Avasil It's a bit difficult to contribute to this part of the codebase at the moment because both @adithyaov and I are making some changes. I can try to find an interesting isolated task for you if you like :)

Avasil: That makes sense, I'll wait a bit until it stabilizes.

> I can try to find an interesting isolated task for you if you like :)

@snowleopard That would be perfect!

jitwit:
It looks like in the Mohri paper `Generic-Topological-Single-Source-Shortest-Distance` is used, which is a mouthful. In response to @adithyaov's question about `Capacity` giving maximum, I guess Mohri slapped generic on the front to indicate both shortest and distance should be understood abstractly!

Huang paper attributes shortest paths from topological sorting to Viterbi (1967) and calls the algorithm `Viterbi`, though in a footnote he mentions it's sometimes also credited to Lawler (1976).

Huang: [https://www.aclweb.org/anthology/C08-5001](https://www.aclweb.org/anthology/C08-5001)
Mohri: [http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.367.5897&rep=rep1&type=pdf](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.367.5897&rep=rep1&type=pdf)

snowleopard:
@jitwit Thanks for the pointers! I haven't seen Huang's paper before.

From these options, I'm still in favour of `optimumPath`.

jitwit:
@snowleopard Yeah, I find Viterbi confusing since it makes me think the semiring is specialized! Also, dag is kind of in the module name, so it's probably unnecessary to include that in the function name. 

`optimumPath` is concise and descriptive. Maybe it should be plural since a `Map a e` is produced?

snowleopard:
> Maybe it should be plural since a `Map a e` is produced?

@jitwit I'd stick with the singular name because the source vertex is fixed. I've seen both singular and plural used in the literature, but I generally prefer using singular names in library APIs.

test/Algebra/Graph/Test/Acyclic/Labelled/AdjacencyMap.hs
snowleopard: @Avasil Here is an interesting issue which I haven't had time to think about: #141. It would be nice to make an experiment about whether it's practical to implement a generic `AdjacencyMap a` by keeping `AdjacencyIntMap` internally together with a bijection between `a` and `Int`.

src/Algebra/Graph/Acyclic/Labelled/Algorithm.hs
```diff
@@ -0,0 +1,35 @@
...
+-- from a given source vertex.
+--
+-- The following examples assume that the edges are distances,
+-- ie. the edge 'Semiring' is 'Distance'.
+-- @
+-- dijkstra ('LAM.toAcyclicOrd' $ 'LAM.edges' [(2, 'b', 'c'), (1, 'a', 'b'), (3, 'a', 'c')]) 'z' == Map.'Map.fromList' [('z', 0), ('a', Infinite), ('b', Infinite), ('c', Infinite)]
+-- dijkstra ('LAM.toAcyclicOrd' $ 'LAM.edges' [(2, 'b', 'c'), (1, 'a', 'b'), (3, 'a', 'c')]) 'a' == Map.'Map.fromList' [('a', 0), ('b', 1), ('c', 3)]
+-- @
+optimumPath :: (Semiring e, Ord a) => AdjacencyMap e a -> a -> Map a e
+optimumPath am s = foldl (relaxVertex em) (initialize em) vL
```
snowleopard: Are you sure you don't need a strict version of `foldl` here, as explained [here](https://wiki.haskell.org/Foldr_Foldl_Foldl')?

src/Algebra/Graph/Acyclic/Labelled/Algorithm.hs
```diff
@@ -0,0 +1,35 @@
...
+
+-- TODO: Add time complexity
+-- TODO: Add examples using 'Optimum' data type
+-- | Compute the /shortest path/ to each vertex in the graph
+-- from a given source vertex.
+--
+-- The following examples assume that the edges are distances,
+-- ie. the edge 'Semiring' is 'Distance'.
+-- @
+-- dijkstra ('LAM.toAcyclicOrd' $ 'LAM.edges' [(2, 'b', 'c'), (1, 'a', 'b'), (3, 'a', 'c')]) 'z' == Map.'Map.fromList' [('z', 0), ('a', Infinite), ('b', Infinite), ('c', Infinite)]
```
snowleopard: Still old name.

src/Algebra/Graph/Acyclic/Labelled/Algorithm.hs
```diff
@@ -0,0 +1,35 @@
...
+-- dijkstra ('LAM.toAcyclicOrd' $ 'LAM.edges' [(2, 'b', 'c'), (1, 'a', 'b'), (3, 'a', 'c')]) 'a' == Map.'Map.fromList' [('a', 0), ('b', 1), ('c', 3)]
+-- @
+optimumPath :: (Semiring e, Ord a) => AdjacencyMap e a -> a -> Map a e
+optimumPath am s = foldl (relaxVertex em) (initialize em) vL
+  where
+    vL = dropWhile (/=s) $ topSort am
+    em = (LAM.adjacencyMap . fromAcyclic) am
+    initialize = Map.insert s one . Map.map (const zero)
+    relaxVertex em m v = Map.foldrWithKey (relaxEdge v) m (em ! v)
+    relaxEdge v1 v2 e m = Map.insert v2 (((m ! v1) <.> e) <+> (m ! v2)) m
```
snowleopard: It feels to me that here you make use of the fact that `e` is a `Dioid`, not just a `Semiring`.

snowleopard:
@adithyaov A general comment: it looks to me that there is a more general function that can be factored out of your implementation. Something `fold`-like, but for an acyclic graph, where you start with an initial state `s` and then traverse the graph in the topological order, at each step modifying the state with a function like `s -> a -> [a] -> s`, where `a` is the current vertex and `[a]` is the list of its predecessors. In case of `optimumPath` your state `s` is the resulting map `Maybe (Map a e)` which switches from `Nothing` to an initial map as soon as you encounter the source path vertex.

src/Algebra/Graph/Acyclic/Labelled/Algorithm.hs
```diff
@@ -0,0 +1,35 @@
...
+-- @
+-- dijkstra ('LAM.toAcyclicOrd' $ 'LAM.edges' [(2, 'b', 'c'), (1, 'a', 'b'), (3, 'a', 'c')]) 'z' == Map.'Map.fromList' [('z', 0), ('a', Infinite), ('b', Infinite), ('c', Infinite)]
+-- dijkstra ('LAM.toAcyclicOrd' $ 'LAM.edges' [(2, 'b', 'c'), (1, 'a', 'b'), (3, 'a', 'c')]) 'a' == Map.'Map.fromList' [('a', 0), ('b', 1), ('c', 3)]
+-- @
+optimumPath :: (Semiring e, Ord a) => AdjacencyMap e a -> a -> Map a e
+optimumPath am s = foldl (relaxVertex em) (initialize em) vL
+  where
+    vL = dropWhile (/=s) $ topSort am
+    em = (LAM.adjacencyMap . fromAcyclic) am
+    initialize = Map.insert s one . Map.map (const zero)
```
snowleopard: If the `s` doesn't belong to the graph you still mark it with `one` in the result. I'm not sure this is the right behaviour.

adithyaov: I agree I should probably change the result to `Maybe (Map a e)` as suggested.

src/Algebra/Graph/Acyclic/Labelled/Algorithm.hs
adithyaov: Ah, I see, I am assuming the idempotent property. Do you suggest I change the implementation or restrict the function to `Dioid e`? I believe the later is more better as ideally one would want to use this function with data type `Optimum` which is a `Dioid`.

src/Algebra/Graph/Acyclic/Labelled/Algorithm.hs
adithyaov: Oh, My bad. Will change it right away.

adithyaov:
> @adithyaov A general comment: it looks to me that there is a more general function that can be factored out of your implementation. Something fold-like, but for an acyclic graph, where you start with an initial state s and then traverse the graph in the topological order, at each step modifying the state with a function like s -> a -> [a] -> s, where a is the current vertex and [a] is the list of its predecessors. In case of optimumPath your state s is the resulting map Maybe (Map a e) which switches from Nothing to an initial map as soon as you encounter the source path vertex.

Ah, I see. Having such a function in the library would be useful. I guess, I'll create something like this and implement `optimumPath` on top of that. I'll create it in `Acyclic.Labelled.AdjacencyMap` for now.

src/Algebra/Graph/Acyclic/Labelled/Algorithm.hs
snowleopard: I'm not suggesting to go with `Maybe (Map a e)` as the final result, but it might appear somewhere internally. I am just saying we should think carefully about what we return in the case `s` is not in the graph. Is `s -> one` what we want? Maybe `s -> zero`? Or perhaps even `Nothing` but that may be inconvenient.

snowleopard: Perhaps, the right thing to do is to return a map that doesn't have an entry for `s` in this case.

Do we guarantee that the resulting map contains an entry for each of the graph vertices?

src/Algebra/Graph/Acyclic/Labelled/Algorithm.hs
snowleopard: Yes, I think in this case you better switch to `Dioid e` in the constraint.

snowleopard:
> I'll create it in `Acyclic.Labelled.AdjacencyMap` for now.

I think it should go into `Algebra.Graph.Acyclic.Labelled.AdjacencyMap.Algorithm`. Oh, the module names are getting ridiculously long, we need to do something with this.

src/Algebra/Graph/Acyclic/Labelled/Algorithm.hs
```diff
@@ -0,0 +1,46 @@
...
+topSort :: (Ord a) => AdjacencyMap e a -> [a]
+topSort = fromMaybe [] . AM.topSort . LAM.skeleton . fromAcyclic
+
+-- TODO: Improve documentation for 'foldAcyclic'
+-- | Compute the final state by using the initial state and
+-- traversing the entire graph in topological order.
+foldAcyclic :: (Ord a) => (s -> a -> [a] -> s) -> s ->  AdjacencyMap e a -> s
+foldAcyclic f s = snd . foldl' f' ([], s) . topSort
+  where
+    f' (p, s) x = (x:p, f s x p)
```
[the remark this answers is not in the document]
snowleopard: I think you actually misunderstood what I meant by the list `[a]`: it's not a list of vertices that appear earlier in the topological sort, it's a list of direct predecessors of the vertex.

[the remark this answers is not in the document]
snowleopard: Also, let's call this function simply `fold`. The module name already has `Acyclic` in it.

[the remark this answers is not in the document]
adithyaov: > I think you actually misunderstood what I meant by the list [a]: it's not a list of vertices that appear earlier in the topological sort, it's a list of direct predecessors of the vertex.

Ah, I see. My bad, Will change it.

src/Algebra/Graph/Acyclic/Labelled/Algorithm.hs
adithyaov: > Do we guarantee that the resulting map contains an entry for each of the graph vertices?

Yes.

> Perhaps, the right thing to do is to return a map that doesn't have an entry for s in this case.

I think to return `Nothing` is more efficient (as I am using `fmap`) and meaningful in this case.

src/Algebra/Graph/Acyclic/Labelled/Algorithm.hs
adithyaov: Ah, yes, I think using a strict version of foldl would be better, considering that the library is made to work with huge graphs.

src/Algebra/Graph/Acyclic/Labelled/Algorithm.hs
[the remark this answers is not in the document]
adithyaov: @snowleopard Do you think something like the following would be better for labeled acyclic graphs.
`fold :: (Ord a) => (s -> a -> [(e, a)] -> s) -> s ->  AdjacencyMap e a -> s`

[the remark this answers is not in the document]
snowleopard: Yes, good point! We need the labels on the incoming edges too.

src/Algebra/Graph/Acyclic/Labelled/Algorithm.hs
snowleopard: If using `Nothing` is more efficient you can use in the implementation, but you don't necessarily need to use it in the interface too.

As for being meaningful, I think using a `const zero` map is more meaningful. Why? Because you are saying that all graph vertices are unreachable from the given vertex, which is correct. With `Maybe` you can use `Nothing` to mean the same thing, but then `Just zeroMap` will be an impossible and unused output (because if the source vertex belongs to the graph then it will have `one` associated with it). As a result, I think using `Maybe` doesn't bring any type-safety, it's merely a choice between using `Nothing` and `zeroMap` to represent the same thing. Using `Maybe` will probably be just less convenient in practice.

snowleopard:
@adithyaov Looking at the current implementation of `optimumPath` via `fold`, the latter doesn't look very convenient. Perhaps, we should switch to the following type signature?

```haskell
fold :: (Ord a) => (e -> a -> a -> s -> s) -> s -> AdjacencyMap e a -> s
```
Instead of folding a graph vertex-by-vertex, we go edge-by-edge (the `e -> a -> a` part) in the topological order. With this change, I think `optimumPath` will look much more natural. Could you try it?

I've also switched to `... -> s -> s` hopefully making the step function easier to apply to the state.

test/Algebra/Graph/Test/Arbitrary.hs
```diff
@@ -74,6 +75,20 @@ instance (Ord a, Arbitrary a) => Arbitrary (AAM.AdjacencyMap a) where
           let edges = AAM.edgeList g
           in [ AAM.removeEdge x y g | (x, y) <- edges ]
 
+-- An Arbitrary instance for Acyclic.Labelled.AdjacencyMap
+instance (Ord a, Arbitrary a, Eq e, Monoid e, Arbitrary e) => Arbitrary (ALAM.AdjacencyMap e a) where
+    arbitrary = ALAM.toAcyclicOrd <$> arbitrary
```
snowleopard: This is not a very good `Arbitrary` instance: it can only produce graphs where edges go from "lower" to "higher" vertices.

src/Algebra/Graph/Acyclic/Labelled/AdjacencyMap/Algorithm.hs
```diff
@@ -0,0 +1,47 @@
...
+-- | Compute the /shortest path/ to each vertex in the graph
+-- from a given source vertex.
+--
+-- The following examples assume that the edges are distances,
+-- ie. the edge 'Semiring' is 'Distance'.
+-- @
+-- optimumPath ('LAM.toAcyclicOrd' $ 'LAM.edges' [(2, 'b', 'c'), (1, 'a', 'b'), (3, 'a', 'c')]) 'z' == Map.'Map.fromList' [('a', distance infinite), ('b', distance infinite), ('c', distance infinite)]
+-- optimumPath ('LAM.toAcyclicOrd' $ 'LAM.edges' [(2, 'b', 'c'), (1, 'a', 'b'), (3, 'a', 'c')]) 'a' == Map.'Map.fromList' [('a', 0), ('b', 1), ('c', 3)]
+-- @
+optimumPath :: (Dioid e, Ord a) => AdjacencyMap e a -> a -> Map a e
```
snowleopard: Please add a couple of simpler examples: a single vertex, a single edge.

src/Algebra/Graph/Acyclic/Labelled/AdjacencyMap/Algorithm.hs
```diff
@@ -0,0 +1,47 @@
...
+-- TODO: Make 'topSort' more efficient
+-- TODO: Add examples and tests for 'topSort'
+topSort :: (Ord a) => AdjacencyMap e a -> [a]
+topSort = fromMaybe [] . AM.topSort . LAM.skeleton . fromAcyclic
+
+-- TODO: Improve documentation for 'fold'
+-- TODO: Add examples and tests for 'fold'
+-- | Compute the final state by using the initial state and
+-- traversing the entire graph in topological order.
+fold :: (Ord a) => (e -> a -> a -> s -> s) -> s ->  AdjacencyMap e a -> s
```
snowleopard: We need to explain what the first argument does, clarifying the order of arguments. Also, we need a few simple examples.

src/Algebra/Graph/Acyclic/Labelled/AdjacencyMap/Algorithm.hs
```diff
@@ -0,0 +1,47 @@
...
+fold f s am = foldl' f' s . concatMap unfold . topSort $ am
+  where
+    em = LAM.adjacencyMap . fromAcyclic $ am
+    unfold x = map (\(a, e) -> (e, x, a)) . Map.toList $ em ! x
```
snowleopard: I don't think the order of processing edges is correct here. For example, for graph `1 * 2 * 3 * 4` the order will be:

`[(1,2), (1,3), (1,4), (2,3), (2,4), (3,4)]`

which processes the edge `1 -> 4` before the topologically earlier `2 -> 3`.

I think the order should be:

`[(1,2), (1,3), (2,3), (1,4), (2,4), (2,4)]`

adithyaov: Ah, I think I interpreted it incorrectly. In my mind I had a notion of applying a function to all edges from (rather than to) a vertex in topological order. Will change it.

test/Algebra/Graph/Test/Arbitrary.hs
adithyaov: With the functions that we have now since we need to produce Acyclic Labelled graph, I think this is the best we can do. This is similar to the `Arbitrary` instance for `Acyclic.AdjacencyMap`.
Having functions like `shrink` will probably improve the `Arbitrary` instance.

src/Algebra/Graph/Acyclic/Labelled/AdjacencyMap/Algorithm.hs
adithyaov: @snowleopard I think we should change the signature of `fold` to something like `a -> [(e, a)] -> s -> s` as before. This is more general, as, with `e -> a -> a -> s -> s`, if there are no edges then we would fold over nothing.

snowleopard: @adithyaov Yes, good idea. Indeed, we'd like to have a meaningful `fold` for `vertices xs`.

I suppose in `a -> [(e, a)] -> s -> s` we will call the function for every vertex `a` in the topological order, passing all outgoing edges `[(e, a)]` to it.

test/Algebra/Graph/Test/Arbitrary.hs
snowleopard: Here is what we could do:
1) Generate a random `AdjacencyMap` via `Ord`, like you do.
2) Generate a random shuffle of the sequence `[1..n]`.
3) Use `AM.gmap` to remap all vertices according to the shuffle.

Alas, it looks like we can't make it using our current API, since we won't be able to apply `gmap`. 

What about adding `swapVertices :: a -> a -> AdjacencyMap a -> AdjacencyMap a` to the acyclic API? It seems to be quite safe, and then we could use it when computing a random vertex permutation :)

adithyaov: Yes, I agree. `swapVertices` will give more freedom to the graphs.

src/Algebra/Graph/Acyclic/Labelled/AdjacencyMap/Algorithm.hs
adithyaov: I don't think it would matter much but I would prefer `[(e, a)]` to be all the incoming edges to the vertex. It gives a more clear picture and it's cleaner to visualize the flow.

For example, If you think of the acyclic graph as a dependency graph it is more comfortable to talk about a resource and its dependencies rather than a resource and all the resources that depend on it.

snowleopard: Agreed, having a list of dependencies seems more natural. In this case I suggest swapping the order of arguments: `[(e, a)] -> a -> s -> s` -- in this way the first `a` refers to a source of an edge, and the second to its destination, just like in `edges` for example. Seems easier to remember.

src/Algebra/Graph/Acyclic/Labelled/AdjacencyMap/Algorithm.hs
```diff
@@ -0,0 +1,69 @@
...
+-- TODO: Replace this function with 'skeleton' from Acyclic.Labelled to remove the use of fromMaybe
+-- TODO: Make 'topSort' more efficient
+-- TODO: Add examples and tests for 'topSort'
+topSort :: (Ord a) => AdjacencyMap e a -> [a]
+topSort = fromMaybe [] . AM.topSort . LAM.skeleton . fromAcyclic
+
+-- TODO: Improve documentation for 'fold'
+-- TODO: Make 'fold' more efficient
+-- TODO: Add tests for fold
+-- | fold takes any function with the signature @e -> a -> a -> s -> s@.
```
snowleopard: This doesn't match the type signature.

src/Algebra/Graph/Acyclic/Labelled/AdjacencyMap/Algorithm.hs
```diff
@@ -0,0 +1,69 @@
...
+-- This function folds over an acyclic graph, modifying an original
+-- input state. If one assumes the acyclic graph as a dependency
+-- graph and the vertices as resources, then, the expected function
+-- for fold takes all the dependents of the resource, the resource
+-- and the input state in that order and pproduces an output state.
```
snowleopard: Typo

src/Algebra/Graph/Acyclic/Labelled/AdjacencyMap/Algorithm.hs
```diff
@@ -0,0 +1,69 @@
...
+-- @
+fold :: (Ord a) => ([(e, a)] -> a -> s -> s) -> s ->  AdjacencyMap e a -> s
+fold f s wam = snd . foldl' process (nm, s) . topSort $ wam
+  where
+    am = LAM.adjacencyMap (fromAcyclic wam)
+    nm = Map.map (const []) am
+    addP v1 m =
+      let adjust v2 e = Map.adjust ((e, v1):) v2
+      in Map.foldrWithKey adjust m (am ! v1)
+    process (m, s) v2 = (addP v2 m, f (m ! v2) v2 s)
```
snowleopard: The tuple seems unnecessary, let's remove it.

snowleopard:
I've added some more comments.

@adithyaov There are a few `TODO`s in the code: do you plan to address any of them in this PR?

snowleopard:
Also, the CI fails with a compile error.

src/Algebra/Graph/Acyclic/Labelled/AdjacencyMap/Algorithm.hs
adithyaov: Yes, my bad. Will change it.

adithyaov:
@snowleopard I plan to address most of the TODOs before the merge. I apologize for the delay.

snowleopard:
@adithyaov Thanks, and don't worry about the delay.

adithyaov:
@snowleopard That is an interesting failure. I think it's because I changed the arbitrary instance.

adithyaov:
Please don't review this PR yet, I'll include a small writeup with all the major changes in this PR. This would make reviewing easier. I'll probably squash a few commits as well.

PR #220, opened 2019-07-10
 Implement Algebra.Graph.Undirected 

bolt12:
I was based on the 'Graph.hs' file to get the most of what was done on the Graph data type. I kept some of the rewrite rules but will remove them if they do not make sense in this context.

This draft PR compiles fine. I'm looking for some reviews on what I've done so far.

There are a couple of things not working according to the `SymmetricRelation` API, more specifically the functions that involve transforming the graph to an `AdjacencyMap`. I think this should be fixed by implementing an `toUndirectedAdjacencyMap`! The problem is the original `toAdjacencyMap` is built with respect to `AdjacencyMap` which `connect` does not commute.

How should I deal with this case? Implement an entire interface called `UndirectedAdjacencyMap` or implement a version of the graph construction primitive in the `AdjacencyMap` module that preserve the 'Undirected' rules?


algebraic-graphs.cabal
```diff
@@ -66,6 +66,7 @@ source-repository head
 library
     hs-source-dirs:     src
     exposed-modules:    Algebra.Graph,
+                        Algebra.Graph.Undirected.UndirectedGraph,
```
snowleopard: This should be just `Algebra.Graph.Undirected`.

snowleopard: This should be just `Algebra.Graph.Undirected`.

src/Algebra/Graph/Undirected/UndirectedGraph.hs
```diff
@@ -0,0 +1,935 @@
...
+-- of 'connect'.
+-- For graphs that are known to be /non-empty/ at compile time, see
+-- "Algebra.Graph.NonEmpty". 'Graph' is an instance of type classes defined in
+-- modules "Algebra.Graph.Class" and "Algebra.Graph.HigherKinded.Class", which
+-- can be used for polymorphic graph construction and manipulation.
+--
+-----------------------------------------------------------------------------
+module Algebra.Graph.Undirected.UndirectedGraph (
+    -- * Algebraic data type for graphs
+    UndirectedGraph (..),
```
snowleopard: Let's call this data type simply `Graph`. Users will use qualified modules to disambiguate.

snowleopard: Also, let's not export the constructor and see how an abstract implementation will work.

src/Algebra/Graph/Undirected/UndirectedGraph.hs
```diff
@@ -0,0 +1,935 @@
...
+@'isSubgraphOf' x y ==> x <= y@
+
+@'empty' <= x
+x     <= x + y
+x + y <= x * y@
+-}
+newtype UndirectedGraph a = UndirectedGraph { graph :: G.Graph a }
+             deriving (Generic, NFData)
+
+instance Show a => Show (UndirectedGraph a) where
```
snowleopard: Let's implement the `Show` instance using your implementation in `Symmetric.Relation`.

src/Algebra/Graph/Undirected/UndirectedGraph.hs
```diff
@@ -0,0 +1,935 @@
...
+  show = show . graph
+
+{- Note [Functions for rewrite rules]
```
snowleopard: Instead of copying this whole note, better say "See Note [Functions for rewrite rules] in Algebra.Graph.".

src/Algebra/Graph/Undirected/UndirectedGraph.hs
```diff
@@ -0,0 +1,935 @@
...
+--
+-- @
+-- 'isEmpty'     empty == True
+-- 'hasVertex' x empty == False
+-- 'vertexCount' empty == 0
+-- 'edgeCount'   empty == 0
+-- 'size'        empty == 1
+-- @
+empty :: UndirectedGraph a
+empty = UndirectedGraph G.Empty
```
snowleopard: Most of the functions in the API can be implemented using type-safe coercion, e.g. `empty = coerce G.empty`.

src/Algebra/Graph/Undirected/UndirectedGraph.hs
```diff
@@ -0,0 +1,935 @@
...
+--
+-- @
+-- 'isEmpty'     (vertex x) == False
+-- 'hasVertex' x (vertex x) == True
+-- 'vertexCount' (vertex x) == 1
+-- 'edgeCount'   (vertex x) == 0
+-- 'size'        (vertex x) == 1
+-- @
+vertex :: a -> UndirectedGraph a
+vertex = UndirectedGraph . G.Vertex
```
snowleopard: Same here: `vertex = coerce G.vertex`.

snowleopard:
@bolt12 Many thanks for the PR! I left a few comments.

Regarding `toAdjacencyMap`: Right now we do not have `Algebra.Graph.AdjacencyMap.Undirected`, but we do have `Algebra.Graph.Relation.Symmetric` -- let's use it instead, perhaps, adding a TODO that we should really switch to `Algebra.Graph.AdjacencyMap.Undirected` at some point in future (because it's typically faster and more convenient for algorithms).

src/Algebra/Graph/Undirected/UndirectedGraph.hs
bolt12: Got it! :)

bolt12: How can I resolve this error?

image

src/Algebra/Graph/Undirected/UndirectedGraph.hs
bolt12: okay!

algebraic-graphs.cabal
bolt12: Got it!

bolt12:
@snowleopard That's a really good idea! I've already resolved some of the points you mentioned. I left some more TODO's regarding the 'Algebra.Graph.AdjacencyMap.Undirected' but maybe I'll open an issue about this to register it.

Thanks for your reviews! I had problems on the coercion of the `empty` function and I still have problems when running ghci on the project root, but when i run `ghci src/Algebra/Graph/Undirected.hs` it compiles the module just fine!

image

Should I open an issue about this too?

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,907 @@
+{-# LANGUAGE DeriveGeneric, RankNTypes #-}
+-----------------------------------------------------------------------------
+-- |
+-- Module     : Algebra.Graph.Undirected.UndirectedGraph
```
snowleopard: Still old name.

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,907 @@
...
+
+Here are a few examples:
+
+@'vertex' 1 < 'vertex' 2
+'vertex' 3 < 'edge' 1 2
+'vertex' 1 < 'edge' 1 1
+'edge' 1 1 < 'edge' 1 2
+'edge' 1 2 < 'edge' 1 1 + 'edge' 2 2
+'edge' 1 2 < 'edge' 1 3@
+'edge' 1 2 == 'edge' 2 1
```
snowleopard: Misplaced `@`.

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,907 @@
...
+Note that the resulting order refines the 'isSubgraphOf' relation and is
+compatible with 'overlay' and 'connect' operations:
+
+@'isSubgraphOf' x y ==> x <= y@
+
+@'empty' <= x
+x     <= x + y
+x + y <= x * y@
+-}
+newtype Graph a = UndirectedGraph { graph :: G.Graph a }
```
snowleopard: We don't want to export the constructor, so let's give it a short name `UG`. Also, I don't think we need `graph`.

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,907 @@
...
+
+instance Ord a => Eq (Graph a) where
+    (==) = eqR
+
+instance Ord a => Ord (Graph a) where
+    compare = ordR
+
+-- TODO: Find a more efficient equality check.
+-- TODO: Implement toUndirectedAdjacencyMap.
+-- Check if two graphs are equal by converting them to their adjacency maps.
```
snowleopard: Outdated comment.

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,907 @@
...
+eqR :: Ord a => Graph a -> Graph a -> Bool
+eqR x y = toSymmetricRelation x == toSymmetricRelation y
+{-# NOINLINE [1] eqR #-}
+{-# RULES "eqR/Int" eqR = eqIntR #-}
+
+-- Like 'eqR' but specialised for graphs with vertices of type 'Int'.
+eqIntR :: Graph Int -> Graph Int -> Bool
+eqIntR x y = toSymmetricRelation x == toSymmetricRelation y
```
snowleopard: At the moment this function is not specialised at all, since we don't have have `Relation` specialised to vertices of type `Int`. Let's keep the code as is, but add a `TODO`, saying that this needs to be fixed in future.

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,907 @@
...
+-- TODO: Implement toUndirectedAdjacencyIntMap.
+-- Compare two graphs by converting them to their adjacency maps.
+ordR :: Ord a => Graph a -> Graph a -> Ordering
+ordR x y = compare (toSymmetricRelation x) (toSymmetricRelation y)
+{-# NOINLINE [1] ordR #-}
+{-# RULES "ordR/Int" ordR = ordIntR #-}
+
+-- Like 'ordR' but specialised for graphs with vertices of type 'Int'.
+ordIntR :: Graph Int -> Graph Int -> Ordering
+ordIntR x y = compare (toSymmetricRelation x) (toSymmetricRelation y)
```
snowleopard: Same here: this is not really specialised.

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,907 @@
...
+--
+-- @
+-- 'isEmpty'     empty == True
+-- 'hasVertex' x empty == False
+-- 'vertexCount' empty == 0
+-- 'edgeCount'   empty == 0
+-- 'size'        empty == 1
+-- @
+empty :: Graph a
+empty = UndirectedGraph G.empty
```
snowleopard: So, you didn't manage to write `coerce G.empty` here?

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,907 @@
...
+--
+-- @
+-- 'isEmpty'     (vertex x) == False
+-- 'hasVertex' x (vertex x) == True
+-- 'vertexCount' (vertex x) == 1
+-- 'edgeCount'   (vertex x) == 0
+-- 'size'        (vertex x) == 1
+-- @
+vertex :: a -> Graph a
+vertex = coerce . G.vertex
```
snowleopard: I think it should be `coerce G.vertex` -- does it work?

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,907 @@
...
+-- edge x y               == 'connect' ('vertex' x) ('vertex' y)
+-- edge x y               == 'edge' y x
+-- edge x y               == 'edges' [(x,y), (y,x)]
+-- 'hasEdge' x y (edge x y) == True
+-- 'edgeCount'   (edge x y) == 1
+-- 'vertexCount' (edge 1 1) == 1
+-- 'vertexCount' (edge 1 2) == 2
+-- @
+edge :: a -> a -> Graph a
+edge x y = connect (vertex x) (vertex y)
```
snowleopard: And this should be `edge = coerce G.edge`, etc.

snowleopard:
@bolt12 Thanks! I've left some more comments.

Not sure about the `Data.List.Extra` error.

By the way, could you please copy the text of errors instead of using screenshots? Screenshots are sometimes difficult to view, e.g. from a mobile.

src/Algebra/Graph/Undirected.hs
bolt12: Yes, I left it there and meant to leave a note saying that it wasn't really specialized, but forgot! I will fix that!

src/Algebra/Graph/Undirected.hs
bolt12: Yes, it works fine, GHC actually gave me a type error with `coerce G.vertex`, that's why I changed it.

src/Algebra/Graph/Undirected.hs
bolt12: No, GHC gave me the error I posted... It's something with the quantified variables, I don't know if it needs some extension enabled.. but since I'm having problems with the `Data.List.Extra` maybe it's my setup that's not right. Could you try for yourself and check if `coerce G.empty` works fine?

bolt12:
@snowleopard Sure!

```Haskell
[28 of 45] Compiling Algebra.Graph.Test ( test/Algebra/Graph/Test.hs, interpreted )

test/Algebra/Graph/Test.hs:24:1: error:
    Could not find module ‘Data.List.Extra’
    Use -v to see a list of the files searched for.
   |
24 | import Data.List.Extra (nubOrd)
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Failed, 27 modules loaded.
/bin/sh: cls: command not found
```

As I said, maybe my setup has some kind of misconfiguration... if I try to load a specific module it compiles just fine!

This is the `coerce G.empty` error:

```Haskell
src/Algebra/Graph/Undirected.hs:263:9: error:
    • Couldn't match representation of type ‘a0’ with that of ‘a’
        arising from a use of ‘coerce’
      ‘a’ is a rigid type variable bound by
        the type signature for:
          empty :: forall a. Graph a
        at src/Algebra/Graph/Undirected.hs:262:1-16
    • In the expression: coerce G.empty
      In an equation for ‘empty’: empty = coerce G.empty
    • Relevant bindings include
        empty :: Graph a (bound at src/Algebra/Graph/Undirected.hs:263:1)
    |
263 | empty = coerce G.empty
    |         ^^^^^^^^^^^^^^
```

snowleopard:
I'm not entirely sure what's going on with `coerce` in this case but, apparently, GHC just needs some help with type inference. For example, the following compiles:

```haskell
empty :: Graph a
empty = (coerce :: G.Graph a -> Graph a) G.empty

vertex :: a -> Graph a
vertex = (coerce :: (a -> G.Graph a) -> (a -> Graph a)) G.vertex
```

To make the code more readable, I suggest we add a couple of helper functions:

```haskell
coerce0 :: G.Graph a -> Graph a 
coerce0 = coerce

coerce1 :: (a -> G.Graph b) -> (a -> Graph b)
coerce1 = coerce

empty :: Graph a
empty = coerce0 G.empty

vertex :: a -> Graph a
vertex = coerce1 G.vertex
```

src/Algebra/Graph/Undirected.hs
snowleopard: Sorry, apparently things are a bit more complicated than I thought. See https://github.com/snowleopard/alga/pull/220#issuecomment-510891564.

src/Algebra/Graph/Undirected.hs
bolt12: Got it!

src/Algebra/Graph/Undirected.hs
bolt12: Thanks for your help! It's all good now!

bolt12:
@snowleopard thanks for your help on the `coerce`problem!

I've fixed some of the reviews you made. There's this function that I'm not sure if can be optimized somehow: `hasEdge`. Should I just use the the help from the `SymmetricRelation` implementation?

I still need to write the test suite and catch some possible bugs.

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,926 @@
...
+-- of 'connect'.
+-- For graphs that are known to be /non-empty/ at compile time, see
+-- "Algebra.Graph.NonEmpty". 'Graph' is an instance of type classes defined in
+-- modules "Algebra.Graph.Class" and "Algebra.Graph.HigherKinded.Class", which
+-- can be used for polymorphic graph construction and manipulation.
+--
+-----------------------------------------------------------------------------
+module Algebra.Graph.Undirected (
+    -- * Basic graph construction primitives
+    empty, vertex, edge, overlay, connect, vertices, edges, overlays, connects,
```
snowleopard: You need to export the data type `Graph`, as otherwise, it's impossible to use the library :) Just don't export the constructor `UG`.

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,926 @@
...
+-- 'vertexCount' (overlay x y) >= 'vertexCount' x
+-- 'vertexCount' (overlay x y) <= 'vertexCount' x + 'vertexCount' y
+-- 'edgeCount'   (overlay x y) >= 'edgeCount' x
+-- 'edgeCount'   (overlay x y) <= 'edgeCount' x   + 'edgeCount' y
+-- 'size'        (overlay x y) == 'size' x        + 'size' y
+-- 'vertexCount' (overlay 1 2) == 2
+-- 'edgeCount'   (overlay 1 2) == 0
+-- @
+overlay :: Graph a -> Graph a -> Graph a
+overlay (UG g1) (UG g2) = coerce (G.Overlay g1 g2)
```
snowleopard: Should be just `coerce2 G.overlay` I think.

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,926 @@
...
+coerce0 = coerce
+
+-- Help GHC with type inference (direct use of 'coerce' does not
+-- compile).
+coerce1 :: (a -> G.Graph b) -> (a -> Graph b)
+coerce1 = coerce
+
+-- Help GHC with type inference (direct use of 'coerce' does not
+-- compile).
+coerce2 :: (a -> a -> G.Graph b) -> (a -> a -> Graph b)
```
snowleopard: You can make this type signature a bit more general: `(a -> b -> G.Graph c) -> (a -> b -> Graph c)`.

src/Algebra/Graph/Undirected.hs
snowleopard: Same for some other functions below.

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,926 @@
...
+-- | Construct the graph from a list of edges.
+-- Complexity: /O(L^2)/ time, /O(L)/ memory and size, where /L/ is the length of the
+-- given list.
+--
+-- @
+-- edges []          == 'empty'
+-- edges [(x,y)]     == 'edge' x y
+-- edges [(x,y), (y,x)] == 'edge' x y
+-- @
+edges :: Eq a => [(a, a)] -> Graph a
```
snowleopard: I don't think you need to do anything special here, just use `coerce1 G.edges`.

snowleopard: This whole module is mostly just one big `coerce` :)

You'll only need to be careful about deconstructing the graph -- you'll need to treat `Connect` as a commutative operation.

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,926 @@
...
+-- isSubgraphOf ('path' xs)     ('circuit' xs)  ==  True
+-- isSubgraphOf ('edge' x y)    ('edge' y x)    ==  True
+-- isSubgraphOf x y                         ==> x <= y
+-- @
+isSubgraphOf :: Ord a => Graph a -> Graph a -> Bool
+isSubgraphOf x y = SR.isSubgraphOf (toSymmetricRelation x) (toSymmetricRelation y)
+{-# NOINLINE [1] isSubgraphOf #-}
+{-# RULES "isSubgraphOf/Int" isSubgraphOf = isSubgraphOfIntR #-}
+
+-- Like 'isSubgraphOf' but specialised for graphs with vertices of type 'Int'.
```
snowleopard: This isn't currently specialised, so this comment is wrong. Please add a TODO to fix this.

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,926 @@
...
+-- @
+-- hasEdge x y 'empty'            == False
+-- hasEdge x y ('vertex' z)       == False
+-- hasEdge x y ('edge' x y)       == True
+-- hasEdge x y ('edge' y x)       == True
+-- hasEdge x y . 'removeEdge' x y == 'const' False
+-- hasEdge x y                  == 'elem' (min x y, max x y) . 'edgeList'
+-- @
+hasEdge :: Eq a => a -> a -> Graph a -> Bool
+hasEdge s t (UG g) = hit g == Edge || hit' g == Edge
```
snowleopard: I suggest to simplify this to `G.hasEdge s t g || G.hasEdge t s g` for now, and add a TODO to optimise this further.

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,926 @@
...
+-- edgeSet 'empty'      == Set.'Set.empty'
+-- edgeSet ('vertex' x) == Set.'Set.empty'
+-- edgeSet ('edge' x y) == Set.'Set.singleton' (min x y, max x y)
+-- @
+edgeSet :: Ord a => Graph a -> Set.Set (a, a)
+edgeSet = SR.edgeSet . toSymmetricRelation
+{-# INLINE [1] edgeSet #-}
+{-# RULES "edgeSet/Int" edgeSet = edgeIntSetR #-}
+
+-- Like 'edgeSet' but specialised for graphs with vertices of type 'Int'.
```
snowleopard: This isn't specialised.

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,926 @@
...
+-- TODO: This is a very inefficient implementation. Find a way to construct an
+-- symmetric relation directly, without building intermediate representations for all
+-- subgraphs.
+-- TODO: Change this implementation when 'UndirectedAdjacencyMap' is
+-- defined.
+-- Convert a graph to 'SR.Relation'.
+toSymmetricRelation :: Ord a => Graph a -> SR.Relation a
+toSymmetricRelation = foldg SR.empty SR.vertex SR.overlay SR.connect
+
+-- TODO: Change this to a specialised version when
```
snowleopard: Just delete this code for now.

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,926 @@
...
+-- given list.
+--
+-- @
+-- path []        == 'empty'
+-- path [x]       == 'vertex' x
+-- path [x,y]     == 'edge' x y
+-- path . 'reverse' == path 
+-- @
+path :: Eq a => [a] -> Graph a
+path xs = case xs of []     -> empty
```
snowleopard: Keep `coerce`ing!

snowleopard:
@bolt12 I left some more comments. Please don't stop with `coerce` -- almost all functions in this module are going to be just `coerce`! :)

src/Algebra/Graph/Undirected.hs
bolt12: Ups, my bad!

src/Algebra/Graph/Undirected.hs
bolt12: I thought so too but had to define and auxiliar function to help GHC:

```Haskell
-- Help GHC with type inference (direct use of 'coerce' does not
-- compile).
coerce3 :: (Coercible a b, Coercible c d) => (a -> c -> G.Graph e) -> (b -> d -> Graph e)
coerce3 = coerce
```

This one can be used instead of `coerce2`!

src/Algebra/Graph/Undirected.hs
bolt12: I see that now! I'm going to change everywhere I can!

bolt12: I'm following the approach where `Connect a b` means that 'a' is connected to 'b' and 'b' is connected to 'a'. I could write `connect` like this:
```Haskell
connect :: Graph a -> Graph a -> Graph a
connect = coerce2 G.connect . coerce2 (flip G.connect)
```
But I don't think this would be a good idea! But with this approach I would need to filter the duplicate edges on the ' edges' function. What would you prefer?

the `edges`function would be like this:

```Haskell
edges :: Eq a => [(a, a)] -> Graph a
edges = coerce1 G.edges . nubBy (\x y -> x == swap y)
```

`nubBy` can be optimized!

src/Algebra/Graph/Undirected.hs
snowleopard: Aha, I see -- perhaps, you should just make `coerce2` more general using your implementation. By `coerceN` above I meant to name specialised `coerce` functions with N arguments.

src/Algebra/Graph/Undirected.hs
bolt12: Yes, you're right! I forgot this ones too!

src/Algebra/Graph/Undirected.hs
snowleopard: > But with this approach I would need to filter the duplicate edges on the `edges` function.

Why? I don't understand. You don't need to filter out anything.

src/Algebra/Graph/Undirected.hs
bolt12: Sounds good!

src/Algebra/Graph/Undirected.hs
bolt12: Fixed!

src/Algebra/Graph/Undirected.hs
bolt12: Alright!

src/Algebra/Graph/Undirected.hs
bolt12: `edges` in the Graph module is defined like this `edges = overlays . map (uncurry edge)` where `edge` is just `connect (vertex x) (vertex y)` if, for the `Undirected` module I use the same implementation of `edges` then for `edges [(1,2),(2,1)] == Overlay (Connect (Vertex 1) (Vertex 2)) (Connect (Vertex 2) (Vertex 1))`, right?

If we don't want to represent the redundant edges in the Graph structure itself I think we need to filter the redundant edges (that's what I meant with duplicate)

src/Algebra/Graph/Undirected.hs
bolt12: Right now I have 5 coerce type signatures, should I put them somewhere else or can I keep them in the file as of now?

bolt12:
@snowleopard here (line 594 - 609)
```Haskell
-- | The sorted list of vertices of a given graph.
-- Complexity: /O(s * log(n))/ time and /O(n)/ memory.
--
-- @
-- vertexList 'empty'      == []
-- vertexList ('vertex' x) == [x]
-- vertexList . 'vertices' == 'Data.List.nub' . 'Data.List.sort'
-- @
{-# INLINE [1] vertexList #-}
{-# RULES "vertexList/Int" vertexList = vertexIntListR #-}

-- Like 'vertexList' but specialised for graphs with vertices of type 'Int'.
vertexIntListR :: Graph Int -> [Int]
vertexIntListR = IntSet.toList . vertexIntSetR
```

If I use `coerce G.vertexList` do I need to keep the specialised version?

snowleopard:
@bolt12 Yes, keep `Int`-specialised versions, especially if we can properly specialise them!

src/Algebra/Graph/Undirected.hs
snowleopard: Keep them here for now.

bolt12:
@snowleopard Okay! I also realized that some of the specializations can benefit from the 'AdjacencyMap' module if we do not represent the redundant edges on the graph structure!

src/Algebra/Graph/Undirected.hs
snowleopard: > If we don't want to represent the redundant edges in the Graph structure itself

This was never our goal. It would be very hard to guarantee that we never have redundant vertices or edges in an expression. If you look at `Algebra.Graph`, for example, users can also call `edges` with repeated edges, but we never filter them out -- it would complicate the implementation.

bolt12: You are right it would be very hard! I will change my mindset! I was doing this thinking that we would never have redundant edges and that's not the case! Thanks for snapping me out of it

bolt12: With this in mind I think it will be a little hard to keep the API exactly the same as the `SymmetricRelation` one.. for example: `edges [(x,y), (y,x)] == 'edge' x y` this property does not hold.

Should I keep the semantics of the undirected graph and relax some of the properties that hold with `SymmetricRelation`?

snowleopard: > EDIT: Well it actually holds because the Show,Eq, etc.. instances convert it to a symmetric relation.

Exactly. We never expose the internal implementation, which allows us to have some redundant parts in expressions.

snowleopard: >  Should this property hold: edgeCount (edges [(1,2),(2,1)]) == 1 ?

Yes, of course, there is only one undirected edge in this graph.

bolt12: Okay! Thanks for your help!

bolt12:
@snowleopard Thank you very much for your quick reviews! I've pretty much coerced everything that made sense and fixed the `edges` function to respect the properties of an undirected graph! I've cleaned up the code too!

src/Algebra/Graph/Undirected.hs
bolt12: Okay!

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,870 @@
...
+-- __Alga__ is a library for algebraic construction and manipulation of graphs
+-- in Haskell. See <https://github.com/snowleopard/alga-paper this paper> for the
+-- motivation behind the library, the underlying theory, and implementation details.
+--
+-- This module defines and abstraction over the core data type 'Graph' and associated algorithms.
+-- Undirected graphs satisfy all laws of the
+-- 'Algebra.Graph.Class.Undirected' type class, including the commutativity
+-- of 'connect'.
+-- For graphs that are known to be /non-empty/ at compile time, see
+-- "Algebra.Graph.NonEmpty". 'Graph' is an instance of type classes defined in
```
snowleopard: The comment about non-empty graphs is out-of-date unless you plan to implement non-empty undirected graphs too. I suggest to just delete it for now.

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,870 @@
...
+-- modules "Algebra.Graph.Class" and "Algebra.Graph.HigherKinded.Class", which
```
snowleopard: These instances are currently missing.

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,870 @@
...
+-- Maintainer : andrey.mokhov@gmail.com
+-- Stability  : experimental
+--
+-- __Alga__ is a library for algebraic construction and manipulation of graphs
+-- in Haskell. See <https://github.com/snowleopard/alga-paper this paper> for the
+-- motivation behind the library, the underlying theory, and implementation details.
+--
+-- This module defines and abstraction over the core data type 'Graph' and associated algorithms.
+-- Undirected graphs satisfy all laws of the
+-- 'Algebra.Graph.Class.Undirected' type class, including the commutativity
```
snowleopard: Why not add this instance?

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,870 @@
...
+coerce1 = coerce
+
+-- Help GHC with type inference (direct use of 'coerce' does not
+-- compile).
+coerce2 :: (Coercible a b, Coercible c d) => (a -> c -> G.Graph e) -> (b -> d -> Graph e)
+coerce2 = coerce
+
+-- Help GHC with type inference (direct use of 'coerce' does not
+-- compile).
+coerce3 :: (b -> (a -> b) -> (b -> b -> b) -> (b -> b -> b) -> G.Graph a -> b)
```
snowleopard: Let's call this `coerceFold`.

snowleopard: Actually, since you use it only once, I think it shouldn't be a separate function. Just add a type signature locally.

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,870 @@
...
+-- | Connect a given list of graphs.
+-- Complexity: /O(L)/ time and memory, and /O(S)/ size, where /L/ is the length
+-- of the given list, and /S/ is the sum of sizes of the graphs in the list.
+--
+-- @
+-- connects []        == 'empty'
+-- connects [x]       == x
+-- connects [x,y]     == 'connect' x y
+-- connects           == 'foldr' 'connect' 'empty'
+-- 'isEmpty' . connects == 'all' 'isEmpty'
```
snowleopard: Please migrate all extra examples you have in `Symmetric.Relation` to this module, together with the associated tests. For example, in this case we have this extra example: `connects == connects . reverse`.

src/Algebra/Graph/Undirected.hs
bolt12: Yes, my bad!

src/Algebra/Graph/Undirected.hs
bolt12: I plan on implementing them and the `Undirected` instance too!

src/Algebra/Graph/Undirected.hs
bolt12: Yes, I am going to do that!

src/Algebra/Graph/Undirected.hs
bolt12: Okay!

src/Algebra/Graph/Undirected.hs
bolt12: It must have failed me to migrate some examples. I'll pay attention to these examples.

src/Algebra/Graph/Undirected.hs
bolt12: In what consists implementing the type classes defined in 'Algebra.Graph.HigherKinded.Class' ?

bolt12:
@snowleopard I've added some functions that were missing from the Symmetric Relation API and implemented the missing type classes with the exception of the HigherKinded because I didn't understand well it's objective. Could you please explain it to me!

Once again, thanks for your reviews!

bolt12:
@snowleopard I started to add the test-suite for this new module. I needed to add the `toGraph` projection for the 'UndirectedGraph' type and export it since we do not export the constructor ( this was needed to get the test-suite working) as well as the `toSymmetricRelation` which I think it's no problem.

There are a couple of problems with the test-suite mainly because there's no UndirectedAdjacencyMap implementation and I need to work around it!

Please tell me what you think :)

snowleopard:
> with the exception of the HigherKinded because I didn't understand well it's objective. Could you please explain it to me!

@bolt12 No need to add an instance for higher-kinded graphs for now. I'm keen to remove this type class at some point anyway.

Could you clarify what exactly is the problem with the testsuite? Please give an example that doesn't work.

bolt12:
@snowleopard Sure!

Here is the error when I run the test suite:

```Haskell
AdjacencyMap

Test failure:
    toAdjacencyMap             == foldg empty vertex overlay connect

*** Failed! Falsified (after 3 tests):
edge 1 2

*** Exception: ExitFailure 1
```

As you can see from the `Undirected` `ToGraph` type class instance: `toAdjacencyMap = toAdjacencyMap . UG.toSymmetricRelation`, this way, although it is slow, we do not need an implementation of `UndirectedAdjacencyMap` (yet) because we convert the `Undirected` Graph into a symmetric relation! The problem with the test suite is that it's testing `toAdjacencyMap == foldg AM.empty AM.vertex AM.overlay AM.connect ` which is not true!

I'm reusing the same testsuite used for `SymmetricRelation` since they should pass all tests but these little detail will make me create a different testsuite just for these case where I'll check that `toAdjacencyMap == foldg AM.empty AM.vertex AM.overlay AM.connect . toSymmetricRelation` (and in the future `toAdjacencyMap == foldg UAM.empty UAM.vertex UAM.overlay UAM.connect`)


Does creating a different test function `testUndirectedToGrap` makes sense? should I work around it differently?

snowleopard:
@bolt12 Thanks, I see. The problem is that `foldg` expects the arguments to satisfy the axioms of undirected graphs, but `AM.connect` doesn't satisfy it -- it's not commutative.

Perhaps, we shouldn't implement `ToGraph` at all. Can you drop this instance?

bolt12:
@snowleopard Sure! Should I leave a TODO note to add the instance once a possible `UndirectedAdjacencyMap` is implemented?

snowleopard:
> Should I leave a TODO note to add the instance once a possible `UndirectedAdjacencyMap` is implemented?

No, no need for it. I think we'll actually need a new `ToGraph` type class for undirected graphs.

Please keep only the functions exported from the new module in the API.

bolt12:
@snowleopard Okay! I think everything is ready to change this PR to a normal one and see what are the problems with the CI checks! :)

test/Main.hs
```diff
@@ -27,17 +28,18 @@ main :: IO ()
 main = do
     selected <- getArgs
     let go current = when (null selected || current `elem` selected)
-    go "Acyclic.AdjacencyMap"   testAcyclicAdjacencyMap
```
snowleopard: Please bring back the dots.

snowleopard:
> Edit: It was just a merge conflict, I hope everything is fine now!

Not quite, I think you resolved the conflict incorrectly.

test/Main.hs
bolt12: Ups, I don't know how those disappeared! Everything is fixed now!

bolt12:
@snowleopard Yes, I'm sorry, my bad! Hopefully there's nothing broken anymore.

test/Algebra/Graph/Test/Undirected.hs
```diff
@@ -0,0 +1,44 @@
+-----------------------------------------------------------------------------
+-- |
+-- Module     : Algebra.Graph.Test.Undirected
+-- Copyright  : (c) Andrey Mokhov 2016-2019
+-- License    : MIT (see the file LICENSE)
+-- Maintainer : andrey.mokhov@gmail.com
+-- Stability  : experimental
+--
+-- Testsuite for "Algebra.Graph.Undirected" and polymorphic functions defined in
+-- "Algebra.Graph.HigherKinded.Class".
```
snowleopard: I'm not sure the "and polymorphic functions..." part should be here.

src/Algebra/Graph/Undirected.hs
snowleopard: No need to implement `Algebra.Graph.HigherKinded.Class`, so please remove it from the comment.

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,913 @@
...
+-- 'Algebra.Graph.Class.Undirected' type class, including the commutativity
+-- of 'connect'.
+-- 'Graph' is an instance of type classes defined in
+-- modules "Algebra.Graph.Class" and "Algebra.Graph.HigherKinded.Class", which
+-- can be used for polymorphic graph construction and manipulation.
+--
+-----------------------------------------------------------------------------
+module Algebra.Graph.Undirected (
+    -- * Algebraic data type for graphs
+    Graph, toGraph,
```
snowleopard: I think we should rename `toGraph` to `fromUndirected` as in `Algebra.Graph.Relation.Symmetric`.

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,913 @@
...
+import Control.DeepSeq
+import Control.Monad (MonadPlus (..))
+import Data.Maybe (fromMaybe)
+import Data.Coerce
+import GHC.Generics
+
+import qualified Algebra.Graph                 as G
+import Algebra.Graph.Internal 
+
+import qualified Algebra.Graph.Relation.Symmetric    as SR
```
snowleopard: Alignment.

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,913 @@
...
+{-| The Undirected 'Graph' data type is an abstraction over the 'Graph' data
+   type and provides the same graph construction
+primitives 'empty', 'vertex', 'overlay' and 'connect'. We define the same 'Num' 
+as 'Graph' instance as a convenient notation for working with graphs:
+
+    > 0           == Vertex 0
+    > 1 + 2       == Overlay (Vertex 1) (Vertex 2)
+    > 1 * 2       == Connect (Vertex 1) (Vertex 2)
+    > 1 + 2 * 3   == Overlay (Vertex 1) (Connect (Vertex 2) (Vertex 3))
+    > 1 * (2 + 3) == Connect (Vertex 1) (Overlay (Vertex 2) (Vertex 3))
```
snowleopard: The above list is no longer correct since constructors are not exported -- please use the expressions that are printed by the `Show` instance.

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,913 @@
...
+x     <= x + y
+x + y <= x * y@
+-}
+newtype Graph a = UG { toGraph :: G.Graph a }
+             deriving (Generic, NFData)
+
+instance (Show a, Ord a) => Show (Graph a) where
+  show = show . toSymmetricRelation
+
+{- See Note [Functions for rewrite rules] in 'Algebra.Graph' -}
```
snowleopard: @nobrakal Do you think we need to use any rewrite rules in this module? We simply wrap `Graph` into a `newtype` and most of the functions below are defined simply via `coerce`. I have a feeling that the rewrite rules from your module may be sufficient, in which case this module could be simplified by removing all rewrite rules. What are your thoughts?

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,913 @@
...
+-- @
+-- removeEdge x y ('edge' x y)       == 'vertices' [x,y]
+-- removeEdge x y . removeEdge x y == removeEdge x y
+-- removeEdge x y                  == removeEdge y x
+-- removeEdge x y . 'removeVertex' x == 'removeVertex' x
+-- removeEdge 1 1 (1 * 1 * 2 * 2)  == 1 * 2 * 2
+-- removeEdge 1 2 (1 * 1 * 2 * 2)  == 1 * 1 + 2 * 2
+-- @
+removeEdge :: Eq a => a -> a -> Graph a -> Graph a
+removeEdge s t = coerce (G.removeEdge s t . G.removeEdge t s)
```
snowleopard: What does `coerce` do in this case? I think it should be removed.

snowleopard:
@bolt12 I made another pass and left some minor comments. I think we're getting ready to merge, thank you!

src/Algebra/Graph/Undirected.hs
bolt12: Makes sense! I will change that :)

src/Algebra/Graph/Undirected.hs
bolt12: Yes! I forgot to change this particular text when I changed the show intance! Thanks for pointing it out!

src/Algebra/Graph/Undirected.hs
bolt12: `coerce` makes the `G.removeEdge s t . G.removeEdge t s` which is of type `G.Graph a -> G.Graph a`, to be of type `Graph a -> Graph a`, right?

I can change the function and not use `coerce` with explicit pattern matching!

bolt12: @snowleopard forgot to mention you in the comment above :-)

snowleopard: Thanks, I think I got it, let's keep it as is.

P.S.: No need to mention me, I get notifications either way :) It's just a very busy week.

bolt12: Thanks, and sorry I forgot it was a review! I didn't mean to be impatient! Take your time!

src/Algebra/Graph/Undirected.hs
nobrakal: Indeed, you do not need to use rules. Express `bindR` using `G.Graph`'s `bind`.
Also inline everything (or at least, remove (now) useless phase control for `bindR`, `foldg` `overlays`...), since we need to show to the compiler that fusion is possible !

snowleopard: @nobrakal Thanks! 

@bolt12 Could you follow Alexandre's suggestions?

bolt12: @snowleopard Sure! :-) Just to make sure: I need to remove the rules, rewrite bind, and inline everything, right? I didn't understand the remove (now) part very well @nobrakal

nobrakal: Sorry if my comment was not clear. We need to:

* Remove all rules from `Algebra.Graph.Undirected` concerning functions that are just synonym of functions from `Algebra.Graph` (since their rules are already living in `Algebra.Graph`).
* Remove phase control (the [X] part in `INLINE` pragma) for functions that are just synonym of functions from `Algebra.Graph`. They are not needed anymore.
* The goal is to use rules from `Algebra.Graph`, thus we need to express all possible functions using functions from `Algebra.Graph`. This apply particularly to `bindR`.

This should (I hope) make things work :)

bolt12: Yes, thank you very much!

bolt12:
@snowleopard, @nobrakal I think I did what I need to! Since I don't have experience with ghc inlining pragmas I dont know if I inlined something I shouldn't or if I removed the correct phase control... I could use some help to see if everything is alright!

Thanks for your help!

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,919 @@
...
+newtype Graph a = UG { fromUndirected :: G.Graph a }
+             deriving (Generic, NFData)
+
+instance (Show a, Ord a) => Show (Graph a) where
+  show = show . toSymmetricRelation
+
+{- See Note [Functions for rewrite rules] in 'Algebra.Graph' -}
+
+instance Functor Graph where
+    fmap = fmapR
```
nobrakal:
```suggestion
    fmap f = UG . fmap f . fromUndirected
    {-# INLINE fmap #-}
```

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,919 @@
...
+
+-- Like 'ordR' but specialised for graphs with vertices of type 'Int'.
+-- NOTE: This is not specialised to vertices of type 'Int'. But it's still
+-- here for when 'UndirectedAdjacencyIntMap' is implemented.
+ordIntR :: Graph Int -> Graph Int -> Ordering
+ordIntR x y = compare (toSymmetricRelation x) (toSymmetricRelation y)
+
+instance Applicative Graph where
+    pure  = vertex
+    (<*>) = apR
```
nobrakal:
```suggestion
    (<*>) = coerce3 (<*>)
    {-# INLINE (<*>) #-}
```

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,919 @@
...
+
+apR :: Graph (a -> b) -> Graph a -> Graph b
+apR f x = bindR f (<$> x)
+{-# INLINE apR #-}
+
+instance Monad Graph where
+    return = pure
+    (>>=)  = bindR
```
nobrakal:
```suggestion
    (>>=) g = UG . (>>=) (fromUndirected g) . coerce
    {-# INLINE (>>=) #-}
```

nobrakal:
Hi @bolt12, you get the idea :) I have added a couple of suggestions to use even more code from `Algebra.Graph`.

This should do the trick, some tests will need to be made after #228 to ensure that all is well!

bolt12:
@nobrakal, Thanks for your suggestions! I've applied them and all the tests pass as of now, hope everything is alright when your PR is merged :-D

test/Algebra/Graph/Test/API.hs
```diff
@@ -292,6 +293,51 @@ graphAPI = API
     , compose                    = G.compose
     , box                        = G.box }
 
+-- | The API of 'UG.Graph.
```
snowleopard: Please check all generated docs: the closing quote is missing here.

test/Main.hs
```diff
@@ -33,6 +34,7 @@ main = do
     go "Bipartite.AdjacencyMap" testBipartiteAdjacencyMap
     go "Export"                 testExport
     go "Graph"                  testGraph
+    go "Undirected"             testUndirected
```
snowleopard: Please keep this list ordered alphabetically.

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,911 @@
...
+    * 'overlay' is commutative and associative:
+
+        >       x + y == y + x
+        > x + (y + z) == (x + y) + z
+
+    * 'connect' is associative, commutative and has 'empty' as the identity:
+
+        >   x * empty == x
+        >   empty * x == x
+        >   x * y     == y * x
```
snowleopard:
```suggestion
        >       x * y == y * x
```

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,911 @@
...
+-- TODO: Implement toUndirectedAdjacencyMap.
+-- Check if two graphs are equal by converting them to their symmetric
+-- relations.
+eqR :: Ord a => Graph a -> Graph a -> Bool
+eqR x y = toSymmetricRelation x == toSymmetricRelation y
+{-# NOINLINE [1] eqR #-}
+{-# RULES "eqR/Int" eqR = eqIntR #-}
+
+-- Like 'eqR' but specialised for graphs with vertices of type 'Int'.
+-- NOTE: This is not specialised to vertices of type 'Int'. But it's still
```
snowleopard:
```suggestion
-- TODO: This is not specialised to vertices of type 'Int'. But it's still
```

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,911 @@
...
+-- TODO: Find a more efficient comparison.
+-- TODO: Implement toUndirectedAdjacencyIntMap.
+-- Compare two graphs by converting them to their adjacency maps.
+ordR :: Ord a => Graph a -> Graph a -> Ordering
+ordR x y = compare (toSymmetricRelation x) (toSymmetricRelation y)
+{-# NOINLINE [1] ordR #-}
+{-# RULES "ordR/Int" ordR = ordIntR #-}
+
+-- Like 'ordR' but specialised for graphs with vertices of type 'Int'.
+-- NOTE: This is not specialised to vertices of type 'Int'. But it's still
```
snowleopard:
```suggestion
-- TODO: This is currently not specialised to vertices of type 'Int'. But it's still
```

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,911 @@
...
+-- isSubgraphOf ('edge' x y)    ('edge' y x)    ==  True
+-- isSubgraphOf x y                         ==> x <= y
+-- @
+isSubgraphOf :: Ord a => Graph a -> Graph a -> Bool
+isSubgraphOf x y = SR.isSubgraphOf (toSymmetricRelation x) (toSymmetricRelation y)
+{-# NOINLINE [1] isSubgraphOf #-}
+{-# RULES "isSubgraphOf/Int" isSubgraphOf = isSubgraphOfIntR #-}
+
+-- Like 'isSubgraphOf' but specialised for graphs with vertices of type 'Int'.
+-- NOTE: This is not specialised to vertices of type 'Int'. But it's still
```
snowleopard:
```suggestion
-- TODO: This is currently not specialised to vertices of type 'Int'. But it's still
```

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,911 @@
...
+-- vertexCount                   ==  'length' . 'vertexList'
+-- vertexCount x \< vertexCount y ==> x \< y
+-- @
+vertexCount :: Ord a => Graph a -> Int
+vertexCount = coerce3 G.vertexCount
+{-# INLINE [1] vertexCount #-}
+{-# RULES "vertexCount/Int" vertexCount = vertexIntCountR #-}
+
+-- Like 'vertexCount' but specialised for graphs with vertices of type 'Int'.
+-- NOTE: This is not specialised to vertices of type 'Int'. But it's still
```
snowleopard:
```suggestion
-- TODO: This is currently not specialised to vertices of type 'Int'. But it's still
```

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,911 @@
...
+-- edgeCount ('edge' x y) == 1
+-- edgeCount            == 'length' . 'edgeList'
+-- @
+edgeCount :: Ord a => Graph a -> Int
+edgeCount = length . edgeList
+{-# INLINE [1] edgeCount #-}
+{-# RULES "edgeCount/Int" edgeCount = edgeCountIntR #-}
+
+-- Like 'edgeCount' but specialised for graphs with vertices of type 'Int'.
+-- NOTE: This is not specialised to vertices of type 'Int'. But it's still
```
snowleopard:
```suggestion
-- TODO: This is currently not specialised to vertices of type 'Int'. But it's still
```

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,911 @@
...
+-- edgeList ('edge' x y)     == [(min x y, max y x)]
+-- edgeList ('star' 2 [3,1]) == [(1,2), (2,3)]
+-- @
+edgeList :: Ord a => Graph a -> [(a, a)]
+edgeList = SR.edgeList . toSymmetricRelation
+{-# INLINE [1] edgeList #-}
+{-# RULES "edgeList/Int" edgeList = edgeIntListR #-}
+
+-- Like 'edgeList' but specialised for graphs with vertices of type 'Int'.
+-- NOTE: This is not specialised to vertices of type 'Int'. But it's still
```
snowleopard:
```suggestion
-- TODO: This is currently not specialised to vertices of type 'Int'. But it's still
```

src/Algebra/Graph/Undirected.hs
```diff
@@ -0,0 +1,911 @@
...
+-- edgeSet ('vertex' x) == Set.'Set.empty'
+-- edgeSet ('edge' x y) == Set.'Set.singleton' ('min' x y, 'max' x y)
+-- @
+edgeSet :: Ord a => Graph a -> Set.Set (a, a)
+edgeSet = SR.edgeSet . toSymmetricRelation
+{-# INLINE [1] edgeSet #-}
+{-# RULES "edgeSet/Int" edgeSet = edgeIntSetR #-}
+
+-- Like 'edgeSet' but specialised for graphs with vertices of type 'Int'.
+-- NOTE: This is not specialised to vertices of type 'Int'. But it's still
```
snowleopard:
```suggestion
-- TODO: This is currently not specialised to vertices of type 'Int'. But it's still
```

snowleopard:
@bolt12 Apologies for taking so long to review and merge. I've added some more suggestions.

test/Algebra/Graph/Test/API.hs
bolt12: Thanks! I've checked all the generated docs and fixed some small issues like this, I hope it's all good now!

bolt12:
@snowleopard no problem at all, this month is a little busy for me too!

I resolved your reviews and gave a look at the generated docs. I think it's all good! Let me know if there's still anything I should correct!

Thanks for your reviews!

test/Main.hs
```diff
@@ -1,20 +1,21 @@
...
+import           Algebra.Graph.Test.Graph
+import           Algebra.Graph.Test.Internal
+import           Algebra.Graph.Test.Labelled.AdjacencyMap
+import           Algebra.Graph.Test.Labelled.Graph
+import           Algebra.Graph.Test.NonEmpty.AdjacencyMap
+import           Algebra.Graph.Test.NonEmpty.Graph
+import           Algebra.Graph.Test.Relation
+import           Algebra.Graph.Test.Relation.SymmetricRelation
+import           Algebra.Graph.Test.Undirected
+import           Data.Graph.Test.Typed
```
snowleopard: These changes seem accidental, please undo them. Same for the two imports below.

test/Algebra/Graph/Test/Undirected.hs
```diff
@@ -0,0 +1,43 @@
...
+t :: TestsuiteInt Graph
+t = fmap toIntAPI tPoly
+
+type G = Graph Int
+
+testUndirected :: IO ()
+testUndirected = do
+    putStrLn "\n============ Graph.Undirected ============"
+    test "Axioms of undirected graphs" $
+        size10 (undirectedAxioms :: GraphTestsuite G)
```
snowleopard: Compared to `SymmetricRelation`, we miss some tests:

https://github.com/snowleopard/alga/blob/master/test/Algebra/Graph/Test/Relation/SymmetricRelation.hs

For example, `fromUndirected` is not tested. I guess we need to test it. 

Do we need to add `toUndirected` too, for consistency of the API?

snowleopard:
@bolt12 Thanks, I've just left a couple of more comments :-)

test/Main.hs
bolt12: It is! I'm sorry about this, my Haskell setup was doing this automatically. I already disabled it!

test/Algebra/Graph/Test/Undirected.hs
bolt12: On it! I think it makes sense to provide `toUndirected` it wont hurt since it will be just a coercion, right?

bolt12: Actually, I don't think that `toUndirected` and `fromUndirected` should work as `toSymmetric` and  `fromSymmetric` since we decided to convert it to a normal `Graph` without redundant edges. Some of the properties will be different, is that okay?

snowleopard:
@bolt12 I think `fromUndirected` is currently wrong: it allows you to observe the difference of `edge 1 2` and `edge 2 1`. I believe we should make the symmetric closure before allowing to introspect the structure of an `Undirected` graph.

src/Algebra/Graph/Undirected.hs
```diff
@@ -179,7 +180,7 @@ instance (Show a, Ord a) => Show (Graph a) where
 {- See Note [Functions for rewrite rules] in 'Algebra.Graph' -}
 
 instance Functor Graph where
-    fmap f = UG . fmap f . fromUndirected
+    fmap f = UG . fmap f . _fromUndirected
```
snowleopard: I think we probably don't need `_fromUndirected` at all, and should use `coerce` here and in `>>=`.

bolt12: You're right!

snowleopard:
@bolt12 Many thanks for the PR and all the fixes. Merged now! 

It's very nice that we now have undirected algebraic graphs! I'll make a pass with minor revisions at some point but your implementation and tests look great already.

bolt12:
@snowleopard Once again, it was my pleasure! Thank you for your time and reviews, I hope everything is right on point!
I want to thank @nobrakal too, for the help!

Issue #175, opened 2019-03-04
Shortest path in labeled graph.


Josef-Vonasek:
The current API of labeled graphs does not seem to capture the notion of disconnected components. For example for `Graph Bool String`, both False / True and True / False can denote the lack of and the existence of an unlabelled edge, respectively, so algorithms like `shortestPath` would require extra parameter `noEdge : e -> Bool`.

snowleopard:
We expect the type of edge labels to be at least a monoid, and the identity element of this monoid corresponds to the non-edge. You can therefore implement `noEdge`:

```haskell
noEdge :: (Monoid e, Eq e) => e -> Bool
noEdge e = (e == mempty)
```

There is a recent talk that should clarify this:

https://skillsmatter.com/skillscasts/12361-labelled-algebraic-graphs

Admittedly, this needs to be documented much better. I haven't yet managed to write this up.

Josef-Vonasek:
I tried to watch the video but it cannot be played because of "privacy issues". 

Another question I have is how do you express a - b - c, because if connect behaves like join then in `Connect True a (Connect True b c)` a should be connected both to b and c.



5. 3. 2019 3:33 AM, 3:33 AM, Andrey Mokhov <notifications@github.com> napsal/a:

snowleopard:
> I tried to watch the video but it cannot be played because of "privacy issues".

@Josef-Vonasek Strange, I just tried and it worked fine for me. It does require you to login though, e.g. using the GitHub account.

> Another question I have is how do you express a - b - c, because if connect behaves like join then in `Connect True a (Connect True b c)` a should be connected both to b and c.

If we take the Boolean monoid with `mempty = False` and `(<>) = (||)`, then your `a - b - c` example can be represented as `Connect False (Connect True a b) (Connect True b c)`.

adithyaov:
> (maybe write a paper?)

@snowleopard Quite intriguing! Let me know if I can help :-)

snowleopard:
@adithyaov If your acyclic graphs project succeeds, it may lead to some publishable results. Going through existing literature and writing an overview of existing approaches would be a good start.

snowleopard:
Just adding a note about another issue with the documentation: right now it's not clear that in most cases we require the type of edges to be a commutative and idempotent monoid. For arbitrary monoids like `Sum` a few common properties will be violated, e.g. `x + x == x`, `isSubgraphOf x x == True`, etc. #221.

PR #221, opened 2019-07-17
Modified isSubGraph in Labelled.AdjacencyMap, minor error rectified

adithyaov:
Changed `Labelled.AdjacencyMap.isSubGraph`

Please note that I am unable to run the tests (QuickCheck) on my system due to a problem with WSL. So the only option for me to verify if the tests are passing is Travis CI.


snowleopard:
@adithyaov The following properties should be satisfied by `isSubgraphOf`:

```haskell
isSubgraphOf x (overlay   x y) == True
isSubgraphOf x (connect e x y) == True
isSubgraphOf y (connect e x y) == True
```
I don't think your current implementation satisfies these requirements.

snowleopard:
Also, if you are fixing a bug, please add a test that currently fails but passes with yours fix.

adithyaov:
@snowleopard 
```
isSubgraphOf x (overlay   x y) == True
isSubgraphOf x (connect e x y) == True
isSubgraphOf y (connect e x y) == True
```
Satisfying the properties above becomes really complex. Please note the following, say,
```
x = edge (Sum 2) 0 1
y = edge (Sum -2) 0 1
z = overlay x y = vertices [0, 1]
```
Justifying that `x` is a subgraph of `z` is quite dufficult. We would probably need to check different conditions depending on the Monoid (edge).
In case of `Sum`, `isSubgraphOf  x y = isSubMapOfBy (isSubMapOfBy (\_ _ -> True)) x y` as `Sum` can result in anything.
In case of `Capacity`, `isSubgraphOf  x y = isSubMapOfBy (isSubMapOfBy (\p q -> q <= p)) x y` as `Capacity` can only decrease.
In case of `Distance`, `isSubgraphOf  x y = isSubMapOfBy (isSubMapOfBy (\p q -> q >= p)) x y` as `Distance` can only increase.
I'll get back to this PR after some amount of thought.

snowleopard:
@adithyaov We typically want edge labels to come from a *dioid*:

http://hackage.haskell.org/package/algebraic-graphs-0.4/docs/Algebra-Graph-Label.html#t:Dioid

`Sum` is not a dioid, which is why you have issues here. Combining algebraic graphs and non-dioids is problematic, because many basic properties no longer hold, e.g. `x + x = x`.

Both `Distance` and `Capacity` are dioids, so there should be no issues with them.

adithyaov:
@snowleopard 
Ah, I see, In that case, I believe the previous example is correct. I think one should probably change `isSubgraphOf` to work only with dioids.

snowleopard:
> I think I got an error as I was checking it for a Sum Monoid.

Right, I see. Indeed, with `Sum` there will be a few issues. I added a note about this: https://github.com/snowleopard/alga/issues/175#issuecomment-513222730.

If the current implementation works for dioids, then I think there is no need to change anything for now, but we'll definitely need to improve documentation for edge-lebelled graphs, as discussed in #175.

adithyaov:
#175 Closing this PR for now. Please feel free to reopen this if required.

PR #222, opened 2019-07-18
Implement induce in terms of induceJust

Avasil:
@nobrakal 
I tried running benchmarks with [your script](https://github.com/nobrakal/benchAlgaPr/blob/master/compare.sh) but I might be doing something wrong and don't have results yet

```
[pgawrys@localhost alga]$ ./compare.sh Stack ./ QuickComparison b88a4a25cf09ac7e17fab3769ac283ab8b2d61bf False induce
Alga Commit ID: 17ae076511bb43a853b667d214c894fa086eefc0
HEAD is now at b88a4a2 Add Acyclic.Ord (#215)
AlgaOld Commit ID: b88a4a25cf09ac7e17fab3769ac283ab8b2d61bf

Install dependencies and build the benchmarking suite

Args: run -g ("Mesh",3) -g ("Clique",3) -g ("Circuit",3) -i -d QuickComparison -l Alga -l AlgaOld -n dff -n topSort  --only induce
~/alga/alga
```
I run it from `~alga/alga` where my fork is located. I also tried without last two parameters. Do you have any guess what can be wrong (I assume it should work for `stack` )

### DIFF
--- a/src/Algebra/Graph.hs
+++ b/src/Algebra/Graph.hs
@@ -990,7 +990,6 @@ transpose :: Graph a -> Graph a
 transpose = foldg Empty Vertex Overlay (flip Connect)
 {-# INLINE transpose #-}
 
--- TODO: Implement via 'induceJust' to reduce code duplication.
 -- | Construct the /induced subgraph/ of a given graph by removing the
 -- vertices that do not satisfy a given predicate.
 -- Complexity: /O(s)/ time, memory and size, assuming that the predicate takes
@@ -1004,11 +1003,7 @@ transpose = foldg Empty Vertex Overlay (flip Connect)
 -- 'isSubgraphOf' (induce p x) x == True
 -- @
 induce :: (a -> Bool) -> Graph a -> Graph a
-induce p = foldg Empty (\x -> if p x then Vertex x else Empty) (k Overlay) (k Connect)
-  where
-    k _ x     Empty = x -- Constant folding to get rid of Empty leaves
-    k _ Empty y     = y
-    k f x     y     = f x y
+induce p = induceJust . fmap (\x -> if p x then Just x else Nothing)
 {-# INLINE [1] induce #-}
 
 -- | Construct the /induced subgraph/ of a given graph by removing the vertices
@@ -1198,7 +1193,7 @@ this line: http://hackage.haskell.org/package/base/docs/src/GHC.Base.html#mapFB.
   build's unfolding) else we'd get an infinite loop in the rules. Hence the
   activation control below.
 
-* composeR and matchR are here to remember the original function after applying
+* composeR, matchR and maybeR are here to remember the original function after applying
   a "buildR/f" rule. These functions are higher-order functions and therefore
   benefit from inlining in the final phase.
 
@@ -1219,6 +1214,10 @@ matchR :: b -> (a -> b) -> (a -> Bool) -> a -> b
 matchR e v p = \x -> if p x then v x else e
 {-# INLINE [0] matchR #-}
 
+maybeR :: b -> (a -> b) -> Maybe a -> b 
+maybeR = maybe
+{-# INLINE [0] maybeR #-}
+
 -- These rules transform functions into their buildR equivalents.
 {-# RULES
 "buildR/bindR" forall f g.
@@ -1226,6 +1225,9 @@ matchR e v p = \x -> if p x then v x else e
 
 "buildR/induce" [~1] forall p g.
     induce p g = buildR (\e v o c -> foldg e (matchR e v p) o c g)
+
+"buildR/induceJust" [~1] forall g.
+    induceJust g = buildR (\e v o c -> foldg e (maybeR e v) o c g)
 
 "buildR/foldg(fc)" [~1] forall (f :: forall b. (b -> b -> b) -> (b -> b -> b)) g.
     foldg Empty Vertex Overlay (f Connect) g = buildR (\e v o c -> foldg e v o (f c) g)
@@ -1257,6 +1259,9 @@ matchR e v p = \x -> if p x then v x else e
 {-# RULES
 "graph/induce" [1] forall f.
     foldg Empty (matchR Empty Vertex f) Overlay Connect = induce f
+
+"graph/induceJust" [1]
+    foldg Empty (maybeR Empty Vertex) Overlay Connect = induceJust
  #-}
 
 -- 'Focus' on a specified subgraph.
--- a/src/Algebra/Graph/Labelled.hs
+++ b/src/Algebra/Graph/Labelled.hs
@@ -492,7 +492,6 @@ transpose = foldg empty vertex (fmap flip connect)
 emap :: (e -> f) -> Graph e a -> Graph f a
 emap f = foldg Empty Vertex (Connect . f)
 
--- TODO: Implement via 'induceJust' to reduce code duplication.
 -- | Construct the /induced subgraph/ of a given graph by removing the
 -- vertices that do not satisfy a given predicate.
 -- Complexity: /O(s)/ time, memory and size, assuming that the predicate takes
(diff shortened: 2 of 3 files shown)


nobrakal:
@Avasil  Of course! `induce` is not standard (among other Haskell graph libraries), so the bench suite is not made to benchmark it.

The whole list can be obtained by running `time list --benchs`.

Anyway, `removevertex` is made using `induce`, and the [Travis run](https://travis-ci.org/snowleopard/alga/jobs/560434081) show bad results:

```
removeVertex: 2.26 (bad)
```

Avasil:
@nobrakal 

> Of course! induce is not standard (among other Haskell graph libraries), so the bench suite is not made to benchmark it.
The whole list can be obtained by running time list --benchs.

Sorry, I'm not following. I don't get any output (event `removeVertex` and others) when I run this locally through script. At the beginning I get output similar to the one in CI but it stops on
```
Args: run -g ("Mesh",3) -g ("Clique",3) -g ("Circuit",3) -i -d QuickComparison -l Alga -l AlgaOld -n dff -n topSort  --only induce
~/alga/alga
```

I guess I can always do it "by hand" operating on github repo directly but I'd rather check other options first. :D

nobrakal:
Sorry I was not clear. You cannot benchmark `induce`. Remove it from the invocation of the script.

```
./compare.sh Stack ./ QuickComparison b88a4a25cf09ac7e17fab3769ac283ab8b2d61bf False removeVertex
``` 

Should works.

Avasil:
> Sorry I was not clear. You cannot benchmark `induce`. Remove it from the invocation of the script.
> 
> ```
> ./compare.sh Stack ./ QuickComparison b88a4a25cf09ac7e17fab3769ac283ab8b2d61bf False removeVertex
> ```
> 
> Should works.

Unfortunately it ends with following output:

```
[pgawrys@localhost alga]$ ./compare.sh Stack ./ QuickComparison b88a4a25cf09ac7e17fab3769ac283ab8b2d61bf False removeVertex
Alga Commit ID: 17ae076511bb43a853b667d214c894fa086eefc0
HEAD is now at b88a4a2 Add Acyclic.Ord (#215)
AlgaOld Commit ID: b88a4a25cf09ac7e17fab3769ac283ab8b2d61bf

Install dependencies and build the benchmarking suite

Args: run -g ("Mesh",3) -g ("Clique",3) -g ("Circuit",3) -i -d QuickComparison -l Alga -l AlgaOld -n dff -n topSort  --only removeVertex
~/alga/alga
[pgawrys@localhost alga]$ 
```

nobrakal:
Oops, my bad, my script was faulty with `Stack`, it should run fine now!

Avasil:
Indeed it works now @nobrakal, thanks a lot for help!

Avasil:
@nobrakal I'm trying to figure out how to do it without performance regression and one of my experiments is showing that even with implementation in terms of `induceJust` results are as good or better than on master if we disable rewrite rules / inlining:

```
addEdge: 1.10 (OK)
addVertex: 1.10 (bad)
dff: 1.00 (OK)
edgeCount: 1.00 (OK)
edgeList: 1.01 (OK)
edges: 1.06 (OK)
equality: 1.00 (OK)
hasEdge: 0.97 (OK)
hasVertex: 0.99 (OK)
isEmpty: 0.92 (OK)
reachable: 1.00 (OK)
removeEdge: 0.96 (OK)
removeVertex: 0.83 (good)
topSort: 1.01 (OK)
transpose: 1.04 (OK)
vertexCount: 0.94 (OK)
vertexList: 0.98 (OK)
``

**Comment by nobrakal** [comment]
@Avasil  I am not surprised that results are the same without rewrite rules: we use them only to optimize the  composition of functions working on graphs. Since `removeVertex` is just `induce`, composed with nothing, then its code should be the same without and with rewrite rules.

The `good` part is indeed strange, I will investigate ^^ (this might be a false positive).

**Comment by Avasil** [comment]
@nobrakal Sorry for flooding you with dummy questions but I'm using this as an opportunity to learn about these rules and I'm learning Haskell as I go. :D 

I'm trying to run benchmark with GHC options such as `-ddump-simpl-stats` but unfortunately without success so far. I modified your script directly to have something like:

```
stack build --ghc-options="-ddump-simpl-stats" "bench-graph:bench:time" --no-run-benchmarks --flag (...)
```
and commented `&> /dev/null` in case something should be printed there.

What's the right way to do it? Or anything else that could help me figure out why my original rule (from the first commit) didn't seem to work.

**Comment by nobrakal** [comment]
There is no problem at all :)

If you are trying to investigate, directly use https://github.com/haskell-perf/graphs and replace alga's sources by yours (ie, modify https://github.com/haskell-perf/graphs/blob/master/stack.yaml ). 

Then run directly `stack bench --ghc-options="-ddump-simpl-stats"`, it should work (or at least, will simplify the whole process).

I firstly misread what you implemented in this PR, please excuse me, I should have spotted something:

### There is an other problem

@snowleopard We have a little problem:

Using @Avasil definitions (without `maybeR`, but the problem is the same).

```Haskell
induce p g 
= induceJust $ fmap (\x -> if p x then Just x else Nothing) g
-- inlining
= buildR $ \e v o c -> foldg e (maybe e v) o c $ 
     buildR \e v o c -> foldg e (\x -> v $ if p x then Just x else Nothing) o c g
-- fusion
= buildR $ \e v o c -> foldg e (\x -> maybe e v $ if p x then Just x else Nothing) o c g
-- inlining of buildR
= foldg Empty (\x -> if p x then Vertex x else Empty) Overlay Connect g
```

After all the process, empty leaves are _not_ removed anymore (this is a problem, since one could expect this behavior). This is due to the `buildR` form of `induceJust` that is only partially valid (because there cannot be a `k` simplifier as in the classical definition of `induceJust`).

More I think about it, the more I don't like `induce`'s rules, since they are not totally valid. Maybe we should remove them, and have `induce` just as a good consumer.

snowleopard:
@nobrakal I am getting more and more inclined to remove the `Empty` constructor and define `Graph` as a `Maybe` of some non-empty underlying representation :) I think a few places in the library would just suddenly become simpler!

nobrakal:
Indeed, I like the idea :D

Actually, this what is hidden in `overlays` and `connects`, and it will solves the problem here.

snowleopard:
@Avasil Are you planning to do anything about the performance regression?

https://travis-ci.org/snowleopard/alga/jobs/567318797

See the `removeVertex1: 1.92 (bad)` entry. I don't think we can merge until this is fixed.

Avasil:
@snowleopard 
Initially, the replaced implementation didn't work well with rewrite rules. Current state is an experiment after removing them completely which improved situation but I'm not sure it is the way to go. It probably just needs more investigation into rewrite rules.

So I feel like we need to revisit this PR after https://github.com/snowleopard/alga/pull/228 is done or just close it for now

snowleopard:
@Avasil I see, thanks! Let's come back to this after #228.

snowleopard:
@Avasil #228 has been merged, so shall we revisit this PR?

Avasil:
> #228 has been merged, so shall we revisit this PR?

Sure, I will read about the changes there and see how it looks like now but unfortunately it will take me some time before I can focus my attention here

snowleopard:
@Avasil Thanks! Take you time, there is no rush.

snowleopard:
@Avasil I just realised that I completely forgot about this PR when writing #271. My apologies! 

I can see that this PR was attempting to do more than I did in #271. I'm unsure if that was necessary but I'm happy to discuss.

I'm also unsure what happened to our performance benchmarks; perhaps, they didn't survive our move to GitHub actions? I'll need to investigate this. I wonder if my PR introduced a performance regression that I missed. /cc @nobrakal

nobrakal:
Hi there, 

I am not familiar with GitHub Actions but the ci script https://github.com/snowleopard/alga/blob/master/.github/workflows/ci.yml does not make any reference to the regression suite.

snowleopard:
@nobrakal Yes, it looks like it's my fault. I'll need to restore it.

PR #223, opened 2019-07-20
Revise Acyclic.AdjacencyMap

snowleopard:
This PR includes:
* A lot of changes to comments, examples, implementation and tests.
* Renaming `disjointOverlay` to `union` and `disjointConnect` to `join`, which are commonly used terms.
* Adding some missing functions: `isSubgraphOf`, `preSet`, `postSet`.
* Removal of `PartialOrder`: I'm still hesitating about providing an unsafe construction method.
* Moving `Acyclic.Ord` to `Acyclic.AdjacencyMap.Ord`, and removing some functions from the API.

### DIFF
--- a/src/Algebra/Graph/Acyclic/AdjacencyMap/Ord.hs
+++ b/src/Algebra/Graph/Acyclic/AdjacencyMap/Ord.hs
@@ -1,6 +1,6 @@
 -----------------------------------------------------------------
 -- |
--- Module     : Algebra.Graph.Acyclic.Ord
+-- Module     : Algebra.Graph.Acyclic.AdjacencyMap.Ord
 -- License    : MIT (see the file LICENSE)
 -- Stability  : experimental
 --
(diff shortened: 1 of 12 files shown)


adithyaov:
> A lot of changes to comments, examples, implementation and tests.

I apologize, I believe I should have done these properly. Please let me know during the PR review whether the format of the tests is proper, I'll change them accordingly.

snowleopard:
> I apologize, I believe I should have done these properly.

No need to apologise! It's much easier to do a revision than to start from scratch :)

Thanks again for your earlier work, and I hope you agree with my changes. If not, please let me know!

snowleopard:
@adithyaov I think I'm finished with this revision. 

I didn't implement the `shrink` [idea](https://github.com/snowleopard/alga/wiki/Acyclic-graphs#using-scc-for-quick-and-safe-construction-of-acyclic-graphs) here, but I think we should do it at some point. If `shrink` works well then we could probably drop `Acyclic.AdjacencyMap.Ord` to keep the API small.

snowleopard:
@adithyaov I'd appreciate if you have a quick look at my changes. I'd like to merge as soon as possible to avoid blocking other PRs.

adithyaov:
@snowleopard I went over most of the files, it looks good to me. I did not cross-check the changes in the run time complexity yet. I'll go through this PR more thoroughly and get back to you by tomorrow.

snowleopard:
@adithyaov Thank you! I'll wait with merging until tomorrow.

> I did not cross-check the changes in the run time complexity yet.

I think I only fixed an error that I had in the documentation of `induce` where for some reason I didn't include the number of vertices `n` into the bound.

adithyaov:
@snowleopard I went through all the files. There seems to be a little inconsistency in the examples of `Acyclic.AdjacencyMap.Ord.edge` and the corresponding test cases. There seem to be a lot more test cases than examples.

Other than that everything seems to be fine.

src/Algebra/Graph/Acyclic/AdjacencyMap/Ord.hs
```diff
@@ -1,42 +1,42 @@
...
 
 -- | Construct the graph comprising /a single edge/.
 -- Complexity: /O(1)/ time, memory.
 --
 -- @
 -- edge x y               == 'connect' ('vertex' x) ('vertex' y)
+-- 'hasEdge' 1 1 (edge 1 1) == False
 -- 'hasEdge' 1 2 (edge 1 2) == True
 -- 'hasEdge' 2 1 (edge 2 1) == False
--- 'edgeCount'   (edge 1 2) == 1
```
adithyaov: The test cases for this exist even though the examples were removed.

snowleopard: Thanks, indeed, the examples and tests have got out of sync. Will fix this!

snowleopard:
@adithyaov Many thanks for having a look! I've fixed the inconsistency you found.

PR #224, opened 2019-07-21
improve dfs & top sort performance

jitwit:
I threw together the monadic tree unfold version.

I also tried doing top sort directly, since I think it could be advantageous to build the final list directly without going through a forest structure first. It should also allow an early exit if a back edge is detected.

### DIFF
--- a/src/Algebra/Graph/Acyclic/AdjacencyMap.hs
+++ b/src/Algebra/Graph/Acyclic/AdjacencyMap.hs
@@ -1,5 +1,3 @@
-{-# language ViewPatterns #-}
-
 -----------------------------------------------------------------------------
 -- |
 -- Module     : Algebra.Graph.Acyclic.AdjacencyMap
--- a/src/Algebra/Graph/Acyclic/AdjacencyMap.hs
+++ b/src/Algebra/Graph/Acyclic/AdjacencyMap.hs
@@ -52,14 +52,13 @@ module Algebra.Graph.Acyclic.AdjacencyMap (
     consistent
     ) where
 
-import Data.Maybe
+import Data.Either
 import Data.Set (Set)
 import Data.Coerce (coerce)
 
 import qualified Algebra.Graph.AdjacencyMap           as AM
 import qualified Algebra.Graph.AdjacencyMap.Algorithm as AM
 import qualified Algebra.Graph.NonEmpty.AdjacencyMap  as NAM
-import qualified Data.Graph.Typed                     as Typed
 import qualified Data.Map                             as Map
 import qualified Data.Set                             as Set
 
@@ -485,7 +484,8 @@ transitiveClosure = coerce AM.transitiveClosure
 -- topSort                       == 'fromJust' . 'AM.topSort' . 'fromAcyclic'
 -- @
 topSort :: Ord a => AdjacencyMap a -> [a]
-topSort (AAM am) = fromJust (AM.topSort am)
+topSort (AAM am) = fromRight impossible (AM.topSort am) where
+  impossible = error "topSort on Acyclic Graph returned cycle"
 
 -- | Compute the acyclic /condensation/ of a graph, where each vertex
 -- corresponds to a /strongly-connected component/ of the original graph. Note
--- a/src/Algebra/Graph/Acyclic/AdjacencyMap.hs
+++ b/src/Algebra/Graph/Acyclic/AdjacencyMap.hs
@@ -486,6 +486,7 @@ transitiveClosure = coerce AM.transitiveClosure
 -- @
 topSort :: Ord a => AdjacencyMap a -> [a]
 topSort (AM.topSort . coerce -> Right vs) = vs
+topSort _ = error "topSort on Acyclic Graph produced cycle"
 
 -- | Compute the acyclic /condensation/ of a graph, where each vertex
 -- corresponds to a /strongly-connected component/ of the original graph. Note
--- a/src/Algebra/Graph/Acyclic/AdjacencyMap.hs
+++ b/src/Algebra/Graph/Acyclic/AdjacencyMap.hs
@@ -52,6 +52,7 @@ module Algebra.Graph.Acyclic.AdjacencyMap (
     consistent
     ) where
 
+import Data.Maybe
 import Data.Set (Set)
 import Data.Coerce (coerce)
 
@@ -479,12 +480,12 @@ transitiveClosure = coerce AM.transitiveClosure
 -- @
 -- topSort 'empty'                 == []
 -- topSort ('vertex' x)            == [x]
--- topSort (1 * (2 + 4) + 3 * 4) == [3, 1, 4, 2]
+-- topSort (1 * (2 + 4) + 3 * 4) == [1, 2, 3, 4]
 -- topSort ('join' x y)            == 'fmap' 'Left' (topSort x) ++ 'fmap' 'Right' (topSort y)
--- topSort                       == 'Data.Maybe.fromJust' . topSort . 'fromAcyclic'
+-- topSort                       == 'fromJust' . 'AM.topSort' . 'fromAcyclic'
 -- @
-topSort :: (Ord a) => AdjacencyMap a -> [a]
-topSort (AAM am) = Typed.topSort (Typed.fromAdjacencyMap am)
+topSort :: Ord a => AdjacencyMap a -> [a]
+topSort (AAM am) = fromJust (AM.topSort am)
 
 -- | Compute the acyclic /condensation/ of a graph, where each vertex
 -- corresponds to a /strongly-connected component/ of the original graph. Note
--- a/src/Algebra/Graph/Acyclic/AdjacencyMap.hs
+++ b/src/Algebra/Graph/Acyclic/AdjacencyMap.hs
@@ -485,8 +485,9 @@ transitiveClosure = coerce AM.transitiveClosure
 -- 'Right' . topSort               == 'AM.topSort' . 'fromAcyclic'
 -- @
 topSort :: Ord a => AdjacencyMap a -> [a]
-topSort (AM.topSort . coerce -> Right vs) = vs
-topSort _ = error "topSort on Acyclic Graph produced cycle"
+topSort g = case AM.topSort (coerce g) of
+  Right vs -> vs
+  Left _ ->error "Internal error: the acyclicity invariant is violated in topSort"
 
 -- | Compute the acyclic /condensation/ of a graph, where each vertex
 -- corresponds to a /strongly-connected component/ of the original graph. Note
--- a/src/Algebra/Graph/Acyclic/AdjacencyMap.hs
+++ b/src/Algebra/Graph/Acyclic/AdjacencyMap.hs
@@ -487,7 +487,7 @@ transitiveClosure = coerce AM.transitiveClosure
 topSort :: Ord a => AdjacencyMap a -> [a]
 topSort g = case AM.topSort (coerce g) of
   Right vs -> vs
-  Left _ ->error "Internal error: the acyclicity invariant is violated in topSort"
+  Left _ -> error "Internal error: the acyclicity invariant is violated in topSort"
 
 -- | Compute the acyclic /condensation/ of a graph, where each vertex
 -- corresponds to a /strongly-connected component/ of the original graph. Note
--- a/src/Algebra/Graph/Acyclic/AdjacencyMap.hs
+++ b/src/Algebra/Graph/Acyclic/AdjacencyMap.hs
@@ -1,3 +1,5 @@
+{-# language ViewPatterns #-}
+
 -----------------------------------------------------------------------------
 -- |
 -- Module     : Algebra.Graph.Acyclic.AdjacencyMap
@@ -52,7 +54,6 @@ module Algebra.Graph.Acyclic.AdjacencyMap (
     consistent
     ) where
 
-import Data.Either
 import Data.Set (Set)
 import Data.Coerce (coerce)
 
@@ -481,11 +482,10 @@ transitiveClosure = coerce AM.transitiveClosure
 -- topSort ('vertex' x)            == [x]
 -- topSort (1 * (2 + 4) + 3 * 4) == [1, 2, 3, 4]
 -- topSort ('join' x y)            == 'fmap' 'Left' (topSort x) ++ 'fmap' 'Right' (topSort y)
--- topSort                       == 'fromJust' . 'AM.topSort' . 'fromAcyclic'
+-- 'Right' . topSort               == 'AM.topSort' . 'fromAcyclic'
 -- @
 topSort :: Ord a => AdjacencyMap a -> [a]
-topSort (AAM am) = fromRight impossible (AM.topSort am) where
-  impossible = error "topSort on Acyclic Graph returned cycle"
+topSort (AM.topSort . coerce -> Right vs) = vs
 
 -- | Compute the acyclic /condensation/ of a graph, where each vertex
 -- corresponds to a /strongly-connected component/ of the original graph. Note
--- a/src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
+++ b/src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
@@ -241,47 +241,39 @@ pattern TreeEdge <- Nothing
 pattern BackEdge :: Maybe Entry
 pattern BackEdge <- Just (Left _)
 -- Pattern to retrieve Parent from table lookup
-pattern Parent :: a -> Maybe (Either a a)
-pattern Parent p <- Just (Left p)
-
-unexplored :: MonadState S m => Int -> m Bool
-unexplored u = gets (not . IntMap.member u . table)
-
-enter :: MonadState S m => Int -> Int -> m ()
-enter u v = modify' (\(S p vs) -> S (IntMap.insert v (Left u) p) vs)
-
-exit :: MonadState S m => Int -> m ()
-exit v = modify' (\(S p vs) -> S (IntMap.alter (fmap done) v p) (v:vs)) where
-  done = \case
-    Left x -> Right x
-    _ -> error "impossible"
-
-classify :: MonadState S m => Int -> m (Maybe Entry)
-classify v = gets (IntMap.lookup v . table)
-
-retrace :: Int -> Int -> ParentTable -> Cycle Int
-retrace x x0 table = aux (x :| []) where
-  aux xs@(x :| _) | x0 == x = xs
-                  | otherwise = case IntMap.lookup x table of
-                      Parent z -> aux (z <| xs)
-                      _ -> error "impossible"
+pattern Parent :: Int -> Entry
+pattern Parent p <- Left p
 
 topSort' :: (MonadState S m, MonadCont m) => AdjacencyIntMap -> m TopSort
-topSort' g = callCC $ \cyclic -> do
-  let vertices = map fst $ IntMap.toDescList $ adjacencyIntMap g
-      adjacent = IntSet.toDescList . flip postIntSet g
-      dfs z x = do
-        enter z x
-        forM_ (adjacent x) $ \y ->
-          classify y >>= \case
-            TreeEdge -> dfs x y
-            BackEdge -> cyclic . Left . retrace x y =<< gets table
-            _        -> return ()
-        exit x
-  forM_ vertices $ \v -> do
-    new <- unexplored v
-    when new $ dfs v v
-  Right <$> gets order
+topSort' g = callCC $ \cyclic ->
+  do let vertices = map fst $ IntMap.toDescList $ adjacencyIntMap g
+         adjacent = IntSet.toDescList . flip postIntSet g
+         dfs z x =
+           do enter z x
+              forM_ (adjacent x) $ \y ->
+                classify y >>= \case
+                  TreeEdge -> dfs x y
+                  BackEdge -> cyclic . Left . retrace x y =<< gets table
+                  _        -> return ()
+              exit x
+     forM_ vertices $ \v ->
+       do new <- unexplored v
+          when new $ dfs undefined v
+     Right <$> gets order
+  where
+    classify v = gets (IntMap.lookup v . table)
+    unexplored u = gets (not . IntMap.member u . table)
+    enter u v = modify' (\(S p vs) -> S (IntMap.insert v (Left u) p) vs)
+    exit v = modify' (\(S p vs) -> S (IntMap.alter (fmap done) v p) (v:vs))
+    done = \case
+      Left x -> Right x
+      _ -> error "impossible"
+    retrace x x0 table = aux (x :| []) where
+      aux xs@(x :| _) | x0 == x = xs
+                      | otherwise = case table IntMap.! x of
+                          Parent z -> aux (z <| xs)
+                          _ -> error "impossible"
+
 
 -- | Compute a topological sort of the vertices of a graph. Given a
 --  DAG, the lexicographically least topological ordering is returned,
@@ -301,7 +293,7 @@ topSort' g = callCC $ \cyclic -> do
 -- @
 topSort :: AdjacencyIntMap -> Either (Cycle Int) [Int]
 topSort g = runContT (evalStateT (topSort' g) initialState) id where
-  initialState = S mempty mempty 
+  initialState = S mempty mempty
 
 -- | Check if a given graph is /acyclic/.
 --
@@ -314,7 +306,7 @@ topSort g = runContT (evalStateT (topSort' g) initialState) id where
 isAcyclic :: AdjacencyIntMap -> Bool
 isAcyclic g = case topSort g of
   Right _ -> True
-  Left  _ -> False
+  Left _ -> False
 
 -- | Check if a given forest is a correct /depth-first search/ forest of a graph.
 -- The implementation is based on the paper "Depth-First Search and Strong
--- a/src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
+++ b/src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
@@ -258,7 +258,7 @@ topSort' g = callCC $ \cyclic -> do
 -- is cyclic.
 --
 -- @
--- topSort (1 * 2 + 3 * 1)               == Just [3,1,2]
+-- topSort (1 * 2 + 3 * 1)               == Right [3,1,2]
 -- topSort (1 * 2 + 2 * 1)               == Nothing
 -- fmap ('flip' 'isTopSortOf' x) (topSort x) /= Just False
 -- 'isJust' . topSort                      == 'isAcyclic'
--- a/src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
+++ b/src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
@@ -1,4 +1,4 @@
-{-# language LambdaCase, ViewPatterns, PatternSynonyms #-}
+{-# language LambdaCase, PatternSynonyms #-}
 
 -----------------------------------------------------------------------------
 -- |
@@ -189,7 +189,7 @@ dfsForestFrom :: [Int] -> AdjacencyIntMap -> Forest Int
 dfsForestFrom vs g = dfsForestFrom' [ v | v <- vs, hasVertex v g ] g
 
 dfsForestFrom' :: [Int] -> AdjacencyIntMap -> Forest Int
-dfsForestFrom' vs g = evalState (explore vs) mempty where
+dfsForestFrom' vs g = evalState (explore vs) IntSet.empty where
   explore (v:vs) = discovered v >>= \case
     True -> (:) <$> walk v <*> explore vs
     False -> explore vs
@@ -281,16 +281,15 @@ topSort' g = callCC $ \cyclic ->
     classify v = gets (IntMap.lookup v . table)
     unexplored u = gets (not . IntMap.member u . table)
     enter u v = modify' (\(S p vs) -> S (IntMap.insert v (Left u) p) vs)
-    exit v = modify' (\(S p vs) -> S (IntMap.alter (fmap done) v p) (v:vs))
-    done = \case
-      Left x -> Right x
-      _ -> error "impossible"
+    exit v = modify' (\(S p vs) -> S (IntMap.alter (fmap done) v p) (v:vs)) where
+      done = \case
+        Left x -> Right x
+        _ -> error "Internal error: dfs node was exitted before it was entered"
     retrace x x0 table = aux (x :| []) where
       aux xs@(x :| _) | x0 == x = xs
                       | otherwise = case table IntMap.! x of
                           Parent z -> aux (z <| xs)
-                          _ -> error "impossible"
-
+                          _ -> error "Internal error: dfs back edge was mistakingly identified"
 
 -- | Compute a topological sort of a DAG or discover a cycle.
 --
@@ -318,10 +317,12 @@ topSort' g = callCC $ \cyclic ->
 -- 'isRight' . topSort                          == 'isAcyclic'
 -- @
 topSort :: AdjacencyIntMap -> Either (Cycle Int) [Int]
-topSort g = runContT (evalStateT (topSort' g) initialState) id where
-  initialState = S mempty mempty
+topSort g = runContT (evalStateT (topSort' g) (S IntMap.empty [])) id 
 
 -- | Check if a given graph is /acyclic/.
+--
+--   Let /W/ be the number of bits in a machine word. Complexity:
+--   /O((n+m)*W)/ time and /O(n)/ space.
 --
 -- @
 -- isAcyclic (1 * 2 + 3 * 1) == True
--- a/src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
+++ b/src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
@@ -207,7 +207,7 @@ dfsForestFrom' vs g = evalState (explore vs) mempty where
 -- dfs [3] $ 'circuit' [1..5] + 'circuit' [5,4..1] == [3,2,1,5,4]
 -- @
 dfs :: [Int] -> AdjacencyIntMap -> [Int]
-dfs vs = concatMap flatten . dfsForestFrom vs
+dfs vs = dfsForestFrom vs >=> flatten
 
 -- | Compute the list of vertices that are /reachable/ from a given source
 -- vertex in a graph. The vertices in the resulting list appear in
@@ -227,37 +227,32 @@ dfs vs = concatMap flatten . dfsForestFrom vs
 reachable :: Int -> AdjacencyIntMap -> [Int]
 reachable x = concat . bfs [x]
 
-data S = S { table :: !(IntMap.IntMap Int)
-           , seen :: !(IntMap.IntMap Bool)
-           , order :: [Int] } deriving (Show)
-
 type TopOrder = Either [Int] [Int]
-
-backtrack :: Int -> [Int] -> IntMap.IntMap Int -> [Int]
-backtrack w vs@(v:_) table = case IntMap.lookup v table of
-  Just u -> if u == w then vs else backtrack w (u:vs) table
-  Nothing -> vs
+type ParentTable = IntMap.IntMap (Maybe Int,Bool)
+data S = S { table :: !ParentTable, order :: [Int] }
 
 topSort' :: (MonadState S m, MonadCont m) => AdjacencyIntMap -> m TopOrder
 topSort' g = callCC $ \cyclic -> do
-  let enter v = modify' (\(S p s vs) -> S p (IntMap.insert v False s) vs)
-      exit v = modify' (\(S p s vs) -> S p (IntMap.insert v True s) (v:vs))
-      parent u v = modify' (\(S p s vs) -> S (IntMap.insert v u p) s vs)
-      nodeState v = gets (IntMap.lookup v . seen)
+  let parent u v = modify' (\(S p vs) -> S (IntMap.alter (register u) v p) vs)
+      register u = const (Just (u, False)) -- mark parent as u 
+      explored v = modify' (\(S p vs) -> S (IntMap.alter mark v p) (v:vs))
+      mark = fmap (fmap (const True)) -- mark tree as explored/done
+      node_state v = gets (IntMap.lookup v . table)
       unexplored u = gets (not . IntMap.member u . table)
-      explore u =
-        do enter u
-           forM_ (IntSet.toDescList (postIntSet u g)) $ \v ->
-             nodeState v >>= \case
-               Nothing -> parent u v >> explore v
-               -- True => tree fully explored, False => not (cycle)
-               Just False -> cyclic . Left . backtrack v [u,v] =<< gets table
-               Just True -> pure () 
-           exit u
-           
-  forM_ (map fst $ IntMap.toDescList $ adjacencyIntMap g) $ \v ->
-    do new <- unexplored v
-       when new $ explore v
+      build_cycle u v = cyclic . Left . expand_cycle v [u,v] =<< gets table
+      expand_cycle w vs@(v:_) table =
+        case IntMap.lookup v table of
+          Just (Just u,_) -> if u == v then vs else expand_cycle w (u:vs) table
+          Just _ -> tail vs
+      dfs u = do forM_ (IntSet.toDescList $ postIntSet u g) $ \v ->
+                   node_state v >>= \case
+                     Nothing -> parent (Just u) v >> dfs v -- new node
+                     Just (_,True) -> pure () -- part of explored tree
+                     Just (_,False) -> build_cycle u v -- part of cycle
+                 explored u
+  forM_ (map fst $ IntMap.toDescList $ adjacencyIntMap g) $
+    \v -> do new_tree <- unexplored v
+             when new_tree $ parent Nothing v >> dfs v
   Right <$> gets order
 
 -- | Compute the /topological sort/ of a graph or return @Nothing@ if the graph
@@ -271,7 +266,7 @@ topSort' g = callCC $ \cyclic -> do
 -- @
 topSort :: AdjacencyIntMap -> Either [Int] [Int]
 topSort g = runContT (runStateT (topSort' g) initialState) fst where
-  initialState = S mempty mempty mempty
+  initialState = S mempty mempty 
 
 -- | Check if a given graph is /acyclic/.
 --
--- a/src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
+++ b/src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
@@ -32,12 +32,10 @@ import Control.Monad
 import Control.Monad.State.Strict
 import Control.Monad.Cont
 import Data.Either
-import Data.Maybe
 import Data.Tree
 
 import Algebra.Graph.AdjacencyIntMap
 
-import qualified Data.Graph.Typed   as Typed
 import qualified Data.IntMap.Strict as IntMap
 import qualified Data.IntSet        as IntSet
 
@@ -239,6 +237,7 @@ type TopOrder = Either [Int] [Int]
 backtrack :: Int -> Int -> [Int] -> IntMap.IntMap Int -> Either [Int] a
 backtrack v u vs table = case IntMap.lookup u table of
   Just w -> if w == v then Left (u:vs) else backtrack v w (u:vs) table
+  Nothing -> Left vs
 
 topSort' :: (MonadState S m, MonadCont m) => AdjacencyIntMap -> m TopOrder
 topSort' g = callCC $ \cyclic -> do
--- a/src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
+++ b/src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
@@ -186,12 +186,11 @@ dfsForestFrom' :: [Int] -> AdjacencyIntMap -> Forest Int
 dfsForestFrom' vs g = evalState (explore vs) mempty where
   explore (v:vs) = discovered v >>= \case
     True -> (:) <$> walk v <*> explore vs
-    False -> explore vs 
+    False -> explore vs
   explore [] = return []
-  walk v = Node v <$> (explore $ adjacent v)
+  walk v = Node v <$> explore (adjacent v)
   adjacent v = IntSet.toList (postIntSet v g)
-  undiscovered v = gets (not . IntSet.member v)
-  discovered v = do new <- undiscovered v
+  discovered v = do new <- gets (not . IntSet.member v)
                     when new $ modify' (IntSet.insert v)
                     return new
 
--- a/src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
+++ b/src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
@@ -238,9 +238,9 @@ topSort' g = callCC $ \cyclic -> do
       exit v = modify' (\(S p vs) -> S (IntMap.alter mark v p) (v:vs)) where
         mark = fmap (fmap (const True)) -- mark tree as explored/done
       node_state v = gets (IntMap.lookup v . table)
-      retrace v vs@(u:_) table@(IntMap.lookup u -> Just (Just w,_))
-        | w == v = v:vs
+      retrace v vs@(u:_) table@(IntMap.lookup u -> ~(Just (Just w,_)))
         | v == u = vs
+        | v == w = v:vs
         | otherwise = retrace v (w:vs) table
       dfs u =
         do forM_ (IntSet.toDescList $ postIntSet u g) $ \v ->
--- a/src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
+++ b/src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
@@ -274,9 +274,9 @@ topSort' g = callCC $ \cyclic -> do
 -- topSort (1 * 2 + 3 * 1)               == Right [3,1,2]
 -- topSort ('path' [1..5])                 == Right [1..5]
 -- topSort (3 * (1 * 4 + 2 * 5))         == Right [3,1,2,4,5]
--- topSort (1 * 2 + 2 * 1)               == Left [1,2]
--- topSort ('path' [5,4..1] + 'edge' 2 4)    == Left [4,3,2]
--- topSort ('circuit' [1..5])              == Left [1..5]
+-- topSort (1 * 2 + 2 * 1)               == Left (2 ':|' [1])
+-- topSort ('path' [5,4..1] + 'edge' 2 4)    == Left (4 ':|' [3,2])
+-- topSort ('circuit' [1..5])              == Left (5 ':|' [1..4])
 -- fmap ('flip' 'isTopSortOf' x) (topSort x) /= Right False
 -- 'isRight' . topSort                     == 'isAcyclic'
 -- @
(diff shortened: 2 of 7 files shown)


jitwit:
Another idea is I could try is basing the implementation off the Data.Graph one, without converting to arrays

jitwit:
Hah, so topsort did well: `topSort: 0.56 (good)`
dff didn't: `dff: 1.12 (bad)`

I suspect the issue with dff is that the clique is one of the graphs tested...

nobrakal:
@jitwit Great :D Be aware that the suite ran on Travis use pretty small graphs (100 vertices), and results may not be totally accurate.

Don't hesitate to run it by hand with bigger graphs and a detailed output to investigate what's going on for `topsort`.

jitwit:
@nobrakal good to know! I've been running some of my own ad-hoc benches, but I'll try using your benchmark suite as well

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
```diff
@@ -222,9 +236,22 @@ reachable x = concat . bfs [x]
 -- 'isJust' . topSort                      == 'isAcyclic'
 -- @
 topSort :: Ord a => AdjacencyMap a -> Maybe [a]
-topSort m = if isTopSortOf result m then Just result else Nothing
-  where
-    result = Typed.topSort (Typed.fromAdjacencyMap m)
+topSort g = check $ execState (explore $ vertexList g) (Set.empty,[]) where
+   -- todo add early exit if cycle detected. also add ability to detect
```
snowleopard: It would be nice if we could have `topSort :: AdjacencyMap a -> Either [a] [a]` where the `Left` case returns a cycle :)

snowleopard:
@jitwit Thanks, looks nice!

I suggest benchmarking the performance of DFS on `AdjacencyMap` directly. The performance suite requires a conversion from `Graph` to an `AdjacencyMap` which complicates the picture. Also, I'd say the performance on fully connected graphs is not a very important use-case: I'd personally prefer to optimise the library for sparse graphs, which are much more common. Happy to pay an extra constant factor for cliques.

By the way, could you by chance also fix this issue with your `topSort` implementation: #2?

jitwit:
Sounds good, I think the implementation does fairly well for sparse graphs. I'll double check with more benches!

Regarding #2, I think the current implementation might return the lexicographically greatest! Exploring the vertices high to low should give the least, I think... I'll double check.
Currently:
```
λ> topSort $ 3*(1*4+2*5)
Just [3,2,5,1,4]
```
 I'll also try to figure out a nice way to return the cycle if detected

jitwit:
@snowleopard Actually yes! I simply added `reverse` before `vertexList` and `postSet` to get:
```
λ> topSort $ 3*(1*4+2*5)
Just [3,1,2,4,5]
```
Would it be worth adding a functions to return the `postSet` descending using Data.IntSet's `toDescList` in the `Algebra.Graph.AdjacencyMap` modules?

snowleopard:
@jitwit Great!

> Would it be worth adding functions to return the `postSet` descending using Data.IntSet's `toDescList` in the `Algebra.Graph.AdjacencyMap` modules?

I'd prefer to avoid extending the API since it feels too big already. As far as I understand you don't lose any performance by simply applying `toDescList` to the result, so let's just do that.

jitwit:
@snowleopard, right, wasn't thinking clearly! 

For `vertexList` it looks like Data.IntMap doesn't export such a function besides `toDescList`, which returns keys and values. I guess `reverse $ vertexList g` is acceptable?

snowleopard:
> I guess `reverse $ vertexList g` is acceptable?

Why not `Set.toDescList $ vertexSet g` instead? I think this will be faster.

jitwit:
@snowleopard In the documentation the complexity for `keysSet` is indicated as `O(n*min(n,W))`, which is odd because I thought IntSet's were also Patricia Trees, just with no values. I can try both though and see if there's any measurable difference with benchmarks

jitwit:
Right, I guess another option is calling `foldlWithKey (\k _ xs -> k:xs) []`. That is how `toDescList` is implemented, without the tuples

snowleopard:
@jitwit I'm sorry I don't follow. Why are you referring to `Data.IntMap.toDescList`? In the code I suggested

```haskell
Set.toDescList $ vertexSet g 
```

the function `toDescList` comes from the module `Data.Set`.

snowleopard:
The `IntSet` equivalent would be: `IntSet.toDescList $ vertexIntSet g`.

jitwit:
@snowleopard Ah yes, sorry I was getting tired at that point of the day! Yeah I was assuming we were talking about IntMaps/Sets. 

I changed the topSort's to give the lexicographically least order, but that invalidated some of the tests for `Acyclic`, so I made some tweaks there.

To return the cycle if found, I think I'll add a Map keeping track of parents to the state, so that it is possible to backtrack and present the cycle if found.

snowleopard:
@jitwit Thanks! Have you done any further DFS/`topSort` benchmarking directly on adjacency maps?

> To return the cycle if found, I think I'll add a Map keeping track of parents to the state, so that it is possible to backtrack and present the cycle if found.

Ideally, this extra logic would be free in case of the `Right` answer, i.e. we would only pay the cost of finding the cycle when given a graph with a cycle. Hope it's possible.

jitwit:
> Ideally, this extra logic would be free in case of the `Right` answer, i.e. we would only pay the cost of finding the cycle when given a graph with a cycle. Hope it's possible.

@snowleopard Yes, that's my goal. I think it is possible, but I'm still working at it.

As far as benches, yes, I finally got around to them:

[https://github.com/jitwit/alga/blob/bench-dfs/bench/current_results.txt](https://github.com/jitwit/alga/blob/bench-dfs/bench/current_results.txt)

Unfortunately for AdjacencyMaps `dff` is  slower (11ms vs 7.6ms on the word graph). However, on that same graph represented as an AdjacencyIntMap, `dff` is faster (3.2 ms vs 7.5ms). 

Thankfully both AdjacencyMaps and AdjacencyIntMaps `topSort` appear to be faster. The graph I thought to test may not be the best example, but I wanted to test on something that wasn't a clique or path, and that's acyclic (word graph has cycles):
```
some_acyclic_graph :: AM.AdjacencyMap Integer
some_acyclic_graph = foldr (\x g -> (AM.vertex x * (AM.edge (x+10) (x+15)) + g)) AM.empty [1..10000]
```
In any case for AdjacencyMap Integer the speedup is 26ms vs 48ms, and 4ms vs 27ms for AdjacencyIntMaps. What might be a good example to test `topSort` on otherwise?

jitwit:
> Ideally, this extra logic would be free in case of the `Right` answer, i.e. we would only pay the cost of finding the cycle when given a graph with a cycle. Hope it's possible.

@snowleopard Yes, that's my goal. I think it is possible, but I'm still working at it.

As far as benches, yes, I finally got around to them:

[https://github.com/jitwit/alga/blob/bench-dfs/bench/current_results.txt](https://github.com/jitwit/alga/blob/bench-dfs/bench/current_results.txt)

There is a small speedup for AdjacencyMaps `dff` 25.5ms vs 26.5ms on the word graph. On that same graph represented as an AdjacencyIntMap, `dff` is faster (10 ms vs 29ms). 

Both AdjacencyMaps and AdjacencyIntMaps `topSort` appear to be quite a bit faster. The graph I thought to test may not be the best example, but I wanted to test on something that wasn't a clique or path, and that's acyclic (word graph has cycles):
```
some_acyclic_graph :: AM.AdjacencyMap Integer
some_acyclic_graph = foldr (\x g -> (AM.vertex x * (AM.edge (x+10) (x+15)) + g)) AM.empty [1..10000]
```
For AdjacencyMap Integer the speedup is 26ms vs 49ms, and for AdjacencyIntMaps 4ms vs 29ms . What might be a good example to test `topSort` on otherwise?

jitwit:
@snowleopard I think I have a correct version of `topSort` that identifies cycles with no extra cost if called on a DAG! 

One issue that's come up is changing `fromRight` not being in base for ghc 7.10.3 and 8.0.2. Several of the tests and some of the functions in the `Acyclic` module were calling`fromJust`. What should it be replaced with?

jitwit:
```
λ> topSort (path [1..10] + edge 5 2)
Left [2,3,4,5]
λ> topSort (path [1..10])
Right [1,2,3,4,5,6,7,8,9,10]
```

jitwit:
Changes to previous discussion about benchmarks:

* With early escape `topSort` on AdjacencyMaps is somehow faster (20ms vs 28 previously)? Twice as slow for AdjacencyIntMaps (8ms vs 4ms). I'm guessing the overhead is from monad transformers?
* With early escape, `topSort` is now on the order of 100,000 times faster than the old version using Data.Graph on the Word Graph (has cycles)! old alga: 24.7ms, early escape: 323ns. Same graph with AdjacencyIntMaps the difference is 26.7 ms to 112 ns

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
jitwit: Done! (well, first draft)

test/Algebra/Graph/Test/Generic.hs
```diff
@@ -1986,14 +1986,15 @@ testReachable (prefix, API{..}) = do
...
 testTopSort (prefix, API{..}) = do
     putStrLn $ "\n============ " ++ prefix ++ "topSort ============"
-    test "topSort (1 * 2 + 3 * 1)               == Just [3,1,2]" $
-          topSort (1 * 2 + 3 * 1)               == Just [3,1,2]
+    test "topSort (1 * 2 + 3 * 1)               == Right [3,1,2]" $
+          topSort (1 * 2 + 3 * 1)               == Right [3,1,2]
 
-    test "topSort (1 * 2 + 2 * 1)               == Nothing" $
-          topSort (1 * 2 + 2 * 1)               == Nothing
+    test "topSort (1 * 2 + 2 * 1)               == Left [1,2]" $
```
jitwit: Next pass I'll try to fix comment/documentation mismatches

snowleopard:
> I think I have a correct version of `topSort` that identifies cycles with no extra cost if called on a DAG! 

@jitwit Awesome!

> As far as benches, yes, I finally got around to them:
> https://github.com/jitwit/alga/blob/bench-dfs/bench/current_results.txt

Are these still up-to-date? If yes, the results are great! There is only one benchmark where you lose just a little bit to the DFS from `Data.Graph`, but in all other benchmarks you win with a comfortable margin. That's great!

The CI performance suite is happy too:

> dff: 0.55 (good)
> topSort: 0.64 (good)

I haven't yet looked at the implementation carefully, but I'll try to find time later today. As for results, I'm happy to merge this :-)

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
```diff
@@ -212,19 +226,51 @@ dfs vs = concatMap flatten . dfsForestFrom vs
...
+  Right <$> gets order
+
 -- | Compute the /topological sort/ of a graph or return @Nothing@ if the graph
 -- is cyclic.
 --
 -- @
--- topSort (1 * 2 + 3 * 1)               == Just [3,1,2]
--- topSort (1 * 2 + 2 * 1)               == Nothing
+-- topSort (1 * 2 + 3 * 1)               == Right [3,1,2]
+-- topSort (1 * 2 + 2 * 1)               == Left [2,1]
```
snowleopard: Is there a chance to return the lexicographically smallest cycle? In this case, it would be `[1,2]`.

jitwit:
@snowleopard Great! Yeah, the benches should be current to what's here now. 

For `topSort`, theres is a chance that using one map in the state `Map a (Maybe a,Bool)` brought a slight slowdown over having two, one `Map a a` and another `Map a Bool`.  The CI was getting like 0.58 before I made that change (now 0.64). I guess that might be because of the extra indirection in the pattern matching after lookups, but I'm not sure. I switched to one since it should use less memory, but maybe it's better to use two?

Also, there are some cases where I get incomplete pattern warnings. I left some off that are logically impossible. I wouldn't mind filling them out to make ghc happy if you prefer?

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
jitwit: In the case that the `topSort` succeeds we get the smallest topological ordering. To get that, the biggest vertices are expanded first. As soon as a back edge is found, it is expanded and returned. It might be the lexicographically biggest cycle? 

I'm not sure if it would be possible to look for smaller cycles without losing the smallest topSort or earliest short circuit. I could potentially present the found cycle with the smallest element at the head of the list though?

jitwit: Not the lexicographically greatest:
```
λ> topSort (circuit [1..5] + circuit [5,4..1])
Left [5,4]
```

snowleopard: Hmm, it looks more complicated than I thought. Perhaps, it's what happens in bipartite graphs:

https://github.com/snowleopard/alga/blob/00925932e060f005e2bcf463246762e34625f6f0/src/Algebra/Graph/Bipartite/AdjacencyMap.hs#L738-L742

That is, in this case, you found `[1, 2, 3, 4, 5]`, where the prefix is `[1, 2, 3]` and the cycle is `[4, 5]`. You just reported the cycle starting from the second vertex. If you report it from the first then I think we can use the same description as in the bipartite case.

snowleopard: /cc @vasalf

snowleopard:
> The CI was getting like 0.58 before I made that change (now 0.64).

This could be just CI noise, I wouldn't base anything on such a small difference.

> I wouldn't mind filling them out to make ghc happy if you prefer?

Yes, please. I don't like incomplete pattern warnings :) Perhaps, you could eliminate them by using a more sophisticated data structure that doesn't allow for impossible states to be represented.

jitwit:
> Yes, please. I don't like incomplete pattern warnings :) Perhaps, you could eliminate them by using a more sophisticated data structure that doesn't allow for impossible states to be represented.
 
Yeah I find them jarring as well. One is in the Acyclic module, which is supposed to be acyclic by construction. The other is in retracing, where the root of the tree has no parent, but the cycle is guaranteed to be found before falling off the end. I got rid of the ones that were in classifying the edges

jitwit:
The cycle could be returned as a NonEmpty list? that's another source of incomplete pattern warning...

snowleopard:
> The cycle could be returned as a NonEmpty list?

Yes, sounds good! I think we should also do this in bipartite graphs.

jitwit:
Would it be feasible to have a module that exports common type synonyms? Names like `Cycle a` (to stand for `NonEmpty a`) could be shared in the Algorithms modules, the Acyclic one,  the Bipartite one, as well as the Tests

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
jitwit: Hm! So I thought about it while running, and I believe it returns the lexicographically least cycle on `a` with opposite `Ord a` instance.

vasalf: Is it right that this `topSort` code goes through vertices in descending order and tries to find a cycle with a standard depth-first search? Then the returned cycle is not lexicographically largest.

For example, consider the following graph.

```Haskell
edges [(1,2),(2,1),(5,2),(3,4),(4,3)]
```

In this graph the algorithm which goes through vertices in descending order would first go from 5 and find the cycle `[2, 1]`, while there is a cycle `[4,3]`.

If I'm correct, `path ++ cycle` is lexicographically largest, just as in `detectParts` case. In my example, it is `[5] ++ [2,1]` versus `[] ++ [4,3]`.

jitwit: Ah yes, you're right. I think it is still the largest within in the connected component containing the largest vertex?

jitwit: You're characterization fails in the example `circuit [5,4..1] + circuit [1..5]`, where the lexicographically largest `path ++ cycle` would be `[5,4,3,2,1]`, but the cycle `[5,4]` is found by the algorithm. `[5,4]` being smallest with respect to `flip <`.

vasalf: Oh, there's a difference between «lexicographically largest» and «lexicographically smallest with respect to `flip <`» and now I see it.
Yes, I agree that `path ++ cycle` should be lexicographically smallest with respect to `flip <`.

jitwit: hah yes. I just noticed one more subtlety: taking your example, but with 5 -> 1 instead of 5 -> 2, we get
```
λ> topSort $ 1*2+2*1+3*4+4*3+5*1
Left (1 :| [2])
```
So it's lexicographically least with respect to flip <, up to rotation... I think

jitwit:
Switching from `Map a (Maybe a, Bool)` to `Map a (Either a a)` seems to give a 10% improvement for topSort!

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -479,12 +480,13 @@ transitiveClosure = coerce AM.transitiveClosure
...
--- topSort (1 * (2 + 4) + 3 * 4) == [3, 1, 4, 2]
+-- topSort (1 * (2 + 4) + 3 * 4) == [1, 2, 3, 4]
 -- topSort ('join' x y)            == 'fmap' 'Left' (topSort x) ++ 'fmap' 'Right' (topSort y)
--- topSort                       == 'Data.Maybe.fromJust' . topSort . 'fromAcyclic'
+-- 'Right' . topSort               == 'AM.topSort' . 'fromAcyclic'
 -- @
-topSort :: (Ord a) => AdjacencyMap a -> [a]
-topSort (AAM am) = Typed.topSort (Typed.fromAdjacencyMap am)
+topSort :: Ord a => AdjacencyMap a -> [a]
+topSort (AM.topSort . coerce -> Right vs) = vs
```
snowleopard: I find this rather cryptic: why not do a simple pattern matching on the result of `AM.topSort`?

snowleopard: (I guess I'm just not a fan of `ViewPatterns`, although in some cases its use is justified.)

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
```diff
@@ -479,12 +480,13 @@ transitiveClosure = coerce AM.transitiveClosure
...
+topSort _ = error "topSort on Acyclic Graph produced cycle"
```
snowleopard: Just a minor rephrasing:
```suggestion
topSort _ = error "topSort on Acyclic Graph produced cycle"
topSort _ = error "Internal error: the acyclicity invariant is violated in topSort"
```

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -169,10 +184,27 @@ dfsForest = Typed.dfsForest . Typed.fromAdjacencyIntMap
 -- 'forest' (dfsForestFrom [3] $ 'circuit' [1..5] + 'circuit' [5,4..1]) == 'path' [3,2,1,5,4]
 -- @
 dfsForestFrom :: [Int] -> AdjacencyIntMap -> Forest Int
-dfsForestFrom vs = Typed.dfsForestFrom vs . Typed.fromAdjacencyIntMap
+dfsForestFrom vs g = dfsForestFrom' [ v | v <- vs, hasVertex v g ] g
+
+dfsForestFrom' :: [Int] -> AdjacencyIntMap -> Forest Int
+dfsForestFrom' vs g = evalState (explore vs) mempty where
```
snowleopard: Unless I'm mistaken, `mempty` here corresponds to `IntSet.empty`, so let's avoid unnecessary indirection.
```suggestion
dfsForestFrom' vs g = evalState (explore vs) IntSet.empty where
```

src/Algebra/Graph/Acyclic/AdjacencyMap.hs
jitwit: Yeah, I probably went overboard with ViewPatterns, I think most can be removed

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
```diff
@@ -173,10 +185,26 @@ dfsForest g = dfsForestFrom (vertexList g) g
...
+dfsForestFrom vs g = dfsForestFrom' [ v | v <- vs, hasVertex v g ] g
+
+dfsForestFrom' :: Ord a => [a] -> AdjacencyMap a -> Forest a
+dfsForestFrom' vs g = evalState (explore vs) mempty where
+  explore (v:vs) = discovered v >>= \case
+    True -> (:) <$> walk v <*> explore vs
+    False -> explore vs 
+  explore [] = return []
+  walk v = Node v <$> explore (adjacent v)
+  adjacent v = Set.toList (postSet v g)
```
snowleopard: I find it interesting that `postSet` is the only function on graphs that you need. This means you could probably write it generically, for any `ToGraph`, and simply re-export it here with a nicer specialised type (unless there are some nasty import cycles).

snowleopard: On second thought, it's not clear if a generic implementation is really useful: it may be faster to convert a whole graph to an `AdjacencyMap` instead of computing `postSet` on a vertex-by-vertex basis.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -206,32 +241,96 @@ dfs vs = concatMap flatten . dfsForestFrom vs
 -- 'isSubgraphOf' ('vertices' $ reachable x y) y == True
 -- @
 reachable :: Int -> AdjacencyIntMap -> [Int]
-reachable x = concat . bfs [x]
+reachable x = dfs [x]
```
snowleopard: Interesting that switching back to DFS gives us another performance improvement `reachable: 0.83` :)

jitwit: Yeah, I was happy to see that. I felt intuitively that something must be wrong if bfs is performing better for reachable than dfs!

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: There were a few other cases where I had used `mempty`, I think I caught all of them

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -206,32 +241,96 @@ dfs vs = concatMap flatten . dfsForestFrom vs
...
+pattern BackEdge <- Just (Left _)
+-- Pattern to retrieve Parent from table lookup
+pattern Parent :: Int -> Entry
+pattern Parent p <- Left p
+
+topSort' :: (MonadState S m, MonadCont m) => AdjacencyIntMap -> m TopSort
+topSort' g = callCC $ \cyclic ->
+  do let vertices = map fst $ IntMap.toDescList $ adjacencyIntMap g
+         adjacent = IntSet.toDescList . flip postIntSet g
+         dfs z x =
```
snowleopard: Shall we give a bit more descriptive names to variables? I suggest: `z -> prev`, `x -> cur`, `y -> next`.

(I'm slowly going through the implementation, so apologies if didn't get the meaning right.)

jitwit: Yes, z was prev, x current :). I can also get rid of the name shadowing if it makes the code too cryptic.

When the representation was more convoluted, I found the PatternSynonyms were kind of necessary to avoid horrible pattern matching. Currently, things are a bit simpler, and I could do away with them?

snowleopard: Yes, I think in this case you could rename `classify` to something like `getBackEdge` and then results `Nothing` and `Just _` may become clear without pattern synonyms. I haven't yet figured out what `Left` and `Right` mean, some more comments (perhaps at the top next to the definitions of data types/synonyms) would be helpful!

jitwit: Ah, yes that makes sense, a suitable name should make the patterns clear. 

`enter prev curr` enteres `(Left prev)` as the parent of `curr`, and on exiting promotes it to `Right prev`. I was considering just defining a datatype along the lines of:

`data Entry a = Root | Started a | Finished a` 

to indicate parents. Then entering `undefined` for `prev` could also be avoided when starting a new component

snowleopard:
> Would it be feasible to have a module that exports common type synonyms? Names like `Cycle a` (to stand for `NonEmpty a`) could be shared in the Algorithms modules, the Acyclic one, the Bipartite one, as well as the Tests

@jitwit Yes, we might do that at some point, but for now, let's just define type synonyms locally. In the case of `Bipartite.OddCycle` we might at some point switch a data type that forces the odd length.

P.S.: This and the next weeks are quite busy, so apologies for the delays.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
snowleopard: Ah, yes, that seems like a nice naming scheme! I hope it doesn't come with any performance penalty?

jitwit: Yeah same, I'll try it and double check

jitwit: > @jitwit Yes, we might do that at some point, but for now, let's just define type synonyms locally. In the case of `Bipartite.OddCycle` we might at some point switch a data type that forces the odd length.
> 
> P.S.: This and the next weeks are quite busy, so apologies for the delays.

Sounds good. Also no worries.

By the way, in thinking about complexity analysis for `dfs` I realized the ones for `bfs` were either for a previous implementation or were just wrong! Should I open a separate pull request to correct them?

snowleopard: > Should I open a separate pull request to correct them?

Ideally, unrelated changes should be in a separate PR, but I don't feel too strongly about minor fixes, typos, etc., so I'll let you decide :)

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -169,10 +184,27 @@ dfsForest = Typed.dfsForest . Typed.fromAdjacencyIntMap
 -- 'forest' (dfsForestFrom [3] $ 'circuit' [1..5] + 'circuit' [5,4..1]) == 'path' [3,2,1,5,4]
 -- @
 dfsForestFrom :: [Int] -> AdjacencyIntMap -> Forest Int
-dfsForestFrom vs = Typed.dfsForestFrom vs . Typed.fromAdjacencyIntMap
+dfsForestFrom vs g = dfsForestFrom' [ v | v <- vs, hasVertex v g ] g
+
+dfsForestFrom' :: [Int] -> AdjacencyIntMap -> Forest Int
+dfsForestFrom' vs g = evalState (explore vs) IntSet.empty where
+  explore (v:vs) = discovered v >>= \case
```
nobrakal: Isn't `explore` just a `foldr` under the hood ?

jitwit: We want it to be foldl to expand the from the smallest . Also, if a monadielementsc foldl is used, the components would come back reversed, so I went with explicit recursion. I may be missing a nicer way to express this of course

jitwit: I guess like a `filterMapM` type thing is what's wanted, of a `mapM f`, where `f :: a -> Maybe (Tree a)`, but that adds unnecessary indirection

jitwit: The way I really wanted to express it was using 

`unfoldForestM :: Monad m => (b -> m (a, [b])) -> [b] -> m (Forest a)` 

The issue was that each `b` is given a tree, so I ended up to wrapping them with `Maybe` and pruning the result to discard the `Nothing` nodes.

snowleopard: I don't have a strong objection to using explicit recursion here but this does look like `foldlM` to me:

https://hackage.haskell.org/package/base-4.12.0.0/docs/Data-Foldable.html#v:foldlM

@jitwit Could you clarify why `foldlM` doesn't work?

snowleopard:
@jitwit #227 was merged and now we've got some conflicts. Could you please rebase?

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: It would work in the sense that smaller vertices get explored first, but the forest comes out reversed:
```
λ> map flatten $ dfsForest $ path [1..3] + path [4..6]
[[4,5,6],[1,2,3]]
```
That's with internal functions written as:
```
search = foldlM explore' [] vs
explore' vs v = discovered v >>= \case
  True -> (:vs) <$> walk v
  False -> return vs
```

snowleopard: > but the forest comes out reversed

Well, this is expected since you reverse the result by placing `vs` at the end in `(:vs) <$> walk v`.

I see what you mean though: DFS is more naturally expressed using your custom function `explore`, so let's keep it :)

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -213,32 +239,86 @@ dfs vs = concatMap flatten . dfsForestFrom vs
...
+-- topSort ('circuit' [1..3] + 'circuit' [3,2,1]) == Left (3 ':|' [2])
+-- topSort (1*2 + 2*1 + 3*4 + 4*3 + 5*1)      == Left (1 ':|' [2])
+-- fmap ('flip' 'isTopSortOf' x) (topSort x)      /= Right False
+-- 'isRight' . topSort                          == 'isAcyclic'
 -- @
-topSort :: AdjacencyIntMap -> Maybe [Int]
-topSort m = if isTopSortOf result m then Just result else Nothing
-  where
-    result = Typed.topSort (Typed.fromAdjacencyIntMap m)
+topSort :: AdjacencyIntMap -> Either (NonEmpty Int) [Int]
```
snowleopard: I suggest exporting the type synonym `Cycle` and using it in the type signature for greater clarity.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -213,32 +239,86 @@ dfs vs = concatMap flatten . dfsForestFrom vs
...
 -- @
 -- isAcyclic (1 * 2 + 3 * 1) == True
 -- isAcyclic (1 * 2 + 2 * 1) == False
 -- isAcyclic . 'circuit'       == 'null'
--- isAcyclic                 == 'isJust' . 'topSort'
+-- isAcyclic                 == 'isRight' . 'topSort'
 -- @
 isAcyclic :: AdjacencyIntMap -> Bool
-isAcyclic = isJust . topSort
+isAcyclic g = case topSort g of
```
snowleopard: I think this should be just `isRight . topSort`:

https://hackage.haskell.org/package/base-4.12.0.0/docs/Data-Either.html#v:isRight

src/Algebra/Graph/ToGraph.hs
```diff
@@ -266,7 +267,7 @@ class ToGraph t where
     -- @
     -- topSort == Algebra.Graph.AdjacencyMap.'AM.topSort' . toAdjacencyMap
     -- @
-    topSort :: Ord (ToVertex t) => t -> Maybe [ToVertex t]
+    topSort :: Ord (ToVertex t) => t -> Either (NonEmpty (ToVertex t)) [ToVertex t]
```
snowleopard: I guess you should be able to reuse `AM.Cycle` here as well.

snowleopard: Actually, the documentation above is outdated, it should be:

`topSort == Algebra.Graph.AdjacencyMap.Algorithm.'AM.topSort' . toAdjacencyMap`

It would be great if you fix this and other similar problems (if any) in this module. Feel free to do this in this PR.

test/Algebra/Graph/Test/Generic.hs
```diff
@@ -1986,14 +1987,35 @@ testReachable (prefix, API{..}) = do
...
-    test "topSort (1 * 2 + 2 * 1)               == Nothing" $
-          topSort (1 * 2 + 2 * 1)               == Nothing
-
-    test "fmap (flip isTopSortOf x) (topSort x) /= Just False" $ \x ->
-          fmap (flip isTopSortOf x) (topSort x) /= Just False
+    test "topSort (1 * 2 + 3 * 1)                    == Right [3,1,2]" $
+          topSort (1 * 2 + 3 * 1)                    == Right [3,1,2]
+                                                     
+    test "topSort (1 * 2 + 3 * 1)                    == Right [3,1,2]" $
+          topSort (1 * 2 + 3 * 1)                    == Right [3,1,2]
```
snowleopard: Seems to be a duplicated test?

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -213,32 +239,86 @@ dfs vs = concatMap flatten . dfsForestFrom vs
...
+--
+--   Complexity: /O((n+m)*min(n,W))/ time and /O(n)/ space.
 --
 -- @
--- topSort (1 * 2 + 3 * 1)               == Just [3,1,2]
--- topSort (1 * 2 + 2 * 1)               == Nothing
--- fmap ('flip' 'isTopSortOf' x) (topSort x) /= Just False
--- 'isJust' . topSort                      == 'isAcyclic'
+-- topSort (1 * 2 + 3 * 1)                    == Right [3,1,2]
+-- topSort ('path' [1..5])                      == Right [1..5]
```
snowleopard: We might also add something like `topSort . vertices == nub . sort`.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -213,32 +239,86 @@ dfs vs = concatMap flatten . dfsForestFrom vs
...
+-- topSort (3 * (1 * 4 + 2 * 5))              == Right [3,1,2,4,5]
+-- topSort (1 * 2 + 2 * 1)                    == Left (2 ':|' [1])
+-- topSort ('path' [5,4..1] + 'edge' 2 4)         == Left (4 ':|' [3,2])
+-- topSort ('circuit' [1..3])                   == Left (3 ':|' [1,2])
```
snowleopard: I wish we could return `[1..3]` in this case. Is it difficult to achieve? Right now all outputs seem unintuitive and it would be great if at least some of them were.

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
```diff
@@ -213,32 +240,87 @@ dfs vs = concatMap flatten . dfsForestFrom vs
...
+topSort' g = callCC $ \cyclic ->
+  do let vertices = map fst $ Map.toDescList $ adjacencyMap g
+         adjacent = Set.toDescList . flip postSet g
+         dfs prev curr =
+           do enter prev curr
+              forM_ (adjacent curr) $ \next ->
+                nodeState next >>= \case
+                  Nothing -> dfs curr next
+                  Just (Exited _) -> return ()
+                  _ -> cyclic . Left . retrace curr next =<< gets table
```
snowleopard: If you expand the remaining pattern `Just (Entered _)` it would be a bit easier to read. I hope it doesn't come with any performance cost.

snowleopard: In fact, if you had `Just (Entered v)` and traced the cycle starting from `v`, could it produce more intuitive results?

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
```diff
@@ -213,32 +240,87 @@ dfs vs = concatMap flatten . dfsForestFrom vs
...
+              exit curr
+     forM_ vertices $ \v -> nodeState v >>= \case
+       Nothing -> dfs undefined v
```
snowleopard: Ouch. Is it possible to avoid `undefined` here?

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: Should it be exported from both `Algebra.Graph.AdjacencyMap` and `Algebra.Graph.AdjacencyIntMap`? I think it would be nice if there was a module such that it could be defined once and visible where needed, but I'm not sure where makes sense.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: I went with `isRight` at first, but I think it caused some of the earlier GHCs to fail the CI tests

src/Algebra/Graph/ToGraph.hs
jitwit: Will do. Would it matter that `AM.Cycle` and `AIM.Cycle` both define the synonym?

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
jitwit: In cases like `edge 4 4`, or `circuit [1..3]`, `Entered v` will indicate the root of the search tree, which was previously `undefined`. `next` should be `v` in general, I think. 

The current documentation for `topSort` may be too convoluted, but it should accurately describe which cycle is found first. I definitely agree that it's not intuitive, and could modify it to present the cycle from the smallest node or something like that...

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
jitwit: I added two constructors to the `Entry` type to avoid this. No undefined, but more lines of code in the definitions

jitwit: Hopefully it doesn't affect performance too much

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: I could change it to scan the cycle for the minimum, and present from there. But I'm not sure that makes the result more intuitive? Currently, the first vertex listed is the first one visited during the search that belongs to the cycle

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: never mind! not sure what went wrong before!

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
snowleopard: > Currently, the first vertex listed is the first one visited during the search that belongs to the cycle

But `3` is not "the first one visited during the search that belongs to the cycle", is it? Yet it is listed first, this is what makes the result unintuitive to me. The first vertex on a cycle that we visit here is `1`, not `3`.

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
snowleopard: The CI result is still `dff: 0.50 (good)`, so performance seems to be unaffected!

What about your own experiments, do you see any degradation of performance because of the addition of new constructors?

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
snowleopard: Now that you added new constructors, it can no longer be `undefined`, right? If yes, please expand all `_` patterns, as otherwise, it's always necessary to make the additional effort of remembering the remaining constructors when reading the code. (Also, if a new constructor is added in future such catch-all pattern matches will be difficult to notice.)

src/Algebra/Graph/ToGraph.hs
snowleopard: I think it's fine if both modules define their own synonyms.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
snowleopard: Let's follow the simple approach with each module having its own synonym. It doesn't seem like a big duplication to me -- we'll never be mixing `Cycle`s from different modules in the same context.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: It is, since the list of vertices that gets searched is descending. In the case of `circuit [1..3]`, `3` is the first node expanded in `dfs`, it is the root of the search tree. 

I went with this ordering so we get the lexicographically least topological sort when there are no cycles. Since the `topSort` is the list of vertices sorted by their exit times during `dfs`, expanding the smaller vertices as late as possible gives the lexicographically least such enumeration.

snowleopard: Ah, my apologies, I keep forgetting about the fact that we use descending order.

I'm quite perplexed by the fact that it's difficult to have two good features, even though each feature seems easy enough on its own :)

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
jitwit: There appears to be no impact!

snowleopard: Awesome!

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
jitwit: Yes, the search now starts by `dfs OpenRoot v` when `v` is not yet in the search tree. 

There is still `_` for the case of a cycle, where otherwise 

```
                  Just (Entered v) -> cyclic . Left . retrace curr next =<< gets table
                  Just OpenRoot  -> cyclic . Left . retrace curr next =<< gets table
```

would need to be written. Shall I fill all such cases in?

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: Haha yeah. I was thinking it might be nice to have a search that wouldn't exit on the first cycle, but would classify all back edges found during the search and all tree edges.

This would lose some performance for the specific case where one wants to check for any cycle at all, but could be useful more generally.

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
snowleopard: I see. If there is a repetition `cyclic . Left . retrace curr next =<< gets table` you can just give it a name, something like `findCycle` and use in both cases.

What about `Just ClosedRoot`? Is it unreachable? Then better raise an `error` to at least make it obvious for the reader that this case shouldn't happen.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
snowleopard: > it might be nice to have a search that wouldn't exit on the first cycle, but would classify all back edges found during the search and all tree edges.

Hmm, sounds interesting. Could we find something like a forest of all "fundamental cycles" of a graph, and then use it compute not just one cycle, but all cycles if need be? 

It's probably not a very good idea since there will be around 0 users for this cool feature... :-)

OK, let's stop here and keep the order as is. I feel like I'm just overthinking it -- most likely, whatever cycle we return, the user will be happy :)

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
jitwit: `ClosedRoot` is matched against in `dfs` in the current implementation, where it's possible. It's not possible when recovering the cycle though, but where we still have `_` to match that, and an error is raised. I'll just add complete pattern matches everywhere?

snowleopard: Yes, I think it's just safer and clearer to use complete pattern matches everywhere.

jitwit: One small side effect of writing out full pattern matches is `-Wunused-matches` warnings during builds. Should that be suppressed by adding to the flags in the cabal file or something along those lines?

snowleopard: > One small side effect of writing out full pattern matches is `-Wunused-matches` warnings during builds.

Could you clarify which pattern is unused? I've had a look at the code but all patterns seem necessary.

> Should that be suppressed by adding to the flags in the cabal file or something along those lines?

No, this is a useful warning. If a pattern is indeed unused we should delete it.

jitwit: Changing ones like `Just (Exited _)` to `Just (Exited v)` gives the warning, the `v` is unused. 

I guess what's confusing is that the parent table and the node state table are blended into one. We could instead have two maps (one for parents, and one for node state `Entered` &`Exited`) , but I believe performance will be slower

jitwit: There will be two map operations instead of one each time a vertex is visited, but I'm trying it out in case the performance hit isn't as much as I expect

jitwit: Well, my intuition was wrong! It's faster than before (on cases with cycles and without). I guess less indirection tends to help! I also find it more readable.

snowleopard: Awesome, much nicer!

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -213,32 +243,86 @@ dfs vs = concatMap flatten . dfsForestFrom vs
...
+    enterRoot v = modify' (\(S m n vs) -> S m (IntMap.insert v Entered n) vs)
+    exit v = modify' (\(S m n vs) -> S m (IntMap.alter (fmap leave) v n) (v:vs))
+      where leave = \case
+              Entered -> Exited
+              Exited  -> error "Internal error: dfs search order violated"
+    retrace curr head table = aux (curr :| []) where
+      aux xs@(curr :| _)
+        | head == curr = xs
+        | otherwise = case table IntMap.! curr of
+            prev -> aux (prev <| xs)
```
snowleopard: I'm a bit confused about why this branch is implemented this way. Could we do the following instead?

```
aux (table IntMap.! curr <| xs)
```

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -213,32 +243,86 @@ dfs vs = concatMap flatten . dfsForestFrom vs
 -- 'isSubgraphOf' ('vertices' $ reachable x y) y == True
 -- @
 reachable :: Int -> AdjacencyIntMap -> [Int]
-reachable x = concat . bfs [x]
+reachable x = dfs [x]
+
+type Cycle = NonEmpty
+data NodeState = Entered | Exited
+data S = S { table :: IntMap.IntMap Int
```
snowleopard: Could we give a more meaningful name to `table`? What do the entries of the table correspond to?

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -176,10 +188,26 @@ dfsForest = Typed.dfsForest . Typed.fromAdjacencyIntMap
...
+dfsForestFrom' :: [Int] -> AdjacencyIntMap -> Forest Int
+dfsForestFrom' vs g = evalState (explore vs) IntSet.empty where
+  explore (v:vs) = discovered v >>= \case
+    True -> (:) <$> walk v <*> explore vs
+    False -> explore vs
+  explore [] = return []
+  walk v = Node v <$> explore (adjacent v)
+  adjacent v = IntSet.toList (postIntSet v g)
+  discovered v = do new <- gets (not . IntSet.member v)
+                    when new $ modify' (IntSet.insert v)
```
snowleopard: Just curiosity: if you used a non-strict `modify` here, would that lead to poor performance?

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -176,10 +188,26 @@ dfsForest = Typed.dfsForest . Typed.fromAdjacencyIntMap
 -- 'forest' (dfsForestFrom [3] $ 'circuit' [1..5] + 'circuit' [5,4..1]) == 'path' [3,2,1,5,4]
 -- @
 dfsForestFrom :: [Int] -> AdjacencyIntMap -> Forest Int
-dfsForestFrom vs = Typed.dfsForestFrom vs . Typed.fromAdjacencyIntMap
+dfsForestFrom vs g = dfsForestFrom' [ v | v <- vs, hasVertex v g ] g
```
snowleopard: What happens if we do not filter our non-existent vertices? It's obvious that the implementation will break. If it will, I think a comment would be useful here.

snowleopard:
@jitwit Apologies for taking so long to review & merge. I've just left a few more questions.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: Good point! This was an artifact from earlier when the pattern matching was needed

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: The entries correspond to edges where `table ! u = v` means that `v` is `u`'s parent in the search tree. I changed the name to `parent` if that works?

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: I added a remark in the haddocks that seed vertices are ignored if not in the graph. 

If such vertices weren't filtered out, they would be included in the search tree with no children, since `postIntSet` returns an empty set when the vertex isn't in the adjacencymap, I think

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: I tried it out, and the difference seems to be pretty negligible:

[https://github.com/jitwit/alga/blob/bench-dfs/bench/current_results.txt](https://github.com/jitwit/alga/blob/bench-dfs/bench/current_results.txt)

Might lazy be the better choice (even if there's a small hit)?

snowleopard: Let's keep it as is. Thanks for checking!

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
snowleopard: Yes, `parent` is great!

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
snowleopard: I actually wanted to say "It's **not** obvious..." in the above comment :-)

Anyway, thanks for clarifying. No need to do anything else.

snowleopard:
@jitwit I have finally merged the PR. It's great, thank you very much! 

I like the resulting purely functional algorithms for DFS and topological sort. Perhaps you could write a blog post about them?

jitwit:
PS, apologies for the delay! I'm still slowly (but surely) writing out a post about them

snowleopard:
@jitwit No worries, take your time :)

PR #225, opened 2019-07-23
Dijkstra for Labelled.AdjacencyMap.Algorithm

adithyaov:
Added shortest path algorithm for a general labeled graph.

### DIFF
--- a/src/Algebra/Graph/Labelled/AdjacencyMap/Algorithm.hs
+++ b/src/Algebra/Graph/Labelled/AdjacencyMap/Algorithm.hs
@@ -1,0 +1,47 @@
+module Algebra.Graph.Labelled.AdjacencyMap.Algorithm where
+
+import Data.Set (Set)
+import Algebra.Graph.Label
+import Data.Map.Strict (Map, (!))
+import Algebra.Graph.Labelled.AdjacencyMap
+
+import qualified Data.Set as Set
+import qualified Data.Map.Strict as Map
+
+dijkstra :: (Ord a, Ord e, Dioid e) => AdjacencyMap e a -> a -> Map a e
+dijkstra = dijkstra' zero one
+
+dijkstra' :: (Ord a, Ord e, Dioid e) => e -> e -> AdjacencyMap e a -> a -> Map a e
+dijkstra' z o am src = snd (processG processI)
+  where
+    im = adjacencyMap am
+    processI = (Set.singleton (one, src), Map.insert src one . Map.map (const zero) $ im)
+    processG sm@(s, _)
+      | o < z = processS (Set.minView s) sm
+      | otherwise = processS (Set.maxView s) sm
+    processS Nothing sm = sm
+    processS (Just ((_, v1), s)) (_, m) = processG $ relaxV v1 (s, m)
+    relaxV v1 sm =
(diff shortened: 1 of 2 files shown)


src/Algebra/Graph/Labelled/AdjacencyMap/Algorithm.hs
```diff
@@ -0,0 +1,51 @@
...
+import Algebra.Graph.Labelled.AdjacencyMap
+
+import qualified Data.Set as Set
+import qualified Data.Map.Strict as Map
+
+-- TODO: Improve documentation for 'dijkstra'.
+-- | A generic Dijkstra algorithm that relaxes the list of edges
+-- based on the 'Dioid'.
+--
+-- If the 'Dioid' is 'Distance' (negative 'Dioid') the relaxation
```
snowleopard: Are "negative" and "positive" dioids common terms? What is their precise definition?

snowleopard: Couldn't you just use `ShortestDistance` and `WidestPath` from `Algebra.Graph.Label`?

https://hackage.haskell.org/package/algebraic-graphs-0.4/docs/Algebra-Graph-Label.html#t:ShortestPath
https://hackage.haskell.org/package/algebraic-graphs-0.4/docs/Algebra-Graph-Label.html#t:WidestPath

adithyaov: > Are "negative" and "positive" dioids common terms? What is their precise definition?

I apologize, I actually was talking about the order of the underlying semi-ring.
If `1 > 0`, it would be a positive semi-ring. These definitions assume that there is a partial order defined on the set of elements which I believe I did not mention.

> Couldn't you just use ShortestDistance and WidestPath from Algebra.Graph.Label

This is not a generic algorithm but a Dijkstra algorithm. Dijkstra requires a priority heap depending on the order of the semi-ring. Even if one would use `ShortestDistance` the order of relaxation is decided by the heap. 
Using `ShortestDistance` with a max-heap or `WidestPath` with a min-heap would result in improper results. The type of heap is decided using the order of the semi-ring.

A generic single source shortest path algorithm does not require a priority heap but has a higher runtime complexity.

adithyaov: @snowleopard I think one cannot define the order of the relaxation (min-heap vs max-heap) based on the type of Semiring. I believe we should take into account how `+` and `*` function as well.

snowleopard: > These definitions assume that there is a partial order defined on the set of elements which I believe I did not mention.

Right. And I guess you are using `Ord` as this partial order?

Perhaps, you could use the following partial order defined for any idempotent semiring?

```
po x y = (x + y == y)
```

This is the order I use for algebraic graphs: `isSubgraphOf x y = (x + y == y)`.

> Please let me know if you know any good books/papers.

I think you already identified some very good papers on this topic. Not sure there is anything I can add to your list.

snowleopard:
@adithyaov I guess this PR is blocked by #232, right?

adithyaov:
@snowleopard Yes. I'll address #232 and this as soon as possible. I need to write the tests and improve the docs I believe.

PR #226, opened 2019-07-23
Bipartite graphs: a few more properties and graph families

vasalf:


snowleopard:
@vasalf Thanks! Please rebase now that `detectParts` is merged.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -678,15 +687,144 @@ vertexSet = Set.fromAscList . vertexList
...
+-- | The sorted /adjacency list/ of the graph.
+-- Complexity: /O(n + m)/ time and /O(n)/ memory.
+--
+-- @
+-- adjacencyList 'empty'            == []
+-- adjacencyList ('vertices' xs ys) == [ (v, []) | v <- (map Left (sort (nub xs))) ++ (map Right (sort (nub ys))) ]
+-- adjacencyList ('edge' x y)       == [(Left x, [Right y]), (Right y, [Left x])]
+-- adjacencyList                  == 'Algebra.Graph.AdjacencyMap.adjacencyList' . 'fromBipartite'
+-- @
+adjacencyList :: AdjacencyMap a b -> [(Either a b, [Either a b])]
```
snowleopard: This function looks a bit awkward because `Left`s are only paired with `Right`s and vice versa. The user of this function will likely need to use some unsafe constructions to extract the contents. 

Perhaps, `leftAdjacencyList` and `rightAdjacencyList` would be sufficient for the API.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -678,15 +687,144 @@ vertexSet = Set.fromAscList . vertexList
...
+-- This property is heavily used in the examples below.
+--
+-- The 'Show' instance matches the 'Show' instance for lists.
+--
+-- @
+-- 'show' Empty                              == "[]"
+-- 'show' ([1, 2, 3] :: List Int Int)        == "[1,2,3]"
+-- 'show' (Cons 1 (Cons "a" (Cons 3 Empty))) == "[1,\\"a\\",3]"
+-- @
+data List a b = Empty | Cons a (List b a)
```
snowleopard: Perhaps we should follow the convention and use `Nil` instead of `Empty` since we are already using `Cons`. This would also avoid potential confusion with the `empty` graph.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -678,15 +687,144 @@ vertexSet = Set.fromAscList . vertexList
...
+     deriving (Eq, Generic)
+
+instance (Show a, Show b) => Show (List a b) where
```
snowleopard: It looks like this `Show` instance produces invalid Haskell code. Ideally, we should be able to just copy the text printed via `Show` and use it directly as the code that compiles to the original value.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -678,15 +687,144 @@ vertexSet = Set.fromAscList . vertexList
...
+    toList (Cons x xt) = x:(toList xt)
+
+-- | Construct a 'List' of even length from a list of pairs.
+--
+-- @
+-- fromEvenList []                   == 'Empty'
+-- fromEvenList [(1, 2), (3, 4)]     == [1, 2, 3, 4] :: 'List' Int Int
+-- fromEvenList [(1, "a"), (2, "b")] == 'Cons' 1 ('Cons' "a" ('Cons' 2 ('Cons' "b" 'Empty')))
+-- @
+fromEvenList :: [(a, b)] -> List a b
```
snowleopard: It may be better to name these functions `toListEven` and `toListOdd` so that can be found when looking for "toList".

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -678,15 +687,144 @@ vertexSet = Set.fromAscList . vertexList
...
+--
+-- @
+-- fromOddList 1 []                   == 'Cons' 1 'Empty'
+-- fromOddList 1 [(2, 3), (4, 5)]     == [1, 2, 3, 4, 5] :: 'List' Int Int
+-- fromOddList 1 [("a", 2), ("b", 3)] == 'Cons' 1 ('Cons' "a" ('Cons' 2 ('Cons' "b" ('Cons' 3 'Empty'))))
+-- @
+fromOddList :: a -> [(b, a)] -> List a b
+fromOddList x = Cons x . fromEvenList
+
+-- | The /path/ on a even list of vertices.
```
snowleopard: a -> an

snowleopard: I don't think "even list" is a clear term.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -678,15 +687,144 @@ vertexSet = Set.fromAscList . vertexList
...
+-- path 'Empty'                          == 'empty'
+-- path ('Cons' x 'Empty')                 == 'leftVertex' x
+-- path ('Cons' x ('Cons' y 'Empty'))        == 'edge' x y
+-- path ([1, 2, 1, 3] :: 'List' Int Int) == 'star' 1 [2, 3]
+-- path ([1, 2, 3, 1] :: 'List' Int Int) == 'edges' [(1, 2), (3, 2), (3, 1)]
+-- @
+path :: (Ord a, Ord b) => List a b -> AdjacencyMap a b
+path Empty                      = empty
+path (Cons x Empty)             = leftVertex x
+path xs@(Cons _ xt@(Cons _ xr)) = edges $ (zip (odds xs) (odds xt)) ++
```
snowleopard: Redundant parentheses.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -678,15 +687,144 @@ vertexSet = Set.fromAscList . vertexList
...
+                                          (zip (odds xr) (odds xt))
+    where
+        odds :: forall a b. List a b -> [a]
+        odds Empty                = []
+        odds (Cons x Empty)       = [x]
+        odds (Cons x (Cons _ xt)) = x:(odds xt)
```
snowleopard: Redundant parentheses.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -709,6 +847,120 @@ biclique xs ys = let sxs = Set.fromList xs
...
+-- @
+-- stars []                      == 'empty'
+-- stars [(x, [])]               == 'leftVertex' x
+-- stars [(x, [y])]              == 'edge' x y
+-- stars [(x, ys)]               == 'star' x ys
+-- stars                         == 'overlays' . 'map' ('uncurry' 'star')
+-- 'overlay' (stars xs) (stars ys) == stars (xs ++ ys)
+-- @
+stars :: (Ord a, Ord b) => [(a, [b])] -> AdjacencyMap a b
+stars xs = overlay (vertices (map fst xs) []) (edges (concat (map sequenceA xs)))
```
snowleopard: Isn't this just `overlays . map star`?

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -709,6 +847,120 @@ biclique xs ys = let sxs = Set.fromList xs
...
+-- @
+-- boxc x y               ~~ boxc y x
+-- boxc x (boxc y z)      ~~ boxc (boxc x y) z
+-- boxc x ('overlay' y z)   == 'overlay' (boxc x y) (boxc x z)
+-- boxc x ('vertex' ())     ~~ x
+-- boxc x 'empty'           ~~ 'empty'
+-- 'vertexCount' (boxc x y) <= 'vertexCount' x * 'vertexCount' y
+-- 'edgeCount'   (boxc x y) <= 'vertexCount' x * 'edgeCount' y + 'edgeCount' x * 'vertexCount' y
+-- @
+box :: (Ord e, Ord f) =>
```
snowleopard: I think this should be called `boxWith`, with a shorthand `box = boxWith (,) (,) (,) (,)`.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -709,6 +847,120 @@ biclique xs ys = let sxs = Set.fromList xs
...
+       (a -> c -> e) -> (b -> d -> e) -> (a -> d -> f) -> (b -> c -> f) ->
+       AdjacencyMap a b -> AdjacencyMap c d -> AdjacencyMap e f
+box ac bd ad bc (BAM lr1 rl1) (BAM lr2 rl2) = overlay x y
```
snowleopard: Couldn't you reuse `box` from `AdjacencyMap` here?

test/Algebra/Graph/Test/Bipartite/AdjacencyMap.hs
```diff
@@ -27,34 +31,47 @@ import qualified Data.Tuple
...
+type BAII   = AdjacencyMap Int Int
+type BAIS   = AdjacencyMap Int String
+type BAIIII = AdjacencyMap (Int, Int) (Int, Int)
+type BAICIC = AdjacencyMap (Int, Char) (Int, Char)
 
 testBipartiteAdjacencyMap :: IO ()
 testBipartiteAdjacencyMap = do
     putStrLn "\n============ Bipartite.AdjacencyMap.consistent ============"
-    test "consistent empty            == True" $
+    test "consistent empty                     == True" $
```
snowleopard: Why these changes?

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
snowleopard: The shorthand probably deserves to be exported, since it's a lot easier to grasp what it does.

snowleopard:
@vasalf Thank you! I've left some comment. Apologies for the delay.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
vasalf: I guess that the best option is to derive `Show` instance then?

snowleopard: Yes, I think you are right.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
vasalf: It is `overlays . map (uncurry star)` as in the example.

It's not clear whether the time complexity remains the same. It uses `Map.unionsWith (Set.union)` inside and their complexity is undocumented. It seems for me that the only way for them to work for `O(L log(L))` in the worst case is if they merge the smaller set/map into the larger.

This actually might be true, but I can't be sure.

The complexity of this function is not critical, but this implementation doesn't seem too complex.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
vasalf: Thanks! Much better this way!

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
snowleopard: Could you compare the performance of simple `overlays . map (uncurry star)` and your version? I always prefer simple solutions to less simple unless there is a good performance reason.

snowleopard: If you find that yours is indeed significantly faster (which is not obvious, as you say), then it's worth adding a comment to the implementation, explaining that it is chosen over `overlays . map (uncurry star)` because it gives `Nx` speed up.

vasalf: Well, I didn't realize there's `Map.fromListWith (Set.union)` in `edges`, too.

vasalf: The benchmark says that there is some difference on some tests. The suggested `overlays` implementation works 1.2 times faster on tests where all edges in the input are distinct and the original implementation works 2 times faster on tests where all input stars are equal (with different order of edges, though).

https://gist.github.com/vasalf/9a77006df65ec67be682386fe0a8685d

snowleopard: In this case, let's go with the simpler implementation via `overlays . map (uncurry star)`. There are plenty of places like this where we could do some optimisations, but that would derail this PR. After we merge this basic implementation, you are welcome to send more patches for optimising individual functions!

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -709,6 +817,113 @@ biclique xs ys = let sxs = Set.fromList xs
...
+-- @
+-- box x y               ~~ box y x
+-- box x (box y z)       ~~ box (box x y) z
+-- box x ('overlay' y z)   == 'overlay' (box x y) (box z)
+-- box x ('vertex' ())     ~~ x
+-- box x 'empty'           ~~ 'empty'
+-- 'vertexCount' (box x y) <= 'vertexCount' x * 'vertexCount' y
+-- 'edgeCount'   (box x y) <= 'vertexCount' x * 'edgeCount' y + 'edgeCount' x * 'vertexCount' y
+-- @
+boxWith :: forall a b c d e f. (Ord a, Ord b, Ord c, Ord d, Ord e, Ord f) =>
```
snowleopard: Can we get rid of `forall a b c d e f` in this type signature? It's best to avoid such top-level `forall`s since they sneak into Haddock. In this case, I think it can just be dropped.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -709,6 +817,113 @@ biclique xs ys = let sxs = Set.fromList xs
...
+-- | Compute the /Cartesian product/ of bipartite graphs. The vertices are
+-- sorted into parts using given combinators. Values returned by combinators
+-- are not necessarily distinct.
+-- Complexity: /O(s1 * s2 * log(s1 * s2))/ time and /O(s1 * s2)/ memory, where
+-- /s1/ and /s2/ are the sizes of the given graphs.
+--
+-- There's also a shorthand 'box' for @boxWith (,) (,) (,) (,)@.
+--
+-- @
+-- 'box' ('path' [0,1]) ('path' [\'a\',\'b\']) == 'edges' [ ((0,\'a\'), (0,\'b\'))
```
snowleopard: It's strange to have `box` examples attached to a different function (`boxWith`) -- I suggest simply moving these examples to `box`, and adjusting the comments accordingly. Also, move `box` in front of more complicated `boxWith` both here and in the export list.

snowleopard:
@vasalf Apologies for taking so long to review and merge. I've left a couple of more suggestions.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -678,15 +686,115 @@ vertexSet = Set.fromAscList . vertexList
...
+-- | Construct a 'List' of even length from a list of pairs.
+--
+-- @
+-- fromListEven []                   == 'Nil'
+-- fromListEven [(1, 2), (3, 4)]     == [1, 2, 3, 4] :: 'List' Int Int
+-- fromListEven [(1, "a"), (2, "b")] == 'Cons' 1 ('Cons' "a" ('Cons' 2 ('Cons' "b" 'Nil')))
+-- @
+fromListEven :: [(a, b)] -> List a b
+fromListEven []          = Nil
+fromListEven ((x, y):xt) = Cons x (Cons y (fromListEven xt))
```
nobrakal: I think we can write `fromListEven` with `foldr`:
```suggestion
fromListEven = foldr (\(x,y) -> Cons x . Cons y) Nil
```

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -709,6 +817,112 @@ biclique xs ys = let sxs = Set.fromList xs
                   in BAM (Map.fromSet (const sys) sxs)
                          (Map.fromSet (const sxs) sys)
 
+-- | The /star/ formed by a center vertex connected to a list of leaves.
+-- Complexity: /O(L log(L))/ time, where /L/ is the length of the given list.
```
nobrakal: I think we can add the times sign to be coherent with the rest of the library (here and in others comments).

```suggestion
-- Complexity: /O(L * log(L))/ time, where /L/ is the length of the given list.
```

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -678,15 +686,115 @@ vertexSet = Set.fromAscList . vertexList
...
+-- 'show' (Cons 1 (Cons \"a\" (Cons 3 Nil))) == \"Cons 1 (Cons \\"a\\" (Cons 3 Nil))\"
+-- @
+data List a b = Nil | Cons a (List b a)
+     deriving (Eq, Show, Generic)
+
+instance IsList (List a a) where
+    type Item (List a a) = a
+
+    fromList []     = Nil
+    fromList (x:xt) = Cons x (fromList xt)
```
nobrakal: We can use `foldr` here:
```suggestion
    fromList = foldr Cons Nil
```

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
```diff
@@ -678,15 +686,115 @@ vertexSet = Set.fromAscList . vertexList
...
+-- path 'Nil'                            == 'empty'
+-- path ('Cons' x 'Nil')                   == 'leftVertex' x
+-- path ('Cons' x ('Cons' y 'Nil'))          == 'edge' x y
+-- path ([1, 2, 1, 3] :: 'List' Int Int) == 'star' 1 [2, 3]
+-- path ([1, 2, 3, 1] :: 'List' Int Int) == 'edges' [(1, 2), (3, 2), (3, 1)]
+-- @
+path :: (Ord a, Ord b) => List a b -> AdjacencyMap a b
+path Nil                        = empty
+path (Cons x Nil)               = leftVertex x
+path xs@(Cons _ xt@(Cons _ xr)) = edges $ zip (odds xs) (odds xt) ++
```
nobrakal: You can factorize `odds xt` here.

src/Algebra/Graph/Bipartite/AdjacencyMap.hs
snowleopard: Yes, we use `*` in the rest of the library.

snowleopard:
@vasalf Just a reminder: I'm awaiting the last few fixes following the comments by @nobrakal.

src/Algebra/Graph/Bipartite/Undirected/AdjacencyMap.hs
```diff
@@ -50,11 +57,12 @@ import Control.Monad
 import Control.Monad.Trans.Maybe
 import Control.Monad.State
 import Data.Either
-import Data.Foldable
+import Data.Foldable (asum)
```
vasalf: For history: we only use `asum` from `Data.Foldable`, and `Data.Foldable.toList` conflicts with `GHC.Exts.toList`. Might have been as well

```
import Data.Foldable hiding toList
```

src/Algebra/Graph/Bipartite/Undirected/AdjacencyMap.hs
```diff
@@ -35,9 +35,16 @@ module Algebra.Graph.Bipartite.Undirected.AdjacencyMap (
...
+    leftAdjacencyList, rightAdjacencyList,
+
+    -- * Alternating lists
+    List(..), fromListEven, fromListOdd,
 
     -- * Standard families of graphs
-    circuit, biclique,
+    path, circuit, biclique, star, stars, mesh,
+
+    -- * Graph transformations
```
snowleopard: This section is usually called "Graph composition".

src/Algebra/Graph/Bipartite/Undirected/AdjacencyMap.hs
```diff
@@ -682,6 +690,106 @@ vertexSet = Set.fromAscList . vertexList
 edgeSet :: (Ord a, Ord b) => AdjacencyMap a b -> Set (a, b)
 edgeSet = Set.fromAscList . edgeList
 
+-- | The sorted /adjacency list/ of the left part of the graph.
+-- Complexity: /O(n + m)/ time and /O(n)/ memory.
```
snowleopard: Pretty sure this has O(n + m) memory complexity.

src/Algebra/Graph/Bipartite/Undirected/AdjacencyMap.hs
```diff
@@ -682,6 +690,106 @@ vertexSet = Set.fromAscList . vertexList
...
+-- leftAdjacencyList ('vertices' [] xs) == []
+-- leftAdjacencyList ('vertices' xs []) == [ (x, []) | x <- nub ('sort' xs) ]
+-- leftAdjacencyList ('edge' x y)       == [(x, [y])]
+-- leftAdjacencyList ('star' x ys)      == [(x, 'nub' ('sort' ys))]
+-- @
+leftAdjacencyList :: AdjacencyMap a b -> [(a, [b])]
+leftAdjacencyList (BAM lr _) = [ (v, Set.toAscList us) | (v, us) <- Map.toAscList lr ]
+
+-- | The sorted /adjacency list/ of the right part of the graph.
+-- Complexity: /O(n + m)/ time and /O(n)/ memory.
```
snowleopard: Same as above.

snowleopard:
@vasalf I had a brief look and I think everything looks good. I merged and may do a minor revision pass soon.

Many thanks for the PR and for the final fixes!

PR #227, opened 2019-07-27
bfs complexity

jitwit:
I mirrored `Data.IntMap` using `W` for `AdjacencyIntMap`s. Also, not sure why I thought the space complexity was `O(m+n)`!

Should there be any further adjustments to the documentation?

### DIFF
--- a/src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
+++ b/src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
@@ -66,9 +66,8 @@ bfsForest g = bfsForestFrom' (vertexList g) g
 -- | Like 'bfsForest', but the traversal is seeded by a list of
 --   vertices. Vertices not in the graph are ignored.
 --
---   Let /L/ be the number of seed vertices and /W/ the number bits in
---   a machine word. Complexity: /O((L+m)*min(n,W))/ time and /O(n)/
---   space.
+--   Let /L/ be the number of seed vertices. Complexity:
+--   /O((L+m)*min(n,W))/ time and /O(n)/ space.
 --
 -- @
 -- 'forest' (bfsForestFrom [1,2] $ 'edge' 1 2)      == 'vertices' [1,2]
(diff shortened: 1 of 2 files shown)


src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -35,8 +35,12 @@ import qualified Data.Graph.Typed   as Typed
 import qualified Data.IntMap.Strict as IntMap
 import qualified Data.IntSet        as IntSet
 
--- | Compute the forest of a graph's vertices in breadth first order. Complexity:
--- /O(n+m*log n)/ time and /O(n+m)/ space.
+-- | Compute the /breadth-first search/ forest of a graph, where
+--   adjacent vertices are expanded smallest to biggest with respect
```
snowleopard: I think "are expanded smallest to biggest" could be phrased a bit more precisely as "are explored in the increasing order".

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -35,8 +35,12 @@ import qualified Data.Graph.Typed   as Typed
...
+--   to their 'Ord' instance.
+-- 
+--   Let /W/ be the number of bits in a machine word. Complexity:
+--   /O((n+m)*W)/ time and /O(n)/ space.
```
snowleopard: In `IntSet` docs, the factor is defined as `min(n,64)` on a 64-bit machine, so this seems a bit imprecise.

Perhaps, we could say "`W = min(n,64)` on a 64-bit machine"?

snowleopard:
@jitwit Thanks for revising the docs! I've left a couple of minor suggestions.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: Yeah. To me, it's awkward that `O(64)` is really `O(1)` and I'm not sure how `O(min(n,64))` makes sense? Ie, for all `n >= 64`, `min(n,64) = 64`. I'm happy to stick to the conventions from `IntSet` in any case

jitwit: If `W` might still work, maybe "word size" instead of "number of bits in a machine word" would be sufficiently better? 

Simply going with `O(log(n))` might be better anyway?

snowleopard: I suggest sticking to the notation used in the standard libraries since people are already familiar with it. We could even say somewhere at the top of the module that we are following the notation from standard libraries since we are using them in our implementaion.

snowleopard:
@jitwit Thank you for the fix, merged!

PR #228, opened 2019-08-04
Rules system optimization

nobrakal:
Hi,

I have been working recently with fusion, and it made me think about some optimizations that can be made to the current system for `Algebra.Graph`.

This:

1. Change to good producers `vertices`, `edges` (ie. they are now expressed directly with `buildR`)... Meaning that `foldg e v o c . vertices xs` is optimized to a single list traversal _without a custom a rule_.

2. `bind` is also a good producer, so I added its `buildR` definition directly. 

3. `star` is now (in addition to be a good producer of graph), a good consumer of lists (meaning that `star . map f` is also optimized`).

4. The `foldr1Safe` rule is now unneeded, I removed it.

5. I added `INLINE` pragmas to functions expressed with `foldg`.

All the previous tests are passing, and the benchmarking suite is happy.

### Questions

I have some questions:

* Can we expose `buildR` ? Maybe moving it in `Algebra.Graph.Internal`, but exposing such a function allows the end-user to define its own function using `buildR` and benefit from fusion using `foldg/buildR`.

* The whole thing needs documentation. I like the "good producer" / "good consumer" vocabulary (stolen from the GHC manual). A good consumer is a function that can be fused with a good producer, and a good producer is a function that can be fused with a good consumer. Then we can tag all concerned functions.

* Because of the use of `buildR`, we _need_ rewrite rules to be enabled (either, some `foldg Empty Vertex Overlay Connect` will not be optimized). There is no problem for the library itself, but we should add a warning about it somewhere.


snowleopard:
@nobrakal Thanks for the PR! I like the overall goal: introducing the vocabulary of "good" producers and consumers and making sure that all exported functions are as good as possible!

It would, of course, be nice if users could write their own good producers and consumers.

My concern is that users who don't know (and perhaps don't want to know) about rewrite rules will be exposed to them if we start exporting functions like `buildR`. It would be great if the user could write simple code like `overlays . map vertex` and get an efficient implementation (i.e. a good producer).

I would have hoped for the notion of "goodness" to be composable, i.e. if `map` preservers goodness in the world of lists, and `overlays` is a good consumer of lists and a good producer of graphs, then their composition is automatically a good consumer of lists and a good producer of graphs. Is there any hope for achieving something like this? 

If we make goodness compositional we will never run out of goodness! :)

snowleopard:
P.S.: I guess what I'm after is a "good category" :-) The identity function `id` is hopefully both a good consumer and producer, and I'd like `.` to preserve goodness too!

nobrakal:
I added some comments, many functions are good producers and good consumers :D (`biclique` even ended to be a good consumer of _both_ its arguments and a good producer !)

> My concern is that users who don't know (and perhaps don't want to know) about rewrite rules ...

That why I propose to export it from `Algebra.Graph.Internal` (like `build` is exported from `GHC.Base` for lists). Then only people that are searching for it will find it :)

> I would have hoped for the notion of "goodness" to be composable ...

 I think that, strictly speaking, the notion of "goodness" is composable. The only problem here for me is that `overlays` is a good consumer of lists, and not of list _of graphs_. An other way to say is: because `buildR` abstracts all the constructors of graphs, thus any good producer should be expressed "in one go". For `overlays . map vertex` to be a good producer, `overlays` should know the previous `map vertex` to abstract the `Vertex` constructor.

src/Algebra/Graph.hs
```diff
@@ -209,6 +213,7 @@ not under our control. We therefore choose the safe and more explicit path of
 creating our own intermediate functions for guiding rewrite rules when needed.
 -}
 
+-- | 'fmap' is a good consumer and a good producer.
```
snowleopard: Since these comments are already attached to specific functions, I suggest writing simply "A good consumer and producer", "A good consumer of both arguments", etc.

snowleopard:
> I added some comments, many functions are good producers and good consumers :D (`biclique` even ended to be a good consumer of both its arguments and a good producer !)

@nobrakal Great! :)

> That why I propose to export it from `Algebra.Graph.Internal` (like `build` is exported from `GHC.Base` for lists). Then only people that are searching for it will find it :)

OK, let's do that. I'd prefer to export it with a simpler name `build` (instead of `buildR`).

> The only problem here for me is that overlays is a good consumer of lists, and not of list of graphs. 

Oh, I see. That's more tricky than I thought -- there are layers of goodness!

src/Algebra/Graph.hs
nobrakal: For class functions, I think its worth mentioning which function is a good consumer/producer.

Otherwise, you are right, I will change this.

nobrakal:
> OK, let's do that. I'd prefer to export it with a simpler name build or buildg (instead of buildR).

I said it before trying, but `buildR` needs to know the graphs constructors, so we cannot put it in `Algebra.Graph.Internal`.

snowleopard:
@nobrakal In this case, let's export it as `buildg` from `Algebra.Graph`.

src/Algebra/Graph.hs
```diff
@@ -1285,29 +1288,37 @@ sparsifyKL n graph = KL.buildG (1, next - 1) ((n + 1, n + 2) : Exts.toList (res
...
 
-buildR :: Foldg a -> Graph a
-buildR g = g Empty Vertex Overlay Connect
-{-# INLINE [1] buildR #-}
+-- | Replace abstracted graphs constructors by concrete ones.
+--
+-- Functions expressed with 'buildg' are good producers.
+--
+-- @
+-- buildg g = g 'Empty' 'Vertex' 'Overlay' 'Connect'
```
snowleopard: Could you add some more examples?

I guess you could add examples like `empty = buildg (\e _ _ _ -> e)` etc.

src/Algebra/Graph.hs
```diff
@@ -1285,29 +1288,37 @@ sparsifyKL n graph = KL.buildG (1, next - 1) ((n + 1, n + 2) : Exts.toList (res
...
 
 * composeR and matchR are here to remember the original function after applying
-  a "buildR/f" rule. These functions are higher-order functions and therefore
+  a "buildg/f" rule. These functions are higher-order functions and therefore
   benefit from inlining in the final phase.
 
 * The "bindR/bindR" rule optimises compositions of multiple bindR's.
 -}
 
+-- | A function abstracting graphs constructors.
```
snowleopard: I think it's worth providing some more explanation. For example, pointing out that it's actually the type of `foldg`, right? I guess it is also related dto the Boehm-Berarducci encoding of algebraic graphs:

https://hackage.haskell.org/package/algebraic-graphs-0.4/docs/Algebra-Graph-Fold.html

Note, however, that `Fold` will be deleted in the next release... Maybe I shouldn't have done that :)

nobrakal: > I guess it is also related dto the Boehm-Berarducci encoding of algebraic graphs

Ah, yes indeed. If you are going to remove it, I don't know if it worth adding a reference.

I am not fond of "this is foldg's type", because, for me, it hides the use-case of building a structure (ie, one can express `vertices` with `buildg`).

src/Algebra/Graph.hs
nobrakal: Done, I also added corresponding tests

nobrakal:
I addressed your comments. More generally, I don't know the amount of documentation we need to add... There are many things to say :)

src/Algebra/Graph.hs
```diff
@@ -1187,68 +1288,76 @@ sparsifyKL n graph = KL.buildG (1, next - 1) ((n + 1, n + 2) : Exts.toList (res
...
+-- The Rank-2 type is used to ensure that one can replace abstracted graphs constructors
+-- (the function arguments) by the arguments of a later 'foldg' operation.
+--
+-- See 'buildg'.
 type Foldg a = forall b. b -> (a -> b) -> (b -> b -> b) -> (b -> b -> b) -> b
 
-buildR :: Foldg a -> Graph a
-buildR g = g Empty Vertex Overlay Connect
-{-# INLINE [1] buildR #-}
+-- | Replace abstracted graphs constructors by concrete ones.
```
snowleopard:
```suggestion
-- | Build a graph given an interpretation of the four graph construction primitives 'empty',
-- 'vertex', 'overlay' and 'connect', in this order. See examples for further clarification.
```

snowleopard:
@nobrakal What about removing the `Foldg` synonym? I've suggested a comment that might work if you simply inline `Foldg` into the definition of `buildg`.

src/Algebra/Graph.hs
```diff
@@ -47,7 +47,10 @@ module Algebra.Graph (
     compose, box,
 
     -- * Context
-    Context (..), context
+    Context (..), context,
+
+    -- * Deforestation
+    buildg
```
snowleopard: I suggest placing it into the same category as `foldg`: these two functions are related and it would be nice if they appear in the documentation together.

src/Algebra/Graph.hs
```diff
@@ -1187,68 +1288,69 @@ sparsifyKL n graph = KL.buildG (1, next - 1) ((n + 1, n + 2) : Exts.toList (res
...
-"foldg/buildR" forall e v o c (g :: Foldg a).
-    foldg e v o c (buildR g) = g e v o c
+-- Fuse a foldg followed by a buildg
+"foldg/buildg" forall e v o c (g :: forall b. b -> (a -> b) -> (b -> b -> b) -> (b -> b -> b) -> b).
+    foldg e v o c (buildg g) = g e v o c
 
 -- Fuse composeR's. This occurs when two adjacent 'bindR' were rewritted into
--- their buildR form.
+-- their buildg form.
 "bindR/bindR" forall c f g.
```
snowleopard: Why is this rule called `bindR/bindR`? I guess it should be `composeR/composeR`.

src/Algebra/Graph.hs
```diff
@@ -268,13 +279,15 @@ apR :: Graph (a -> b) -> Graph a -> Graph b
...
 
+-- | `>>=` is a good consumer and producer.
 instance Monad Graph where
     return = pure
     (>>=)  = bindR
 
 bindR :: Graph a -> (a -> Graph b) -> Graph b
-bindR g f = foldg Empty f Overlay Connect g
-{-# INLINE [0] bindR #-}
+bindR g f = buildg $ \e v o c ->
```
snowleopard: Do we still need `bindR`? I couldn't find any remaining rules for it.

src/Algebra/Graph.hs
```diff
@@ -1187,68 +1288,69 @@ sparsifyKL n graph = KL.buildG (1, next - 1) ((n + 1, n + 2) : Exts.toList (res
...
+buildg f = f Empty Vertex Overlay Connect
+{-# INLINE [1] buildg #-}
 
 composeR :: (b -> c) -> (a -> b) -> a -> c
 composeR = (.)
-{-# INLINE [0] composeR #-}
+{-# INLINE [1] composeR #-}
 
 matchR :: b -> (a -> b) -> (a -> Bool) -> a -> b
 matchR e v p = \x -> if p x then v x else e
```
snowleopard: `matchR` doesn't seem to be used in the LHS of a rewrite rule either.

nobrakal: You can find it in the `"graph/induce"` rule

src/Algebra/Graph.hs
nobrakal: Indeed we can remove them, the same apply to `apR` and `fmapR`, they were leftovers.

src/Algebra/Graph.hs
```diff
@@ -47,7 +47,7 @@ module Algebra.Graph (
     compose, box,
 
     -- * Context
-    Context (..), context
+    Context (..), context,
```
snowleopard: Looks like an accidental change.

src/Algebra/Graph.hs
```diff
@@ -173,12 +173,16 @@ compatible with 'overlay' and 'connect' operations:
...
+Deforestation (fusion) is implemented for some functions in this module. This means
+that when a function tagged as a \"good producer\" is composed with a \"good consumer\",
+the intermediate structure will not be built.
 -}
 data Graph a = Empty
              | Vertex a
              | Overlay (Graph a) (Graph a)
              | Connect (Graph a) (Graph a)
-             deriving (Show, Generic)
+             deriving (Generic)
```
snowleopard: I think it's best to continue using derived `Show`. Performance is not really that important for `Show` because it is not supposed to be used for large graphs anyway. Using derived `Show` allows us to avoid the extra complexity that comes with the custom implementation.

src/Algebra/Graph.hs
```diff
@@ -232,9 +241,11 @@ instance Num a => Num (Graph a) where
     abs         = id
     negate      = id
 
+-- | `==` is a good consumer of both arguments.
```
snowleopard: You use backticks in a few places and I think Haddock doesn't support them -- please double-check the generated docs. I think we need to switch to quotes `'`.

src/Algebra/Graph.hs
```diff
@@ -423,21 +442,28 @@ overlays = fromMaybe empty . foldr1Safe overlay
...
 -- the provided functions to the leaves and internal nodes of the expression.
 -- The order of arguments is: empty, vertex, overlay and connect.
 -- Complexity: /O(s)/ applications of given functions. As an example, the
 -- complexity of 'size' is /O(s)/, since all functions have cost /O(1)/.
 --
+-- Good consumer.
+--
 -- @
--- foldg 'empty' 'vertex'        'overlay' 'connect'        == id
+-- foldg 'empty' 'vertex'        'overlay' 'connect'       == id
```
snowleopard: I think the alignment was fine in the Haddock:

https://hackage.haskell.org/package/algebraic-graphs-0.4/docs/Algebra-Graph.html#v:foldg

Are these changes accidental?

src/Algebra/Graph.hs
snowleopard: Indeed, thanks!

snowleopard:
@nobrakal Apologies for taking so long to review and merge. I've left a few more comments/questions.

Is it possible to test the guarantees of being a good producer/consumer? I'm worried that we might accidentally break these guarantees at some point later without realising it.

src/Algebra/Graph.hs
nobrakal: They are not, the alignment was buggy on master for me. Could you check? Maybe my haddock has a problem.

src/Algebra/Graph.hs
nobrakal: Backquotes seem to be the syntax for infix operators. This is producing a good output for me.

nobrakal:
> Is it possible to test the guarantees of being a good producer/consumer? I'm worried that we might accidentally break these guarantees at some point later without realising it.

We can a write a `inspection-testing` test for each function indeed, but this will make a huge testsuite

snowleopard:
@nobrakal I would very much prefer to include tests for such documentation claims, because they are very likely to break otherwise.

Wouldn't one test per claim be sufficient?

src/Algebra/Graph.hs
snowleopard: Hmm, that's strange, let me check.

snowleopard:
By the way, while we're tweaking rewrite rules, can we also remove this version constraint?

`inspection-testing >= 0.4 && < 0.4.2`

nobrakal:
Hi there, sorry for taking so long to answer.

> Wouldn't one test per claim be sufficient?

Indeed, but it will need more than 50 hand-made tests (the only test I see is to manually make the fusion and compare using `inspection-testing` for each claim). I can do the work, but I prefer to be sure we need it :)
Note that being a good consumer is just "consume the graph argument using `foldg`" and being a good producer is just "abstracting graphs constructors with `buildg`". They are pretty simple requirements but I don't see an easy way to test them.

> By the way, while we're tweaking rewrite rules, can we also remove this version constraint?

We just need to avoid version 0.4.2 which had a bug.

snowleopard:
@nobrakal Yes, adding 50 tests manually doesn't sound like fun. However, for every function in question, we already have a bunch of tests, so adding 1-2 new tests doesn't seem like a big deal. Thorough testing is pretty much always tedious, but it does bring benefits in the long term.

I can easily imagine myself accepting a PR two years from now that accidentally breaks the good consumer/producer property. A test would save the day :)

nobrakal:
@snowleopard Ok, I will add them.

I just rewrite old tests into more systematic ones. Do they seems OK to you ? If yes, I will follow with all the possible functions :)

By the way, I just hit a little problem with `<*>`, I will try to figure why it is not a good consumer of its second argument.

snowleopard:
@nobrakal Great, thank you! Yes, the tests look good.

> By the way, I just hit a little problem with `<*>`, I will try to figure why it is not a good consumer of its second argument.

Aha! So, tests are already proving useful :-)

nobrakal:
I added some tests for "simple" functions, I will investigate others as soon as possible :)

snowleopard:
@nobrakal Glad to see this PR is back to life! :-)

After having looked at the tests, I'm wondering: is it possible to get rid of duplication of many of the definitions from `Algebra.Graph`? Can we express the notions of "good producer" and "good consumer" in a less mechanical manner?

The README of the `inspection-testing` library lists the following as one of the features:

> checking the absence of a certain type (useful in the context of list or stream fusion)

So, what about checking that `size` is a good consumer and `biclique` is a good producer by applying one to another and testing that no `Graph` constructors remain?

nobrakal:
Yes, indeed you are right, I just tested and something like

```Haskell
overlaysC :: Build (Graph a) -> Graph a
overlaysC xs = overlays (build xs)

inspect $ 'overlaysC `hasNoType` ''[]
```

Works like a charm :)

There is two little problems:

1. This will not work with graph homomorphisms (the graph type will appear anyway since it is a good producer and a good consumer).
2. `foldg e v o c (overlays xs)` for example, or others functions using graph internally cannot be tested this way, since the graph type appears inside the function even after fusion.

I think I can add `hasNoType` where possible, and use old ugly duplication where it is not possible. Does it sounds good to you ?

nobrakal:
As I had some time, I rewrote concerned tests using your suggestion, the commit can be reverted without too many trouble I think.
I also added more tests (including the one for `hasEdge` !). We only now need to ensure that all claims are correctly tested, but I think the biggest part of the work is now made.

snowleopard:
@nobrakal This is great, the tests look much better!

For functions that consume and produce graphs, e.g. `transpose`, perhaps we could check that their composition with `buildg` and `foldg` has no `Graph` constructors? I think if `transpose` is indeed a good consumer and a good producer, then `\e v o c g -> foldg e v o c (transpose (buildg g))` should have no graph constructors left, i.e. the result should be just `g e v o (flip c)` :-)

snowleopard:
In fact, I think this "sandwich" approach works for any function that claims to be both a good consumer of X and a good producer of Y: i.e. we can just sandwich it between a producer of X and a consumer of Y and see what happens.

nobrakal:
@snowleopard  You are right, the sandwich approach would work. I can only argue that it is merging two tests for two different things into a single test, but I don't know if this is a problem.

snowleopard:
> I can only argue that it is merging two tests for two different things into a single test, but I don't know if this is a problem.

Intuitively, only a function that is both good consumer and a good producer can pass the sandwich check. Do you agree? I don't see a simple formal argument :-)

nobrakal:
Ok, I agree this is simpler anyway :)

Just a question about `induceJust`, should I add the same set of rules than `induce` to mimic its fusion (since it is not a good producer naturally) ?

snowleopard:
@nobrakal Can't `induce` and `induceJust` be turned into good producers by using `buildg`? I think it's possible, although we might need to add `Maybe` to pattern match against `Nothing` instead of `Empty`. Do you see what I mean?

nobrakal:
@snowleopard I am not sure why, I was sure to have tested this solution and rejected it for efficiency reason. Anyway, I just gave it a try, and all seems to work :) I will wait to see if the benchmarking suite is showing terrible results.

nobrakal:
`removeVertex` testing `induce is perfect!

Inspecting the benchmarking results, I just discovered a little flow I introduced that was dropping the performances of `edges` and corrected it in the last commit.

src/Algebra/Graph.hs
```diff
@@ -232,49 +234,51 @@ instance Num a => Num (Graph a) where
...
 eqR :: Ord a => Graph a -> Graph a -> Bool
 eqR x y = toAdjacencyMap x == toAdjacencyMap y
-{-# NOINLINE [1] eqR #-}
+{-# INLINE [2] eqR #-}
 {-# RULES "eqR/Int" eqR = eqIntR #-}
 
 -- Like 'eqR' but specialised for graphs with vertices of type 'Int'.
 eqIntR :: Graph Int -> Graph Int -> Bool
-eqIntR x y = toAdjacencyIntMap x == toAdjacencyIntMap y
+eqIntR = \x y -> toAdjacencyIntMap x == toAdjacencyIntMap y
```
snowleopard: Is this change accidental? If not, what does it achieve?

src/Algebra/Graph.hs
```diff
@@ -232,49 +234,51 @@ instance Num a => Num (Graph a) where
...
 ordR :: Ord a => Graph a -> Graph a -> Ordering
 ordR x y = compare (toAdjacencyMap x) (toAdjacencyMap y)
-{-# NOINLINE [1] ordR #-}
+{-# INLINE [2] ordR #-}
 {-# RULES "ordR/Int" ordR = ordIntR #-}
 
 -- Like 'ordR' but specialised for graphs with vertices of type 'Int'.
 ordIntR :: Graph Int -> Graph Int -> Ordering
-ordIntR x y = compare (toAdjacencyIntMap x) (toAdjacencyIntMap y)
+ordIntR = \x y -> compare (toAdjacencyIntMap x) (toAdjacencyIntMap y)
```
snowleopard: Same question here.

src/Algebra/Graph.hs
```diff
@@ -232,49 +234,51 @@ instance Num a => Num (Graph a) where
...
+{-# INLINE ordIntR #-}
 
+-- | `<*>` is a good consumer of its first agument and producer.
```
snowleopard: Is it possible to make it a good consumer of both arguments? If not, is there a simple argument why this can't be done?

snowleopard: (If there is such an argument, I think it'd be useful to add it above as a comment.)

src/Algebra/Graph.hs
```diff
@@ -634,17 +673,20 @@ edgeCountIntR = AIM.edgeCount . toAdjacencyIntMap
 -- @
 vertexList :: Ord a => Graph a -> [a]
 vertexList = Set.toAscList . vertexSet
-{-# INLINE [1] vertexList #-}
+{-# INLINE [2] vertexList #-}
```
snowleopard: I think `vertexList` and `edgeList` are good consumers too.

src/Algebra/Graph.hs
```diff
@@ -715,44 +768,58 @@ adjacencyList = AM.adjacencyList . toAdjacencyMap
...
 -- Like @toAdjacencyMap@ but specialised for graphs with vertices of type 'Int'.
 toAdjacencyIntMap :: Graph Int -> AIM.AdjacencyIntMap
 toAdjacencyIntMap = foldg AIM.empty AIM.vertex AIM.overlay AIM.connect
+{-# INLINE toAdjacencyIntMap #-}
 
 -- | The /path/ on a list of vertices.
 -- Complexity: /O(L)/ time, memory and size, where /L/ is the length of the
 -- given list.
 --
+-- Good producer.
```
snowleopard: Can `path` and `circuit` be made good consumers of lists?

src/Algebra/Graph.hs
```diff
@@ -1051,6 +1153,8 @@ induceJust = foldg Empty (maybe Empty Vertex) (k Overlay) (k Connect)
 -- Complexity: the function performs /O(s)/ graph comparisons. It is guaranteed
 -- that the size of the result does not exceed the size of the given expression.
 --
+-- Good consumer.
```
snowleopard: Let's make it a good producer too (seems straightforward).

src/Algebra/Graph.hs
```diff
@@ -1086,6 +1191,8 @@ simple op x y
 -- quadratic, i.e. /m = O(m1 * m2)/, but the algebraic representation requires
 -- only /O(m1 + m2)/ operations to list them.
 --
+-- Good consumer of both arguments.
```
snowleopard: Does it happen to also be a good producer? I'd hope that `overlays` of `biclique`s can fuse nicely.

snowleopard: (Perhaps the list of bicliques does disappear, but not the bicliques themselves...)

src/Algebra/Graph.hs
```diff
@@ -1202,81 +1310,63 @@ sparsifyKL n graph = KL.buildG (1, next - 1) ((n + 1, n + 2) : Exts.toList (res
...
-* The "bindR/bindR" rule optimises compositions of multiple bindR's.
+* The "composeR/composeR" rule optimises compositions of multiple composeR's.
 -}
 
-type Foldg a = forall b. b -> (a -> b) -> (b -> b -> b) -> (b -> b -> b) -> b
-
-buildR :: Foldg a -> Graph a
-buildR g = g Empty Vertex Overlay Connect
-{-# INLINE [1] buildR #-}
+-- | Build a graph given an interpretation of the four graph construction primitives 'empty',
```
snowleopard: Let's move this definition up, after `foldg`, to match the export order (just for ease of navigation).

src/Algebra/Graph.hs
```diff
@@ -1202,81 +1310,63 @@ sparsifyKL n graph = KL.buildG (1, next - 1) ((n + 1, n + 2) : Exts.toList (res
...
--- Fuse composeR's. This occurs when two adjacent 'bindR' were rewritted into
--- their buildR form.
-"bindR/bindR" forall c f g.
-    composeR (composeR c f) g = composeR c (f.g)
+-- Fuse composeR's (from bind's definition).
+"composeR/composeR" forall c f g.
+    composeR (composeR c f) g = composeR c (f . g)
 
--- Rewrite identity (which can appear in the rewriting of bindR) to a much efficient one
+-- Rewrite identity (which can appear in the inlining of 'buildg') to a much efficient one.
```
snowleopard: much -> more/much more

src/Algebra/Graph/Internal.hs
```diff
@@ -108,14 +108,7 @@ connectFoci x y = Focus (ok x || ok y) (xs <> is y) (os x <> ys) (vs x <> vs y)
 -- | A safe version of 'foldr1'.
 foldr1Safe :: (a -> a -> a) -> [a] -> Maybe a
```
snowleopard: I think this function can actually be called `sconcat`, i.e. concatenating a list of semigroup values :)

src/Algebra/Graph.hs
```diff
@@ -381,25 +388,31 @@ connect = Connect
 -- 'vertexSet'   . vertices == Set.'Set.fromList'
 -- @
 vertices :: [a] -> Graph a
-vertices = overlays . map vertex
```
snowleopard: Any hope (perhaps, just in future) to be able to implement `vertices` as `overlays . map vertex` preserving the goodness? :)

test/Algebra/Graph/Test/RewriteRules.hs
```diff
@@ -13,111 +14,352 @@ module Algebra.Graph.Test.RewriteRules where
...
 
-starT1, starTR :: a -> [a] -> Graph a
-starT1 x = transpose . star x
-starTR a [] = vertex a
-starTR a xs = connect (vertices xs) (vertex a)
+inspect $ 'transposeCP `hasNoType` ''Graph
 
-inspect $ 'starT1 === 'starTR
+-- simplify
+simple :: Eq g => (g -> g -> g) -> g -> g -> g
```
snowleopard: I guess if `simplify` becomes a good producer we won't need this re-definition.

test/Algebra/Graph/Test/RewriteRules.hs
```diff
@@ -13,111 +14,352 @@ module Algebra.Graph.Test.RewriteRules where
...
-inspect $ 'fmapFmap1 === 'fmapFmapR
+inspect $ 'simplifyC === 'simplifyCR
 
-bind2, bind2R :: (a -> Graph b) -> (b -> Graph c) -> Graph a -> Graph c
-bind2 f g x = x >>= f >>= g
-bind2R f g x = x >>= (\x -> f x >>= g)
+-- compose
+composeC, composeCR :: Ord a => Buildg a -> Buildg a -> Graph a
+composeC  x y = compose (buildg x) (buildg y)
+composeCR x y = overlays
```
snowleopard: Ouch. It's a bit sad we need to copy this.

snowleopard:
@nobrakal I had a thorough look through the changes and left a bunch of comments, most of which should be easy to address. Hope we can merge this soon!

src/Algebra/Graph/Internal.hs
nobrakal: Indeed :) Do you not fear a name clash with the actual `Data.Semigroup.sconcat` ?

test/Algebra/Graph/Test/RewriteRules.hs
nobrakal: I agree but I don't see a better option here...

src/Algebra/Graph.hs
nobrakal: This was a little trick when I compared directly the definitions of functions and expected form. This is not needed anymore.

src/Algebra/Graph.hs
nobrakal: Same answer :)

src/Algebra/Graph.hs
nobrakal: Sadly, goodness behave well with composition only when types agree.
`overlays . map vertex` will still be a good consumer of lists, but not a good producer, since `overlays` is only a good consumer of lists, and not of lists _of graphs_.
To say it more clearly, if we abstract constructors for the whole `vertices`, we need to reuse the same abstract constructors in `overlays`.

src/Algebra/Graph.hs
nobrakal: Indeed, I even tested it ^^

src/Algebra/Graph.hs
nobrakal: The use of direct pattern-matching is the problem here. You will have to write the two functions with `foldr` for them to be a good consumer of lists. I don't see right now a clear way to do that while still being efficient and elegant.

src/Algebra/Graph.hs
nobrakal: This is not straightforward: when you abstract your constructors using `buildg` you loose the `Eq` instance required by `simple` (intuitively, you could destruct your graph to a value which does not have an `Eq` instance)

nobrakal:
Thank you very much for your detailed review ! I addressed "simple" issues and will come back soon to more difficult ones ^^

src/Algebra/Graph/Internal.hs
snowleopard: Ah, good point. We do often import `Data.Semigroup`, so the name clash may become annoying. OK, let's keep `foldr1Safe` for now.

src/Algebra/Graph.hs
snowleopard: I see. I think you have already explained this to me once :) Could you add it as a comment, perhaps, with a TODO (that we'd like to simplify this to `overlays . map vertex`).

src/Algebra/Graph.hs
snowleopard: OK, your reply looks like a good TODO comment to me :)

src/Algebra/Graph.hs
snowleopard: Hmm, I see, that's interesting. Worth documenting this intuition!

src/Algebra/Graph.hs
nobrakal: Indeed, it is a good producer :)

src/Algebra/Graph.hs
nobrakal: There is no arguments, I don't get it. Inspecting core, the second argument is hidden in a local variable until last phase, preventing the rule to fire... I spent some time without success, so I will add a comment for now.

snowleopard:
@nobrakal Thanks for addressing my comments! 

By the way, it looks like there are some merge conflicts. Please rebase.

nobrakal:
@snowleopard I do not understand: it seems I am up-to-date with `master`. A local merge passes without problem on my machine, and even github seems happy with the PR.

snowleopard:
@nobrakal Oops, looks like a GitHub bug. I see 

> This branch cannot be rebased due to conflicts
Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch.

When "Rebase and merge" is selected. If I switch to "Squash and merge", which is more appropriate in this case anyway, everything gets green!

snowleopard:
@nobrakal OK, I think we can merge this if you are ready!

nobrakal:
I just found a new good consumer! Except this, all seems green for a merge :)

snowleopard:
@nobrakal Merged! Many thanks for this awesome improvement :-)

fawaz990:
ابغا أصير مشرف برامج ومواقع اتواصلو معي

PR #229, opened 2019-08-05
Hopcroft-Karp algorithm for maximum matchings in bipartite graphs

vasalf:


snowleopard:
@vasalf I only managed to have a quick look, but this PR looks very very cool! 

I'll give it a proper review as soon as I can. In the meanwhile, I think all these nice algorithms and new data structures should go into a separate `Algorithm` submodule.

snowleopard:
I temporarily labelled this `blocked` since it's difficult to review before #226 is merged.

vasalf:
@snowleopard I've found some BFS implementations in the library, they look much more functional than either of two suggested there. Maybe I should rewrite the algorithm using this idea?

snowleopard:
@vasalf Reusing the BFS algorithm that we have would be great! Is its interface flexible enough for you, or do you need to maintain your own state during graph traversal?

vasalf:
@snowleopard The return type of the `bfs` family of functions contains all the information we need. In the algorithm, we need two things from BFS:
* Is the «sink» reachable from «source» (in terms of augmenting paths, is any unmatched vertex reachable from another unmatched vertex in another part).
* Distance from the «source» to each matched vertex.

If we build a separate directed graph of augmenting paths, we can reuse `Algebra.Graph.AdjacencyMap.Algorithm.bfs`.

In imperative implementations of the Hopcroft-Karp algorithm, however, this graph isn't implicitly built, both for simplicity and speed.

snowleopard:
@vasalf Aha, I see. Well, I always prefer to first add an obviously correct (even if slow) implementation, and then iterate on it, making it more complex but also faster. 

If you feel that reusing the existing `bfs` would make this PR easier to complete and review, go for it! 

If you'd like to keep your current implementation, which is tailored for HK, this seems fine to me too.

luc-tielen:
So it's been 1.5 years since something happened to this PR.. are there any plans for continuing this work?
I need exactly this algorithm for something I'm building, if this PR is dead I will probably end up having to copy-paste part of it into my own codebase to get the functionality I need.. :sweat_smile:

snowleopard:
> So it's been 1.5 years since something happened to this PR.. are there any plans for continuing this work?

@vasalf planned to finish this PR but I'm not sure if he is still up for it. I know that he is going to be pretty busy for the next couple of months.

> I need exactly this algorithm for something I'm building, if this PR is dead I will probably end up having to copy-paste part of it into my own codebase to get the functionality I need..

Would you be interested in taking this PR to completion? I'm happy to review and merge!

luc-tielen:
@snowleopard I'm also kind of short on time, but I could help if it is just making CI pass for this PR. If it is also modifying the algorithm, then it might be more difficult since I am just "passing by" and need the algorithm, but don't have a deep understanding of it. 😅 

Also some things in this PR tripped me up: `List a b` type in combination with OverloadedLists.. that's not really clear in the docs.

snowleopard:
@luc-tielen Got it! Of course, you should feel free to just copy the existing implementation and use it if it works for you. If you do that, please let us know how it works for you.

> I'm also kind of short on time, but I could help if it is just making CI pass for this PR.

Rebasing and fixing CI would already be very helpful! I don't think any changes to the algorithms will be required, but I do want to make a pass and make some changes to various signatures/data types. I'm happy to do it myself, after merging this PR once the CI is green. Once I'm happy with the API, I'll immediately release the next version of the package, so you can get rid of the copied code.

This all isn't really very difficult, so I could just finish up the PR myself, but I'm currently on a parental leave (woohoo!), so my schedule is very uncertain for the next few weeks :)

luc-tielen:
Oh congratulations! That's also good for me since it gives me some time to fix CI then :wink:.

I will copy the code for now, I already played around with the code a little bit and rewrote it so it's consistent with the rest of my codebase.
Some suggestions for the current algorithm: 

1. Seems like `u` and `v` are swapped with left and right sometimes -> ended up confusing me when trying to understand the algorithm. 
2. Maybe also an idea to call the type variables of `AdjacencyMap` u and v instead of a and b?
3. I would split up the main algorithm into smaller helper functions: dfs, bfs, ... Now it's all in one _big_ function.
4. Make a dedicated distance data type `data Distance = Infinite | Finite Int deriving Eq`, helps a lot with readability over a `Maybe Int`
5. The `bfs` and `dfs` functions feel hard to reason about, and use short/cryptic variable names.

I can post my version later, I'm currently not on my laptop where I was working on the algorithm.

snowleopard:
> Oh congratulations! That's also good for me since it gives me some time to fix CI then 😉.

Thank you :-)

> Some suggestions for the current algorithm:
> [...]
> **EDIT** here is the slightly modified version, I still need to check it for correctness:

This is very helpful! By the way, there is already a `Distance` data type, see [here](https://hackage.haskell.org/package/algebraic-graphs-0.5/docs/Algebra-Graph-Label.html#t:Distance), although it may be a bit overengineered for this particular use case.

snowleopard:
This PR has been superseded by #274. Thanks @luc-tielen and @vasalf!

PR #232, opened 2019-08-18
Added property tests for Algebra.Graph.Label

adithyaov:
The initial draft for `Algebra.Graph.Test.Label`.


adithyaov:
`Minimum` unfortunately fails the right distributive property. It cannot be a Semiring.
Consider `String` as the underlying monoid.
With,
```
<+> ~ min
<.> ~ mappend
0 ~ inf
1 ~ ""
```
Consider,
```
a = Minimum (Finite "a")
b = Minimum (Finite "") ~ 1
c = Minimum (Finite "c")

x = (a <+> b) <.> c = Minimum (Finite "c")
y = (a <.> c) <+> (b <.> c) = Minimum (Finite "ac")
```
Clearly `x` is not equal to `y`.
This property fails if the monoid is `String`.
Similarly, this property will fail if the monoid is `Path a` as well.

`Minimum a` can be described as a **Left Near Ring** but not a Semiring.

test/Algebra/Graph/Test/Label.hs
```diff
@@ -0,0 +1,144 @@
...
+-----------------------------------------------------------------------------
+-- |
+-- Module     : Algebra.Graph.Test.Label
+-- License    : MIT (see the file LICENSE)
+-- Stability  : experimental
+--
+-- Testsuite for "Algebra.Graph.Label".
+-----------------------------------------------------------------------------
+{-
+module Algebra.Graph.Test.Label (
```
snowleopard: Why is this commented out? Let's add this header.

test/Algebra/Graph/Test/Label.hs
```diff
@@ -0,0 +1,144 @@
...
+commutative :: Eq a => BinaryFn a -> a -> a -> Property
+commutative f a b = a `f` b == b `f` a // "Commutative property"
+
+idempotance :: Eq a => BinaryFn a -> a -> Property
+idempotance f a = a `f` a == a // "Idempotence property"
+
+annhilator :: Eq a => BinaryFn a -> Annhilator a -> a -> Property
+annhilator f z a = conjoin
+  [ a `f` z == z // "Annhilator property 1"
+  , z `f` a == z // "Annhilator property 2" ]
```
snowleopard: These properties are usually referred to as "left" (`0 * x = 0`) and "right" (`x * 0 = 0`).

snowleopard: Same for some others below.

snowleopard:
Many thanks @adithyaov! This looks very nice. I've left a couple of minor comments.

test/Algebra/Graph/Test/Label.hs
adithyaov: That's for REPL testing. Will uncomment it in the final draft, before the merge.

snowleopard:
@adithyaov This doesn't currently pass CI. Could you push a fix please?

snowleopard:
Just restarted CI, perhaps we need to increase/remove Travis time limit for long-running commands?

src/Algebra/Graph/Label.hs
```diff
@@ -281,11 +282,17 @@ fromExtended :: Extended a -> Maybe a
 fromExtended (Finite a) = Just a
 fromExtended Infinite   = Nothing
 
-instance Num a => Num (Extended a) where
+-- A type alias for a binary function on Extended.
+type ExtendedBinary a = Extended a -> Extended a -> Extended a
```
snowleopard: Looks like this type is used only in the definition of `(*)` -- I suggest simply inlining this definition.

test/Algebra/Graph/Test/Label.hs
```diff
@@ -0,0 +1,149 @@
...
+associative :: Eq a => BinaryFn a -> a -> a -> a -> Property
+associative f a b c = (a `f` b) `f` c == a `f` (b `f` c) // "Associative property"
+
+commutative :: Eq a => BinaryFn a -> a -> a -> Property
+commutative f a b = a `f` b == b `f` a // "Commutative property"
+
+idempotance :: Eq a => BinaryFn a -> a -> Property
+idempotance f a = a `f` a == a // "Idempotence property"
+
+annhilator :: Eq a => BinaryFn a -> Annhilator a -> a -> Property
```
snowleopard: `annhilator` -> `annihilator`.

test/Algebra/Graph/Test/Label.hs
```diff
@@ -0,0 +1,149 @@
...
+type Annhilator a = a
+
+associative :: Eq a => BinaryFn a -> a -> a -> a -> Property
+associative f a b c = (a `f` b) `f` c == a `f` (b `f` c) // "Associative property"
+
+commutative :: Eq a => BinaryFn a -> a -> a -> Property
+commutative f a b = a `f` b == b `f` a // "Commutative property"
+
+idempotance :: Eq a => BinaryFn a -> a -> Property
+idempotance f a = a `f` a == a // "Idempotence property"
```
snowleopard: `idempotance` -> `idempotence`.

snowleopard: Even better, let's use adjectives, as above, i.e. `idempotent`.

test/Algebra/Graph/Test/Label.hs
```diff
@@ -0,0 +1,149 @@
...
+
+type UnaryFn a = a -> a
+type BinaryFn a = a -> a -> a
+type Plus a = BinaryFn a
+type Mult a = BinaryFn a
+type Star a = UnaryFn a
+type Identity a = a
+type Zero a = a
+type One a = a
+type Annhilator a = a
```
snowleopard: `Annhilator` -> `Annihilator`.

test/Algebra/Graph/Test/Label.hs
```diff
@@ -0,0 +1,149 @@
...
+  [ s a == o `p` (a `m` s a) // "Star property right"
+  , s a == o `p` (s a `m` a) // "Star property left" ]
+
+leftDistribute :: Eq a => Plus a -> Mult a -> a -> a -> a -> Property
+leftDistribute p m a b c = a `m` (b `p` c) == (a `m` b) `p` (a `m` c) // "Left distribute"
+
+rightDistribute :: Eq a => Plus a -> Mult a -> a -> a -> a -> Property
+rightDistribute p m a b c = (a `p` b) `m` c == (a `m` c) `p` (b `m` c) // "Right distribute"
+
+distribute :: Eq a => Plus a -> Mult a -> a -> a -> a -> Property
```
snowleopard: `distribute` -> `distributive` (same above).

test/Algebra/Graph/Test/Label.hs
```diff
@@ -0,0 +1,149 @@
...
+idempotance :: Eq a => BinaryFn a -> a -> Property
+idempotance f a = a `f` a == a // "Idempotence property"
+
+annhilator :: Eq a => BinaryFn a -> Annhilator a -> a -> Property
+annhilator f z a = conjoin
+  [ a `f` z == z // "Annhilator property right"
+  , z `f` a == z // "Annhilator property left" ]
+
+star' :: Eq a => Plus a -> Mult a -> One a -> Star a -> a -> Property
+star' p m o s a = conjoin
```
snowleopard: Let's call this property `closure`.

test/Algebra/Graph/Test/Label.hs
```diff
@@ -0,0 +1,149 @@
...
+testSemiring = semiring (<+>) zero (<.>) one
+
+testDioid :: (Eq a, Dioid a) => a -> a -> a -> Property
+testDioid = dioid (<+>) zero (<.>) one
+
+testStarSemiring :: (Eq a, StarSemiring a) => a -> a -> a -> Property
+testStarSemiring = starSemiring (<+>) zero (<.>) one star
+
+on3 :: (b -> a) -> (a -> a -> a -> c) -> b -> b -> b -> c
+on3 g f a b c = f (g a) (g b) (g c)
```
snowleopard: This appears to be unused?

snowleopard:
@adithyaov Very nice PR, many thanks!

Just managed to find time for review and left a few minor comments.

P.S.: CI has passed after restarting.

adithyaov:
@snowleopard Although `semigroup` and `rightNearRing` are unused I'm leaving them there, gives a sense of completion of properties.
Ideally, I'd want to replace `liftA2` with `coerce` but using `liftA2` seems cleaner. Since the correctness isn't violated, I don't think it's a problem.

adithyaov:
@snowleopard Could you please restart Travis CI. I don't understand why it fails. As you suggested, increasing the time limit might work.

snowleopard:
@adithyaov I've restarted the CI.

If you can find a way to make CI more reliable, please let me know how.

adithyaov:
@snowleopard I think this PR is ready for merge. Profiling might take a little more time. Can we safely ignore timeouts for the time being?

snowleopard:
@adithyaov Having reliable CI is very important since it helps a lot with reviews.

This particular PR seems to push the testsuite too far, making CI unreliable, so I don't think we should merge it until we eigher get rid of slow tests in general, or slow tests in this particular PR that are responsible for slowing down the testsuite.

adithyaov:
@snowleopard Agreed. I'll try to complete profiling the tests as soon as possible.

src/Algebra/Graph/Label.hs
```diff
@@ -133,7 +134,7 @@ instance (Num a, Ord a) => Num (NonNegative a) where
         f = fromInteger x
 
     (+) = liftA2 (+)
-    (*) = liftA2 (*)
+    (*) = coerce ((*) :: Extended a -> Extended a -> Extended a)
```
snowleopard: Why are `+` and `*` treated differently above?

snowleopard: If we can use `(+) = coerce ((+) :: Extended a -> Extended a -> Extended a)` then lets use it to make it more symmetric.

test/Algebra/Graph/Test/Label.hs
```diff
@@ -0,0 +1,147 @@
...
+  [ a `f` z == z // "Annihilator property right"
+  , z `f` a == z // "Annihilator property left" ]
+
+closure :: Eq a => Plus a -> Mult a -> One a -> Star a -> a -> Property
+closure p m o s a = conjoin
+  [ s a == o `p` (a `m` s a) // "Star property right"
+  , s a == o `p` (s a `m` a) // "Star property left" ]
+
+leftDistributive :: Eq a => Plus a -> Mult a -> a -> a -> a -> Property
+leftDistributive p m a b c = a `m` (b `p` c) == (a `m` b) `p` (a `m` c) // "Left distributive"
```
snowleopard: Just a thought: you could probably use operators for better readability, as follows:

```haskell
leftDistributive (*) (+) a b c = a * (b + c) == (a * b) + (a * c) // "Left distributive"

```

snowleopard: Unfortunately, this trick doesn't work with `0` and `1` :)

test/Algebra/Graph/Test/Label.hs
```diff
@@ -0,0 +1,147 @@
...
+
+commutative :: Eq a => BinaryFn a -> a -> a -> Property
+commutative f a b = a `f` b == b `f` a // "Commutative property"
+
+idempotent :: Eq a => BinaryFn a -> a -> Property
+idempotent f a = a `f` a == a // "Idempotence property"
+
+annihilator :: Eq a => BinaryFn a -> Annihilator a -> a -> Property
+annihilator f z a = conjoin
+  [ a `f` z == z // "Annihilator property right"
```
snowleopard: I would drop the word "property" from all `//` annotations, writing simply "Right annihilator" in such cases.

By the way, I think this particular property is usually referred to as "Right zero".

test/Algebra/Graph/Test/Label.hs
```diff
@@ -0,0 +1,147 @@
...
+    test "Semiring"     $ \(a :: Distance Int) b c -> testSemiring a b c
+    test "StarSemiring" $ \(a :: Distance Int) b c -> testStarSemiring a b c
+    test "Dioid"        $ \(a :: Distance Int) b c -> testDioid a b c
+
+    putStrLn "\n============ Capacity ============"
+    test "Semiring"     $ \(a :: Capacity Int) b c -> testSemiring a b c
+    test "StarSemiring" $ \(a :: Capacity Int) b c -> testStarSemiring a b c
+    test "Dioid"        $ \(a :: Capacity Int) b c -> testDioid a b c
+
+    putStrLn "\n============ Minimum ============"
```
snowleopard: You are actually testing `Minimum (Path Int)` below, not `Minimum`.

I think we should add tests for all types of labels we define in `Algebra.Graph.Label`, perhaps, also adding some combinations like `Minimum (Path Int)`.

test/Algebra/Graph/Test/Label.hs
```diff
@@ -0,0 +1,147 @@
...
+    test "LeftNearRing" $ \(a :: Minimum (Path Int)) b c -> testLeftNearRing a b c
+
+    putStrLn "\n============ PowerSet ============"
```
snowleopard: Same here.

snowleopard:
@adithyaov Thanks a lot for fixing the testsuite! I've added some comments.

test/Algebra/Graph/Test/Label.hs
adithyaov: > perhaps, also adding some combinations like Minimum (Path Int).
Perhaps you meant something else? `Minimum (Path Int)` already exists.

snowleopard: I meant similar to `Minimum Path` that you are already testing.

test/Algebra/Graph/Test/Label.hs
adithyaov: @snowleopard I wonder why it won't work for 0 and 1? Do you mean something like `leftDistributive (<*>) (<+>) 0 1 1` won't work as expected?

snowleopard: I mean, you cannot implement `identity` as 
```haskell
identity (+) 0 x = (0 + x == x) && (x + 0 == x)
```
because when we use `0` on the left hand side we are comparing to zero instead of binding the input parameter to a new variable named `0` :)

adithyaov: @snowleopard I am inclined to make this change. But there are also some properties with a single operator. I'd like to change those as well to maintain a sense of similarity. I don't want to use specific operators like `(*)` and `(+)`. Should I use `(..)`? If not, please suggest another operator.

snowleopard: I think `*` works in most cases (e.g. `x * z == z`) apart perhaps from idempotence, where one could use something neutral like `<>` since the operation in question is typically a semigroup.

test/Algebra/Graph/Test/Label.hs
adithyaov: I will do it in another PR. Some types (`Optimum`) need not satisfy `Semiring` yet can be used. I need to come up with a proper/better reasoning.

snowleopard:
@adithyaov Great work, thank you. Merged!

Issue #233, opened 2019-09-16
return type for bfs


jitwit:
The current return type/implementation for `bfs` loses connectivity information and makes it unclear whether vertices appear later in the resulting list because they are in a deeper layer of the forest, or because they are in disjoint trees. For example,
```
λ> bfs [1,5] $ circuit [1..3] + circuit [3,2,1] + 5*6
[[1],[2,3],[5],[6]]
it :: [[Int]]
```
Instead, something like 
```
λ> bfs [1,5] $ circuit [1..3] + circuit [3,2,1] + 5*6
[[[1],[2,3]],[[5],[6]]]
it :: [LevelStructure Int]
```
might make more sense, where `type LevelStructure a = [[a]]`?

The change would be:
```
bfs :: [Int] -> AdjacencyIntMap -> [[Int]]
bfs vs = concatMap levels . bfsForestFrom vs
```
to
```
type LevelStructure a = [[a]]

bfs :: [Int] -> AdjacencyIntMap -> [LevelStructure Int]
bfs vs = map levels . bfsForestFrom vs
```

snowleopard:
@jitwit Hmm, shouldn't the result be `[[1,5],[2,3,6]]` instead? That is, vertices 1 and 5 are reachable in 0 steps from the sources, and vertices 2, 3 and 6 are reachable in 1 step?

I can see how your proposed representation preserves even more information but may be somewhat difficult to use (and understand!).

jitwit:
Yeah, there's a lot of nesting with the triple list. I figured it could be useful to keep the components separate, but otherwise I think we can get result you propose with 
```
bfs vs = map concat . Data.List.transpose . bfsForestFrom vs
```

snowleopard:
> I figured it could be useful to keep the components separate

@jitwit But I don't think the components will be separate. Are you referring to strongly-connected components? Then consider the graph `1 * 1 + 1 * 2 + 2 * 2`: it has two connected components `1 * 1` and `2 * 2` but depending on the order of source vertices you will get different results. With `[1, 2]` you'll get what appears to be a single component, and with `[2, 1]` you'll get two. I think the resulting triply-nested list has a rather non-obvious structure.

jitwit:
Ah you're right, that example shows it doesn't make sense to return `[[[a]]]` for the reason I stated! 

[...]

PR #234, opened 2019-09-16
update bfs to treat input vertices as same level

jitwit:
Addresses #233


src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -117,20 +118,22 @@ bfsForestFrom' vs g = evalState (explore vs) IntSet.empty where
...
--- bfs [1] ('edge' 1 2)                                   == [[1],[2]]
--- bfs [2] ('edge' 1 2)                                   == [[2]]
+-- bfs [1] ('edge' 1 1)                                   == [[[1]]]
+-- bfs [1] ('edge' 1 2)                                   == [[[1],[2]]]
+-- bfs [2] ('edge' 1 2)                                   == [[[2]]]
 -- bfs [3] ('edge' 1 2)                                   == []
--- bfs [1,2] ('edge' 1 2)                                 == [[1],[2]]
--- bfs [2,1] ('edge' 1 2)                                 == [[2],[1]]
+-- bfs [1,2] ('edge' 1 2)                                 == [[[1],[2]]]
+-- bfs [2,1] ('edge' 1 2)                                 == [[2]],[[1]]
```
snowleopard: Aha, this is exactly my example from #233 :) Different number of "components" depending on the order of source vertices.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -118,22 +118,20 @@ bfsForestFrom' vs g = evalState (explore vs) IntSet.empty where
...
--- bfs [1] ('edge' 1 1)                                   == [[[1]]]
--- bfs [1] ('edge' 1 2)                                   == [[[1],[2]]]
--- bfs [2] ('edge' 1 2)                                   == [[[2]]]
+-- bfs [1] ('edge' 1 1)                                   == [[1]]
+-- bfs [1] ('edge' 1 2)                                   == [[1],[2]]
+-- bfs [2] ('edge' 1 2)                                   == [[2]]
 -- bfs [3] ('edge' 1 2)                                   == []
--- bfs [1,2] ('edge' 1 2)                                 == [[[1],[2]]]
--- bfs [2,1] ('edge' 1 2)                                 == [[2]],[[1]]
+-- bfs [1,2] ('edge' 1 2)                                 == [[1],[2]]
```
snowleopard: I'm confused by this example: shouldn't both vertices be in the first group since they are reachable in 0 steps?

jitwit: I got confused by this earlier as well. Basically, the tree from 1 is fully expanded before considering 2.

I remembered when first adding `bfsForestFrom`, I got tripped up by this too, since if the input vertices get searched "in parallel" 
```
bfsForest g = bfsForestFrom (vertexList g)
```
returns a forest of singleton trees for each vertex. This approach was to mirror the dfs API, though I'm not sure it's the most natural approach.

snowleopard: I think it would be great to have the property that vertices in k'th group are k steps away from the source vertices. BFS is primarily used when the distance information is important and I think having such API would be more useful than the one mirroring the approach used in DFS.

Presumably, it's not very difficult implementation-wise?

jitwit: Yeah that makes sense. I think it might actually be an easier implementation, since I had to be careful about the evaluation order with the current implementation. Should I go ahead and make the changes in this pull request?

snowleopard: Great! Sure, feel free to do the changes right here.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -91,17 +96,13 @@ bfsForest g = bfsForestFrom' (vertexList g) g
...
-  explore (v:vs) = discovered v >>= \case
-    True -> (:) <$> unfoldTreeM_BF walk v <*> explore vs
-    False -> explore vs
-  explore [] = return []
+bfsForestFrom' :: AdjacencyIntMap -> [Int] -> State IntSet.IntSet (Forest Int)
+bfsForestFrom' g = unfoldForestM_BF walk <=< filterM discovered where
   walk v = (v,) <$> adjacentM v
   adjacentM v = filterM discovered $ IntSet.toList (postIntSet v g)
-  discovered v = do new <- gets (not . IntSet.member v)
+  discovered v = do new <- gets ((&& hasVertex v g) . not . IntSet.member v)
```
jitwit: Checking here probably isn't the right move. The old way became awkward with the new semantics, but this seems to have impacted performance slightly

test/Algebra/Graph/Test/Generic.hs
```diff
@@ -1763,8 +1763,8 @@ testBfsForestFrom (prefix, API{..}) = do
     test "isSubgraphOf (forest $ bfsForestFrom vs x) x     == True" $ \vs x ->
           isSubgraphOf (forest $ bfsForestFrom vs x) x     == True
 
-    test "bfsForestFrom (vertexList x) x                   == bfsForest x" $ \x ->
-          bfsForestFrom (vertexList x) x                   == bfsForest x
+--    test "bfsForestFrom (vertexList x) x                   == bfsForest x" $ \x ->
+--          bfsForestFrom (vertexList x) x                   == bfsForest x
```
jitwit: This no longer holds with new semantics

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
snowleopard: Wait, what's awkward about keeping `[ v | v <- vs, hasVertex v g ]` at the top level as before?

test/Algebra/Graph/Test/Generic.hs
snowleopard: I think plain `bfsForest` doesn't make sense in the new semantics... so perhaps, we should just drop it and rename `bfsForestFrom` to `bfsForest`?

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: Nevermind! not sure what I was thinking

test/Algebra/Graph/Test/Generic.hs
jitwit: Yeah, I think that makes sense. In fact, fgl doesn't export `bfsForest` or an equivalent either

snowleopard: 👍

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -40,65 +40,41 @@ import Data.Tree
...
 --                                                                             , subForest = [] }]}
 --                                                 , Node { rootLabel = 4
 --                                                        , subForest = [] }]
--- 'forest' (bfsForestFrom [3] ('circuit' [1..5] + 'circuit' [5,4..1])) == 'path' [3,2,1] + 'path' [3,4,5]
+-- 'forest' (bfsForest [3] ('circuit' [1..5] + 'circuit' [5,4..1])) == 'path' [3,2,1] + 'path' [3,4,5]
 -- 
 -- @
-bfsForestFrom :: [Int] -> AdjacencyIntMap -> Forest Int
-bfsForestFrom vs g = bfsForestFrom' [ v | v <- vs, hasVertex v g ] g
+bfsForest :: [Int] -> AdjacencyIntMap -> Forest Int
```
snowleopard: Please fix all remaining references to `bfsForestFrom` -- I found one in the `bfs` comment.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -40,65 +40,41 @@ import Data.Tree
...
--- bfsForestFrom [1,4] (3 * (1 + 4) * (1 + 5))  == [ Node { rootLabel = 1
+-- 'forest' (bfsForest [1,2] $ 'edge' 1 2)      == 'vertices' [1,2]
+-- 'forest' (bfsForest [2]   $ 'edge' 1 2)      == 'vertex' 2
+-- 'forest' (bfsForest [3]   $ 'edge' 1 2)      == 'empty'
+-- 'forest' (bfsForest [2,1] $ 'edge' 1 2)      == 'vertices' [1,2]
+-- 'isSubgraphOf' ('forest' $ bfsForest vs x) x == True
+-- bfsForest vs ('vertices' vs)               == 'map' (\v -> Node v []) ('nub' vs)
+-- bfsForest [] x                           == []
+-- bfsForest [1,4] (3 * (1 + 4) * (1 + 5))  == [ Node { rootLabel = 1
 --                                                        , subForest = [ Node { rootLabel = 5
```
snowleopard: I suspect the alignment is lost here.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -40,65 +40,41 @@ import Data.Tree
...
+bfsForest vs g = bfsForest' [ v | v <- vs, hasVertex v g ] g
 
-bfsForestFrom' :: [Int] -> AdjacencyIntMap -> Forest Int
-bfsForestFrom' vs g = evalState (explore vs) IntSet.empty where
-  explore (v:vs) = discovered v >>= \case
-    True -> (:) <$> unfoldTreeM_BF walk v <*> explore vs
-    False -> explore vs
-  explore [] = return []
+bfsForest' :: [Int] -> AdjacencyIntMap -> Forest Int
+bfsForest' vs g = evalState (explore vs) IntSet.empty where
```
snowleopard: I'd probably just do vertex filtering right here to avoid the indirection, i.e.

```haskell
bfsForest vs g = evalState (explore [ v | v <- vs, hasVertex v g ]) IntSet.empty where
```

test/Algebra/Graph/Test/Generic.hs
```diff
@@ -1812,16 +1766,16 @@ testBfs (prefix, API{..}) = do
           bfs [3]    (edge 1 2)                == []
 
     test "bfs [1,2] $ edge 1 2                 == [[1],[2]]" $
```
snowleopard: The text above doesn't match the test.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
snowleopard: I think this example is still incorrect but the test does seem to use `[[1,2]]` already.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -40,65 +40,41 @@ import Data.Tree
...
--- 'forest' (bfsForest ('circuit' [1..5] + 'circuit' [5,4..1])) == 'path' [1,2,3] + 'path' [1,5,4]
--- @
-bfsForest :: AdjacencyIntMap -> Forest Int
-bfsForest g = bfsForestFrom' (vertexList g) g
-
--- | Like 'bfsForest', but the traversal is seeded by a list of
---   vertices. Vertices not in the graph are ignored.
+--   to their 'Ord' instance. The search is seeded by a list of
+--   argument vertices that will be the roots of the resulting
+--   forest. Argument vertices not in the graph are ignored.
```
snowleopard: I think we should add here that the order of vertices in the list is not important, i.e. we actually treat the list as a set.

snowleopard: Oh, wait, I guess I'm wrong, the order does matter!

snowleopard: Perhaps, we should have a simple example where the order does matter. Something like `bfs [1,2] (1+2)` will give `[1,2]` whereas `bfs [2,1] (1+2)` will give `[2,1]`, right?

snowleopard: In some sense this is subsumed by this example:

```haskell
-- bfsForest vs ('vertices' vs)               == 'map' (\v -> Node v []) ('nub' vs)
```

So, unsure if an extra example is actually needed. I'll let you decide.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -122,15 +98,15 @@ bfsForestFrom' vs g = evalState (explore vs) IntSet.empty where
 -- bfs [2] ('edge' 1 2)                                   == [[2]]
 -- bfs [3] ('edge' 1 2)                                   == []
 -- bfs [1,2] ('edge' 1 2)                                 == [[1],[2]]
--- bfs [2,1] ('edge' 1 2)                                 == [[2],[1]]
+-- bfs [2,1] ('edge' 1 2)                                 == [[2,1]]
 -- bfs [1,2] ( (1*2) + (3*4) + (5*6) )                  == [[1],[2]]
```
snowleopard: This should probably be `[[1,2]]` too?

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -122,15 +98,15 @@ bfsForestFrom' vs g = evalState (explore vs) IntSet.empty where
...
--- bfs [1,3] ( (1*2) + (3*4) + (5*6) )                  == [[1],[2],[3],[4]]
+-- bfs [1,3] ( (1*2) + (3*4) + (5*6) )                  == [[1,3],[2,4]]
 -- bfs [3] (3 * (1 + 4) * (1 + 5))                      == [[3],[1,4,5]]
 -- bfs [2] ('circuit' [1..5] + 'circuit' [5,4..1])          == [[2],[1,3],[5,4]]
 -- 'concat' (bfs [3] $ 'circuit' [1..5] + 'circuit' [5,4..1]) == [3,2,4,1,5]
 -- @
 bfs :: [Int] -> AdjacencyIntMap -> [[Int]]
-bfs vs = bfsForestFrom vs >=> levels
+bfs vs = map concat . List.transpose . map levels . bfsForest vs
```
snowleopard: This is nice brain teaser :) Perhaps, it's worth documenting it as a property, since it's far from obvious for me.

snowleopard:
@jitwit Thanks for further work, I've left a bunch of minor comments.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: It actually seems to be ok. It's still unaligned with
```
-- 'forest' (bfsForest [3] ('circuit' [1..5] + 'circuit' [5,4..1])) == 'path' [3,2,1] + 'path' [3,4,5]
```
though. I think it makes sense to push the others out to match?

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: Makes sense!

jitwit:
Thanks and apologies for the sloppiness with documentation/names/tests!

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: At the top level, the vertices aren't rearranged, so 2 comes before 1 here. 

They do get nubbed at the top leve in `filterM discover [ v | v <- hasVertex v g ]`, possibly it makes sense to present them in ascending order from that? My thinking was that it may be nice to retain control over the order the search trees get expanded.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: I think rearranging the order may help make the relations clearer. Probably more information in the documentation as well would help

jitwit: Actually this example could be strengthened:
```
bfsForest g ('vertexList' g)               == 'map' (\v -> Node v []) ('vertexList' g)
```

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: One issue is that the example before it has quite long lines with the search tree spelled out.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: Haha yes, the realization came as a modest light-bulb moment. I'm not sure I understand what it means to document as a property though?

test/Algebra/Graph/Test/Generic.hs
```diff
@@ -1699,88 +1699,42 @@ testSimplify (prefix, API{..}) = do
...
+          forest (bfsForest [2,1] $ edge 1 2)          == vertices [1,2]
 
-    test "bfsForest . forest . bfsForest        == bfsForest" $ \x ->
-         (bfsForest . forest . bfsForest) x     == bfsForest x
+    test "isSubgraphOf (forest $ bfsForest vs x) x     == True" $ \vs x ->
+          isSubgraphOf (forest $ bfsForest vs x) x     == True
 
-    test "bfsForest (vertices vs)               == map (\\v -> Node v []) (nub $ sort vs)" $ \vs ->
-          bfsForest (vertices vs)               == map (\v -> Node v []) (nub $ sort vs)
+    test "bfsForest vs             (vertices vs)       == map (\\v -> Node v []) (nub vs)" $ \g ->
```
jitwit: As always!

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
```diff
@@ -41,97 +41,72 @@ import Algebra.Graph.AdjacencyMap
...
--- bfs [1,2] ('edge' 1 2)                                 == [[1],[2]]
--- bfs [2,1] ('edge' 1 2)                                 == [[2],[1]]
 -- bfs [1,2] ( (1*2) + (3*4) + (5*6) )                  == [[1],[2]]
--- bfs [1,3] ( (1*2) + (3*4) + (5*6) )                  == [[1],[2],[3],[4]]
+-- bfs [1,3] ( (1*2) + (3*4) + (5*6) )                  == [[1,3],[2,4]]
 -- bfs [3] (3 * (1 + 4) * (1 + 5))                      == [[3],[1,4,5]]
 -- bfs [2] ('circuit' [1..5] + 'circuit' [5,4..1])          == [[2],[1,3],[5,4]]
 -- 'concat' (bfs [3] $ 'circuit' [1..5] + 'circuit' [5,4..1]) == [3,2,4,1,5]
+-- bfs vs == 'map' 'concat' . 'List.transpose' . 'map' 'levels' . 'bfsForest' vs
 -- @
```
jitwit: Is this what you had in mind for documenting as a property?

snowleopard:
@jitwit @nobrakal Looks like the performance suite got broken. Did something go wrong with @jitwit's recent PR?

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
snowleopard: Yes, thank you!

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
snowleopard: I meant the alignment between `{` and `,` in the subsequent line :) It's a nit pick, don't worry about it!

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
snowleopard: In tests you still have:

```haskell
bfs [1,2] ((1*2) + (3*4) + (5*6))    == [[1,2]]
```

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
snowleopard: Aha, good example and test case!

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: Ah I see! I'll get it in the next wave. I think there's still one or two things out of sync anyway.

There are a lot of docs/tests where the text uses `$` but the tests use the necessary parens. I was wondering if it would be better if they matched more closely? It would make it easier to update I think

jitwit:
I'm not sure what's causing it. It looks like the travis build for the pull request have mangled char encoding as well

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
snowleopard: > There are a lot of docs/tests where the text uses `$` but the tests use the necessary parens. I was wondering if it would be better if they matched more closely? It would make it easier to update I think

Yeah, there are a few such cases. I guess it would be better to switch to using parentheses instead of providing examples with `$` that don't actually compile, so feel free to fix this.

jitwit:
Could this be a stack/cabal thing?

nobrakal:
@snowleopard  Indeed, I needed to adapt the script. It should run fine now, unless the two new benchs kills the 50mins limit.

jitwit:
Works again! Thanks

snowleopard:
@nobrakal Thanks! The one for `Graph.NonEmpty` still doesn't seem to work.

@jitwit Am I right that we do not expect any changes in terms of performance here?

jitwit:
Yes, I believe so. If anything, I wasn't sure about `map concat . transpose . map levels` vs `concatMap levels`

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
```diff
@@ -41,97 +41,73 @@ import Algebra.Graph.AdjacencyMap
...
+-- bfs [2]   ('edge' 1 2)                                 == [[2]]
+-- bfs [1,2] ('edge' 1 2)                                 == [[1,2]]
+-- bfs [2,1] ('edge' 1 2)                                 == [[2,1]]
+-- bfs [3]   ('edge' 1 2)                                 == []
+-- bfs [1,2] ( (1*2) + (3*4) + (5*6) )                  == [[1,2]]
+-- bfs [1,3] ( (1*2) + (3*4) + (5*6) )                  == [[1,3],[2,4]]
 -- bfs [3] (3 * (1 + 4) * (1 + 5))                      == [[3],[1,4,5]]
 -- bfs [2] ('circuit' [1..5] + 'circuit' [5,4..1])          == [[2],[1,3],[5,4]]
 -- 'concat' (bfs [3] $ 'circuit' [1..5] + 'circuit' [5,4..1]) == [3,2,4,1,5]
+-- bfs vs == 'map' 'concat' . 'List.transpose' . 'map' 'levels' . 'bfsForest' vs
```
snowleopard: I didn't find the corresponding test, but we should have one: in future, we might change the implementation but we'd still need to make sure this property holds.

snowleopard:
@jitwit I'm back to action and will try to merge both of your PRs this week. See one last remark above.

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
jitwit: Ah yes, that makes a lot of sense! I was initially thinking it was unnecessary to write a test that mirrored exactly the implementation

snowleopard:
Merged, thanks again!

PR #235, opened 2019-09-18
scc algorithm

jitwit:
Implementation of scc algorithm for adjacency maps and adjacency int maps based on [https://en.wikipedia.org/wiki/Path-based_strong_component_algorithm](https://en.wikipedia.org/wiki/Path-based_strong_component_algorithm). 

Avoids using Data.Graph, and benchmarks indicate that this pays off [https://jitwit.github.io/criterion/scc-bench.html](https://jitwit.github.io/criterion/scc-bench.html). In report, old-alga corresponds to alga's KL implementation, new-alga the intmap implementation, ord-alga the new implementation for AMs. Graphs used are the real-world ones from haskell-perf/graphs. AM implementation seems to be around 10% faster and AIM version around 65% faster.

There is no NonEmpty.AdjacencyIntMap module, so for now AdjacencyIntMap.Algorithm returns type `AM.AdjacencyMap AdjacencyIntMap`.


snowleopard:
@jitwit The results are great! You are unstoppable :-) 

`scc` is quite an important algorithm with many users, so perhaps we should have it in our regression suite to keep an eye on its performance. @nobrakal Do you think you could add it?

snowleopard:
P.S.: I'll need some time to review the implementation due to upcoming travel.

jitwit:
I was actually going to ask about how to add things to regression suite, since it would be useful for `bfs` too (using a different queue structure might be an opportunity for slightly better performance).

snowleopard:
Here is where the performance testing script comes from: https://github.com/snowleopard/alga/blob/35465faff04f0f2a608c5761f62cc65dee122b71/.travis.yml#L105.

I guess we could move it to this repository to make changes more convenient. 

@nobrakal What do you think?

nobrakal:
@snowleopard 

I am actually ashamed of this script, and currently, I think rewriting a very simple benchmarking suite and getting rid of https://github.com/haskell-perf/graphs would be the best option.

I would love to gave `haskell-perf/graphs` the rewriting it needs, but I don't have any time to do it now.

Anyway, adding `scc` seems very feasible and I can give it a try if you think it is the best idea :)

snowleopard:
@nobrakal It's always possible to improve things but it's not always necessary :-) The current implementation may be not pretty but the script does work and is very useful!

If you could add `scc` (and perhaps `bfs` too) that would be great. And, of course, I'd be happy to have a prettier script when you find time for this :)

snowleopard:
@jitwit There are some merge conflicts, please resolve.

jitwit:
I re-ran the benchmarks and the results were less favorable in some cases for AdjacencyMaps. The graphs were the current version of this new implementation does poorly are those with a high number of small SCCs, the graphs were it does well have a small number of large SCCs. The AdjacencyIntMap version is much faster regardless. 

The link to the criterion benchmarks from before has the current results.

snowleopard:
@jitwit I'm looking at this link which corresponds to `AdjacencyIntMap` I think:

https://jitwit.github.io/criterion/scc-bench.html

This is better across all benchmarks, but can you show the results for `AdjacencyMap`? I'd like to have more performance info about the problematic cases you mention.

snowleopard:
By the way, the performance regression suite reports `scc: 1.10 (OK)`.

jitwit:
Yeah those names are unclear, I changed them to KL-alga, AM-alga, AIM-alga.

I've looked closer, and the `scc` algorithm does perform well, the issue seems to be converting to the `AM (NonEmptya a)` representation afterwards. When there are many sccs, using `induce` is slow. When there are few `sccs` classifying edges (another approach I'm experimenting with) is slow.

The result of the `scc` search contains some additional information, so I was wondering if a hybrid approach might be feasible? For example, depending on the ratio of the number of sccs to the number of vertices, a different function could be used to convert to the final representation

jitwit:
The benches are updated with the "hybrid" approach

jitwit:
I tried the searches on a twitter graph from SNAP: [https://snap.stanford.edu/data/ego-Twitter.html](https://snap.stanford.edu/data/ego-Twitter.html) (~81k vertices ~1.7m edges) and the results are good:

[https://jitwit.github.io/criterion/twitter-scc.html](https://jitwit.github.io/criterion/twitter-scc.html)

snowleopard:
@jitwit Very impressive, especially the Twitter benchmark! I think it's worth adding it to the benchmark suite at https://github.com/haskell-perf/graphs. 

I'm concerned with the complexity of the resulting code though :( How much improvement is your hybrid approach bringing percentage-wise? If it's not too much then I think I'd prefer to remove the special tricks and instead focus on making a faster `induce`, if that's possible.

jitwit:
The number of passes over the graph with `induce` increases with the number of sccs. I tried the algorithm with `induce` on the twitter graph and it was something like 5x slower. I think the right solution may be to ditch the hybrid approach as well as induce. Something like 
```
partition :: Eq k => (a -> k) -> Graph a -> [Graph a]
```
may be what we really want.

There are still a few cases where the current approach is slower than the old implementation, but for the most part it's faster.

jitwit:
I was also wondering how important is the `AM (NonEmptyAM a)` representation? What queries are most commonly asked of the result of `scc`?

jitwit:
Ach, `scc $ vertices xs` isn't handled properly, I'll fix it soon

snowleopard:
@jitwit 

> The number of passes over the graph with `induce` increases with the number of sccs. I tried the algorithm with `induce` on the twitter graph and it was something like 5x slower. 

Oh, indeed, I didn't realise the current implementation is quadratic due to multiple calls to `induce`. This is certainly a bug! Can we call `induce` only once? I'm fine to introduce another primitive if it helps to solve this issue, although I'm not yet sure how `partition` helps. I think we need a way to induce a two-level structure in a graph without paying the cost of calling `induce` multiple times.

> I was also wondering how important is the `AM (NonEmptyAM a)` representation?

I think the ideal type for `scc` is

```haskell
scc :: Ord a => AdjacencyMap a -> Acyclic.AdjacencyMap (NonEmpty.AdjacencyMap a)
```

I love the ability to express the precise structure of `scc` in types and I hope we'll switch to this type at some point when we have a solid story for acyclic graphs.

For now, let's try to stick to `AM (NonEmptyAM a)`: I think it's worth spending some time thinking how to fit an efficient engine behind a nice shiny API :)

snowleopard:
Perhaps, we need the following variant of `partition`?

```
partition :: Eq k => (a -> k) -> Graph a -> Graph (NonEmpty.Graph a)
```

jitwit:
Hi, sorry for the long pause! 

Yeah, I'm trying to come up with a reasonable approach. The `dfs g` builds a mapping `scc :: V -> Int` which assigns vertices to components. I guess my main point with `partition` was that we're classifying over a larger type than `Bool` with `induce`. 

My best shot so far is to do a fold `Map.foldrWithKey` to classify the edges and vertices followed `gmap`ing by component id. Each edge `(u,v)` of `g` will either be included as an edge between components `(scc u, scc v)` or an edge within the component `scc(u)` `(u,v)`. (I think a `foldg` would be useful, but it's in `Algebra.Graph.ToGraph` which imports the algorithms modules)

The current iteration does comparatively well on most graphs. The exception is graphs with only a few sccs.

[https://jitwit.github.io/criterion/scc-bench.html](https://jitwit.github.io/criterion/scc-bench.html)

The `induce` approach did rather well on the graphs where the current approach does poorly, which was why I thought it may be best to still it for some graphs. Being able to detect the case of 1 `scc` gives a rather large speedup to the current implementation (where we can just return `AM.vertex g`).

jitwit:
I think the main awkwardness is that having `NonEmpty.AM a`s as vertices means that manipulating the outer `Acyclic.AM` map of the condensation is expensive. Returning  a`Graph (NonEmpthy.AM a)` would probably be faster? I haven't tried it yet, though.

snowleopard:
@jitwit Good to have you back! Sorry it took me a while to take a careful look at the recent changes.

> The current iteration does comparatively well on most graphs. The exception is graphs with only a few sccs.

> Being able to detect the case of 1 scc gives a rather large speedup.

My intuition tells me that the problem is not a small number of SCCs, but the large size of some SCCs. Am I right? As an example, consider a graph with 1000 vertices, with one large component (of size 900) and 100 small components (of size 1). I'm pretty sure that in this case the performance will still be low even though there are many SCCs. Could you double-check?

> I think the main awkwardness is that having `NonEmpty.AM` as as vertices means that manipulating the outer `Acyclic.AM` map of the condensation is expensive. 

Indeed, there is an issue about this: #141.

> Returning a `Graph (NonEmpthy.AM a)` would probably be faster? 

I wouldn't be surprised, however, the resulting `Graph` will need to be converted to an adjacency map at some point, which brings us back to the problem.

snowleopard:
@jitwit Anyway, the benchmarking results are pretty good already! I wouldn't be strongly opposed to merging as is: I think we gain more than we sacrifice in terms of performance. However, if there is a way to speed up the remaining problematic cases that would be great.

jitwit:
Your intuition seems to be right given the example `vertices [0..1000] + circuit [1001..10000]`. I'm not sure how to interpret the results for the same graph overlayed with `[ edge x (x+1000) | x <- [0..1000]]` though. The goal of the second was to see what happens in the case where there are also many edges between components as well.

https://jitwit.github.io/benchmarks/criterion/scc_misc.html

I went back and looked at the worst benchmark's graph, which has 10000 vertices, 9999 of which are in the same strongly connected component.

> I wouldn't be surprised, however, the resulting Graph will need to be converted to an adjacency map at some point, which brings us back to the problem.

Makes sense!

snowleopard:
@jitwit For the problematic case, could you isolate the computationally expensive part? As far as I understand, the traversal of the original graph is not overly expensive, it's the cost of the final step, where we construct the resulting graph with heavy vertices. Is that right?

Returning to the `partition` function:

```haskell
partition :: Eq k => (a -> k) -> Graph a -> Graph (NonEmpty.Graph a)
```

Do you think a fast implementation of this function would allows us speed up the problematic case?

jitwit:
I looked at a few of the graphs under `+RTS -p` (twitter, worst nodes10000, and n5). For each, the most expensive search computation is `overlays` inside the sccs, second most is traversing the graph by Map.foldrWithKey, and third is the dfs search. 

`overlays` (inside sccs, not between) tends to take around 20-30+% of the overall computation. Worst case nodes10000 was above 30. Less than half of time is spent in dfs for each graph, and searching is light on allocations.

Building a list of vertices and edges and calling overlays afterwards was a fairly big win over building the graph by inserting each vertex and edge. 

After looking into this, I realized difference lists might help. There's a modest improvement by switching to `Endo [AM a]` while condensing, but the effect wasn't as large as I hoped (around 0-5% improvement depending on graph).

jitwit:
Full profiling results:

https://raw.githubusercontent.com/jitwit/bench-alga/master/nodes10000.prof
https://raw.githubusercontent.com/jitwit/bench-alga/master/n5.prof
https://raw.githubusercontent.com/jitwit/bench-alga/master/twitter.prof

jitwit:
> the main cost is the final step, where we construct the resulting graph with heavy vertices. Is that right?

Forgot to respond to questions. Yeah, it looks like this is right. 

Since a good portion of the time is spent in overlays, I think it may be difficult to improve much on this current approach. One idea I could try is the building the `Endo [AM a]` s with Arrays instead of IntMaps.

A good `partition` could be a way to improve, but I'm not sure how to implement it or how specifically it would help without further thought

jitwit:
Though within range of noise, the last two CI benches show `k =` 0.92 and 0.93 for `scc` -- likely thanks to difference lists!

snowleopard:
@jitwit Thank you for further profiling!

> A good `partition` could be a way to improve, but I'm not sure how to implement it or how specifically it would help without further thought

I had some ideas, let me try to remember and write them down.

> actually come to think of it, this wouldn't change much in the problematic case, it could be a general improvement though.

Are you sure? I think we agreed that the problematic case is where we have components of large size, which probably means that the current implementation is non-linear in the size of components, i.e. instead of being O(|s1|+|s2|+...+|sK|), it is something like O(|s1|^2 + |s2|^2 + ... + |sK|^2), perhaps with some log factors here and there. My hope is to come up with a faster `partition` function that will not suffer from such non-linearity, which would mostly improve the problematic cases.

I conjecture that in non-problematic cases, `partition` (or whatever we currently have in its place) does not take a large proportion of time.

> Though within range of noise, the last two CI benches show k = 0.92 and 0.93 for scc -- likely thanks to difference lists!

Nice! Note that we have a basic implementation of difference lists:

http://hackage.haskell.org/package/algebraic-graphs-0.4/docs/Algebra-Graph-Internal.html#t:List

Perhaps that might help make the code a bit easier to read. Could you reuse it?

jitwit:
I hadn't realized there was the internal implementation, done! It's definitely clearer than writing `Endo $ (++) xs`, though it does seem to require some type annotations. Is it possible to get rid of the type annotations?

Also, right! I only had the specific `nodes10000` benchmark in mind when I made that comment.

jitwit:
I figured out a way to classify the edges inside the dfs search. The benchmarks for most graphs are improved.

The first idea is that for edge `(u,v)` that if `v` gets classified as part of a `scc` either before or during the visit to `u`, `(u,v)` is an outer edge in the condensation. Otherwise it's an inner edge.

One complication is knowing which components the inner edges belong to. The second idea is that the preorder number can be used to track which queued inner edges should be added to a scc (like it is for vertices). 

It doesn't bridge the gap in the worst case benchmark, but it does decrease it. Before, the result was usually something like 220ms for current implementation vs 340 for new -- now I'm getting something like 220ms to 315ms.

jitwit:
Twitter graph is from 10.4/6.9s to 8.8/4.4s for am/aim

jitwit:
After the re-organizing, twitter down to 7.6 and 4.2 s!

Also to clarify why whether or not `dfs v` starts a new component determines if edge (u,v) is inner or outer, it's because sccs are selected based on the preorder number. On back edges, the boundary stack is popped to put the vertex with the lowest possible preorder number on top. If that's not `v`, then `u` is part of the same strongly connected component. If it is `v`, `u` is in a different one.

snowleopard:
> Twitter graph is from 10.4/6.9s to 8.8/4.4s for am/aim

> After the re-organizing, twitter down to 7.6 and 4.2 s!

Whoa, that's great! Our CI performance suite also shows a modest improvement.

> It doesn't bridge the gap in the worst case benchmark, but it does decrease it. Before, the result was usually something like 220ms for current implementation vs 340 for new -- now I'm getting something like 220ms to 315ms.

Is there any hope you can improve it further? 

Sorry, I didn't have time to think about  `partition`. I did look at the new implementation though and got lost pretty quickly :( The `StateSCC` datatype is scary. It would be nice to document, perhaps by adding a large note in a single comment, explaining what's going on. (Unless the current state is just work in progress.)

jitwit:
I started to add some comments to document the `StateSCC` data type. I opted to put them sort of in line with the fields in the type, but if it's preferable I can rearrange them to make the comments about it contiguous?

Would it be helpful to document more of the inner functions of the algorithm as well?

> Is there any hope you can improve it further?

Possibly, though I think the current version is potentially near a local max (unless I've missed something silly). I could still try the idea of using an array to tabulate the inner graphs, but that would complicate the type of the state monad. Otherwise, coming up with a way to identify which situations should use `induce` to build the condensation (like in the last implementation) could do the trick. I don't know what reasonable heuristics would be, though.

> (Unless the current state is just work in progress.)

I'm not sure if any of the fields are possible to eliminate without giving up some of the performance wins. The boundary stack stores vertices with their preorder ids to reduce the map lookups and the last three fields are used for constructing the condensation. The remaining fields are core to the Gabow's algorithm -- ints for giving vertices ids, tables for looking up the ids, and the two stacks.

snowleopard:
@jitwit Many thanks! I'll give the current version a thorough review soon.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -300,6 +304,158 @@ topSort g = runContT (evalStateT (topSort' g) initialState) id where
...
+        -- because they are not always required to compute the
+        -- condensation.
+      , inner_graphs  :: IntMap.IntMap (List AdjacencyIntMap)
+      , inner_edges   :: ![(PreorderId,(Int,Int))]
+      , outer_edges   :: [(Int,Int)]
+      } deriving (Show)
+
+gabowSCC :: AdjacencyIntMap -> State StateSCC ()
+gabowSCC g =
+  do let adjacent = IntSet.toList . flip postIntSet g
```
snowleopard: Could you avoid the construction of an intermediate list here? Instead of `forM_` you could probably use `IntSet.foldr`.

(It may well be the case that the intermediate list is optimised away, but it still feels better to use the set API directly.)

snowleopard:
@jitwit Happy New Year! I reviewed some code, but I can't say I fully understand the implementation yet. Sorry for taking so long.

Given the complexity of the new implementation, I think we need to add a test comparing the new `scc` with the standard `KL.scc`. This would give me a lot more confidence, at least until I fully internalise the code.

jitwit:
Hey, happy new year! Yeah, I wish the code was less convoluted. I'll try and find ways to make things a bit cleaner. Also adding tests to compare against `KL.scc` sounds good

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: That sounds good, but `adjacent` is used like `forM_ (adjacent v) ...` so I think we'd need some `foldM`, which I don't think is exported by either set module? Might it be worth adding `{-# inline ... #-}` declarations?

jitwit:
I sketched together some notes about the implementation. Hopefully they can help clarify the approach or highlight things needing improvement:

https://github.com/jitwit/alga-notes/blob/master/gabow.org

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
snowleopard: Do you mean you need something like `forEach`/`forEachInt`?

```haskell
forEach :: Applicative f => Set a -> (a -> f b) -> f ()
forEach set f = Set.foldr (\x y -> f x *> y) (pure ()) set

... forEach (postSet g u) $ \v -> ...
```

If yes, feel free to add these auxiliary functions into `Algebra.Graph.Internal`.

snowleopard: (Or their strict versions if need be.)

snowleopard:
@jitwit Excellent notes! Thanks for doing this. I only got lost a bit where three vertices `u`, `v` and `w` are involved -- in those cases a diagram would help a lot :-)

Please link to your doc from the comments.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: Sounds good! I added them as `setForEach_` and `intsetForEach_`. Maybe the under bars are not needed/too ugly?

snowleopard: I would just go for `forEach` and `forEachInt` -- these names should read well in the code.

jitwit:
Forgot to add links, will do in the next pass! I also made some kind of `Diagram` showing a pass of the algorithm on `3*1*4*1*5`:
https://raw.githubusercontent.com/jitwit/alga-notes/master/images/gabow.png
gray means unvisited, red visited, blue assigned to a component. vertices labelled with vertex,preorder,scc triple. edges green or black if found to be an edge of the condensation.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -300,6 +304,162 @@ topSort g = runContT (evalStateT (topSort' g) initialState) id where
 isAcyclic :: AdjacencyIntMap -> Bool
 isAcyclic = isRight . topSort
 
+-- TODO: Benchmark and optimise.
```
snowleopard: Is this TODO still relevant? As far as I can tell, you've done both!

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
```diff
@@ -300,6 +304,162 @@ topSort g = runContT (evalStateT (topSort' g) initialState) id where
...
+-- | Compute the /condensation/ of a graph, where each vertex corresponds to a
+-- /strongly-connected component/ of the original graph. Note that component
+-- graphs are non-empty, and are therefore of type
+-- "Algebra.Graph.NonEmpty.AdjacencyMap".
```
snowleopard: Should this be `AdjacencyIntMap`?

snowleopard: In fact, I'm confused: are the inner graphs non-empty?

snowleopard: Answering my own question: we don't seem to have an data type for non-empty adjacency maps with `Int` vertices, so I guess this requires further thinking.

snowleopard: There are three options: 

1. Keep `scc` only in `AdjacencyMap`: this is easy but we lose some specialised code.
2. Make `scc` in `AdjacencyIntMap` have a different type: this is also easy, but is a bit awkward and probably not want we want in the long term.
3. Add a new module `NonEmpty.AdjacencyIntMap` and the corresponding `scc`. This is a lot of work but seems to be the right solution in the long term.

@jitwit How about we go for option (1) right now, but then you attempt (3) in a separate PR?

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
```diff
@@ -321,20 +327,100 @@ isAcyclic = isRight . topSort
...
+
+data StateSCC a
+  = C { current       :: !Int
+      , componentId   :: !Int
+      , boundary      :: ![(Int,a)]
+      , dfsPath       :: ![a]
+      , preorders     :: !(Map.Map a Int)
+      , components    :: !(Map.Map a Int)
+      , inner_graphs  :: (IntMap.IntMap (List (AdjacencyMap a)))
+      , inner_edges   :: ![(Int,(a,a))]
```
snowleopard: Are you sure all the tuples here (and in other fields) have the right strictness? While the list itself is marked as strict, the inner tuples may turn out to be lazy. Is this what you want, or are you sure GHC correctly optimises laziness away?

snowleopard: I'm not sure at all if strictness matters in this case, but in the `Int` version we would probably want to at least unpack all these tuples, right? Otherwise, we may be constantly chasing pointers.

snowleopard:
@jitwit Thanks! The diagram does help.

I've made another pass with comments.

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: True!

src/Algebra/Graph/AdjacencyIntMap/Algorithm.hs
jitwit: Yeah, I had considered opening a pr to add the analogous `NonEmpty.AdjacencyIntMap` so that the types can match, but held off. (1) then (3) makes sense to me

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
jitwit: I'm not sure if the amount of laziness makes sense. I experimented with using `StrictPair` and saw a fairly big regression in the benchmarks. I then tried changing the fields (except counters) to lazy and saw an improvement across the board!

jitwit: My original reasoning was along the lines of if a whole structure is eventually consumed, it is faster when strict

snowleopard: Interesting! I guess one example where "if a whole structure is eventually consumed, it is faster when strict" doesn't work is when intermediate lists are eliminated thanks to list fusion. Perhaps, this is what happened here.

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
```diff
@@ -321,20 +326,100 @@ isAcyclic = isRight . topSort
 -- 'isAcyclic' x     == (scc x == 'gmap' NonEmpty.'NonEmpty.vertex' x)
 -- @
 scc :: Ord a => AdjacencyMap a -> AdjacencyMap (NonEmpty.AdjacencyMap a)
-scc m = gmap (component Map.!) $ removeSelfLoops $ gmap (leader Map.!) m
+scc g = condense g $ execState (gabowSCC g) initialState where
+  initialState = C 0 0 [] [] Map.empty Map.empty IntMap.empty [] []
+
+data StateSCC a
+  = C { current       :: !Int
```
snowleopard: Please correct me if I'm wrong, but I think GHC doesn't automatically unpack strict fields, which means this (and the next) `Int` will still be represented by a pointer that is guaranteed to point to an evaluated integer.

If my understanding is correct, I think one could use `{-# UNPACK #-}` here to ask GHC to unpack the contents of the field `current` into the constructor `C` itself, removing one level of indirection.

https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/glasgow_exts.html#unpack-pragma

Perhaps, this doesn't matter in this case? Then it's fine to leave the code as is but I'm curious to know if this is the case.

jitwit: Ah yes, it probably doesn't matter much (it's a very small part of the work), but I figured it couldn't hurt to add them.

jitwit:
I also added a CPP statement to get rid of the redundant `Data.Monoid` import warning for post 8.2 ghc

snowleopard:
> I also added a CPP statement to get rid of the redundant `Data.Monoid` import warning for post 8.2 ghc

I think you can just import `Data.Semigroup ((<>))` instead, getting rid of CPP. I dropped support for GHC < 8 recently hoping to eliminate most occurrences of CPP.

As long as there are no warnings in GHC 8.6.5, I'm happy!

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
```diff
@@ -321,19 +326,98 @@ isAcyclic = isRight . topSort
...
+      (\(SCC pre scc bnd pth pres sccs gs es_i es_o) ->
+         let gs' = IntMap.insert scc g_i gs
+             sccs' = List.foldl' (\sccs x -> Map.insert x scc sccs) sccs curr
+             scc' = scc + 1
+             bnd' = tail bnd
+             p_v = fst $ head bnd
+             g_i = fromList (uncurry edge.snd <$> es) <> fromList (vertex <$> curr)
+             (es,es_i') = span ((>=p_v).fst) es_i
+             pth' = tail $ dropWhile (/=v) pth
+             curr = v:takeWhile(/=v) pth
```
snowleopard: I think you could use `span` or `break` here instead of traversing the list twice with `dropWhile` and `takeWhile`.

snowleopard: I would also appreciate if you reorder this `let` sequence in the order of execution, otherwise, it's difficult to read.

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
```diff
@@ -321,19 +326,98 @@ isAcyclic = isRight . topSort
...
+                return newComponent
+
+    insertComponent v = modify'
+      (\(SCC pre scc bnd pth pres sccs gs es_i es_o) ->
+         let gs' = IntMap.insert scc g_i gs
+             sccs' = List.foldl' (\sccs x -> Map.insert x scc sccs) sccs curr
+             scc' = scc + 1
+             bnd' = tail bnd
+             p_v = fst $ head bnd
+             g_i = fromList (uncurry edge.snd <$> es) <> fromList (vertex <$> curr)
```
snowleopard: Here you are constructing a lot of small intermediate adjacency maps (`vertex` and `edge`) and place them in a list. I think it should be more efficient to instead use something like `edges` to construct the adjacency map in one go. 

Also, I'm a bit confused about the role of `curr`. I think the only situation where not all vertices will appear in `es` is when the component contains just a single vertex. Is that correct? If yes, then it would be better to treat this as a special case and either create an adjacency map using `edges` or using a single call to `vertex`.

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
```diff
@@ -321,19 +326,98 @@ isAcyclic = isRight . topSort
...
+  initialState = SCC 0 0 [] [] Map.empty Map.empty IntMap.empty [] []
+
+data StateSCC a
+  = SCC { preorder      :: {-# unpack #-} !Int
+        , component     :: {-# unpack #-} !Int
+        , boundaryStack :: [(Int,a)]
+        , pathStack     :: [a]
+        , preorders     :: Map.Map a Int
+        , components    :: Map.Map a Int
+        , innerGraphs   :: IntMap.IntMap (List (AdjacencyMap a))
```
snowleopard: As clarified in my comment below, I think we don't need to store a lot of small adjacency maps here and instead should just have something liks `innerGraphs :: IntMap.IntMap (AdjacencyMap a)`. I don't see what `List` brings us here.

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
```diff
@@ -321,19 +326,98 @@ isAcyclic = isRight . topSort
...
-removeSelfLoops :: Ord a => AdjacencyMap a -> AdjacencyMap a
-removeSelfLoops m = foldr (\x -> removeEdge x x) m (vertexList m)
+condense :: Ord a => AdjacencyMap a -> StateSCC a -> AdjacencyMap (NonEmpty.AdjacencyMap a)
+condense g (SCC _ n _ _ _ assignment inner _ outer)
+  | n == 1 = vertex $ convert g
+  | otherwise = gmap (inner' IntMap.!) outer'
+  where inner' = convert . overlays . toList <$> inner
+        outer' = overlays $ es ++ vs
+        vs = vertex <$> [0..n-1]
+        es = [ edge (sccid x) (sccid y) | (x,y) <- outer ]
```
snowleopard: Similarly to the above comment: it's more efficient to construct a large adjacency map using `edges` rather than via an `overlays` of `edge`s. And we also have `vertices`, which is more efficient than an `overlays` of `vertex`s.

snowleopard:
@jitwit I think I've noticed a few more places where we can save some time. Could you please have a look?

jitwit:
Much better results!

[https://jitwit.github.io/benchmarks/criterion/scc-bench.html](https://jitwit.github.io/benchmarks/criterion/scc-bench.html)

Just building the inner graphs (no difference lists) makes a big difference.

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
jitwit: Yeah, you're definitely right! I changed it to check if the edge list is null, and in that case just insert a `vertex v`.

Not building the intermediate lists makes a big difference to performance, too

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
jitwit: The shameful reason I originally wrote with `takeWhile`+`dropWhile` was to avoid the incomplete pattern match. What's needed is something like `splitAfter (==v)` which would put `v` inside the first field of the tuple. That, or a way to express that `v` is in the list with types, which I'm not sure how

src/Algebra/Graph/AdjacencyMap/Algorithm.hs
jitwit: It would be nice to exploit that `[0..n-1]` is sorted, but that would need an extra function `verticesAsc` or for `AM` to be exported?

snowleopard: Yes, I also thought about this. In this particular case, the benefit is probably not going to be significant, but in general, such optimised functions could be helpful.

I'm not very enthusiastic about adding even more functions into the API though, which seems quite overloaded already. I suggest opening an issue to discuss this further.

snowleopard:
@jitwit Excellent! And the code got simpler too. Looks like we're mostly beating KL now, with only one scenario where we're slightly behind. I think we can tolerate this.

I think we can merge the PR now unless you'd like to do any further improvements.

jitwit:
On larger graphs the performance gain is even bigger. An acyclic facebook network graph gets 200ms to 600ms (4000 vertices 90000 edges) and the twitter one gets 6.4s to 60s (80000 vertices 1,700,000 edges)

The part I'd maybe want to improve is the incomplete pattern on `(curr,_:pth') = span (==v) pathStack`. Any preference for how, extra helper, or using `tail` or something else?

jitwit:
On larger graphs the performance gain is even bigger. An acyclic facebook network graph gets 200ms to 600ms (4000 vertices 90000 edges) and the twitter one gets 6.4s to 60s (80000 vertices 1,700,000 edges)

The part I'd maybe want to improve is the incomplete pattern on `(curr,_:pth') = span (==v) pathStack`. Any preference for how, extra helper, or using `tail` or something else?

jitwit:
On larger graphs the performance gain is even bigger. An acyclic facebook network graph gets 200ms to 600ms (4000 vertices 90000 edges) and the twitter one gets 6.4s to 60s (80000 vertices 1,700,000 edges)

The part I'd maybe want to improve is the incomplete pattern on `(curr,_:pth') = span (==v) pathStack`. Any preference for how, extra helper, or using `tail` or something else?

snowleopard:
> 200ms to 600ms [...] and 6.4s to 60s

@jitwit This is pretty cool! I guess when you're benchmarking KL you are including the overhead of converting an adjacency map to an array representation. What happens if you start with a graph stored in KL-style array? Is your implementation still significantly faster?

I don't really mind `(curr,_:pth') = span (/=v) pth`. I think going via an intermediate function would just obscure the invariant. One slight improvement is perhaps `(curr,_v:pth') = span (/=v) pth`, i.e. giving the unused binding a name.

jitwit:
I haven't done a proper benchmark yet, but from a quick profile quite a lot of the time is spent in `induce` in the old implementation on the twitter graph

snowleopard:
@jitwit Note that your implementation uses `gmap` which might also be quadratic when mapping into vertices with expensive `Ord` instance (as in our case). The complexity of `gmap` given in the docs does not take into account the cost of `Ord`. So, I'm not entirely sure that the worst-case complexity really became subquadratic. But even if it is quadratic in the worst case, it looks like the typical case got much better!

jitwit:
It looks like the `Ord` instance for `Map`s uses `compare ``on`` toAscList`. Because the vertices of the components are disjoint, this will always classify from just the head of the lists, right? Is `toAscList` lazy enough for this to basically have complexity of the height of the left spine of the tree?

jitwit:
Given `n,m` and `n_o,m_o` the number of components and edges between them, I think the complexity is something like:

* `O((n+m)*log n)` for the traversal (constructing the inner graphs takes `SUM (n_c+m_c)*log(n_c)` over components, which is subsumed by `(n+m)*log n`. 
* `O((n_o+m_o)*log n_o*log (n-n_o))` for building the condensation. With the new change to arrays lookup is `O(1)`. Largest possible component size is `(n-n_o)`, so multiplying by `log(n-n_o)` accounts for the comparisons (based on my previous comment)?

jitwit:
Ah! Sorry, should have checked the `Ord` instance. Neat! It seems like this is a case where it pays to use `Data.Map` over `Data.IntMap` since `vertexCount` would be `O(n)` otherwise. I'll just go ahead and put `O((n+m) * log n)` for the time complexity in this case, as you suggest

jitwit:
Should I delete the  old`scc` implementation in `Data.Graph.Typed` as well as the test which compares them?

snowleopard:
> Neat! It seems like this is a case where it pays to use `Data.Map` over `Data.IntMap` since `vertexCount` would be `O(n)` otherwise. 

Indeed, I forgot that `IntMap` doesn't store the size! I guess we could cache it in `AdjacencyIntMap` ourselves. We could also cache a few other fields like `vertexSet`, `vertexList`, etc. Thanks to laziness the overhead would probably be acceptable. (Of course, this should better be done in a separate PR.)

> Should I delete the old `scc` implementation in `Data.Graph.Typed` as well as the test which compares them?

No, let's keep it. It's always good to have a good testsuite. I'm pretty sure we'll be doing further tweaks to the implementation and it's good to know that (at least some) bugs will be caught.

jitwit:
Makes sense! And yeah, would be interesting to see if caching for `AIM`'s could be beneficial

snowleopard:
@jitwit OK, merged! Many thanks for your hard work and for tolerating such long reviews :-)

PR #236, opened 2019-09-20
Removed Ord and Num of Acyclic and introduced shrink

adithyaov:
Introduced shrink for ease in writing examples as `Num` is no longer available.
For example, (1 + 2 * 3) can be changed to (shrink $ 1 + 2 * 3).
Could maybe change the `Show` instance as well. `fromMaybe empty . toAcyclic == shinrk` if the graph is `Acyclic`.


snowleopard:
Many thanks @adithyaov! This looks good. My apologies for taking so long to review.

PR #237, opened 2019-09-21
Implement complement for undirected graphs

### LINKED ISSUE #21: Implement graph complement
(I think this is mostly relevant for undirected graphs.)

_Graph complement_ of an undirected graph (V, E) is (V, K \ E), where K is the clique on all vertices V, e.g. see https://en.wikipedia.org/wiki/Complement_graph.

Interestingly, complements of sparse graphs have as compact algebraic representations as sparse graphs.

gabrielelana:
Regarding the [last tests](https://github.com/gabrielelana/alga/blob/undirected-complement/test/Algebra/Graph/Test/Undirected.hs#L83) we don't know if it's ok for Arbitrary to generate graphs with loops

Fixes #21

Proudly made by @sphaso and @gabrielelana at "Open Source Saturday Milano"

[Open Source Saturday](https://www.meetup.com/it-IT/Open-Source-Saturday-Milano/)

### DIFF
--- a/src/Algebra/Graph/Undirected.hs
+++ b/src/Algebra/Graph/Undirected.hs
@@ -701,13 +701,11 @@ toSymmetricRelation = foldg SR.empty SR.vertex SR.overlay SR.connect
 {-# INLINE toSymmetricRelation #-}
 
 -- | Complement of a graph.
--- Complexity: /O(m^2+n)/ time, /O(m+n)/ memory where
+-- Complexity: /O(m^2)/ time, /O(m^2)/ memory where
 --
 -- @
 -- complement 'empty'           == 'empty'
 -- complement ('vertex' x)      == ('vertex' x)
--- complement ('edge' x y)      == ('vertices' [x, y])
--- complement ('star' x [y, z]) == ('overlay' ('vertex' x) ('edge' y z))
 -- complement . complement    == id
 -- @
 complement :: Ord a => Graph a -> Graph a
--- a/src/Algebra/Graph/Undirected.hs
+++ b/src/Algebra/Graph/Undirected.hs
@@ -701,11 +701,14 @@ toSymmetricRelation = foldg SR.empty SR.vertex SR.overlay SR.connect
 {-# INLINE toSymmetricRelation #-}
 
 -- | Complement of a graph.
--- Complexity: /O(m^2)/ time, /O(m^2)/ memory where
+-- Complexity: /O(n^2*m)/ time, /O(n^2)/ memory where
 --
 -- @
 -- complement 'empty'           == 'empty'
 -- complement ('vertex' x)      == ('vertex' x)
+-- complement ('edge' 1 2)      == ('vertices' [1, 2])
+-- complement ('edge' 0 0)      == ('edge' 0 0)
+-- complement ('star' 1 [2, 3]) == ('overlay' ('vertex' 1) ('edge' 2 3))
 -- complement . complement    == id
 -- @
 complement :: Ord a => Graph a -> Graph a
--- a/src/Algebra/Graph/Undirected.hs
+++ b/src/Algebra/Graph/Undirected.hs
@@ -56,6 +56,7 @@ import           Control.Monad                    (MonadPlus (..))
 import           Data.Coerce
 import           GHC.Generics
 import           Algebra.Graph.ToGraph            (toGraph)
+import           Data.List                        ((\\))
 
 import qualified Algebra.Graph                    as G
 
@@ -64,6 +65,7 @@ import qualified Control.Applicative              as Ap
 import qualified Data.IntSet                      as IntSet
 import qualified Data.Set                         as Set
 import qualified Data.Tree                        as Tree
+
 
 {-| The Undirected 'Graph' data type is an abstraction over the 'Graph' data
    type and provides the same graph construction
@@ -699,6 +701,9 @@ toSymmetricRelation = foldg SR.empty SR.vertex SR.overlay SR.connect
 {-# INLINE toSymmetricRelation #-}
 
 -- | Complement of a graph.
+-- Complexity: /O(E^2+V)/ time, /O(E+V)/ memory where
+-- E is the number of edges and V is the number of vertices
+-- The quadratic bound is due to `edges`.
 --
 -- @
 -- complement 'empty'           == 'empty'
@@ -708,8 +713,12 @@ toSymmetricRelation = foldg SR.empty SR.vertex SR.overlay SR.connect
 -- complement . complement      == id
 -- @
 complement :: Ord a => Graph a -> Graph a
-complement g@(UG _) = foldr (uncurry removeEdge) (cliqueG g) (edgeList g)
-  where cliqueG = clique . vertexList
+complement g@(UG _) = overlay (vertices allVertices) (edges complementEdges)
+ where cliqueG = clique . vertexList
+       allVertices = vertexList g
+       previousEdges = edgeList g
+       loops = filter (uncurry (==)) previousEdges
+       complementEdges = loops ++ (edgeList (cliqueG g) \\ previousEdges)
 
 -- | The /path/ on a list of vertices.
 -- Complexity: /O(L)/ time, memory and size, where /L/ is the length of the
--- a/src/Algebra/Graph/Undirected.hs
+++ b/src/Algebra/Graph/Undirected.hs
@@ -701,15 +701,14 @@ toSymmetricRelation = foldg SR.empty SR.vertex SR.overlay SR.connect
 {-# INLINE toSymmetricRelation #-}
 
 -- | Complement of a graph.
--- Complexity: /O(E+V)/ time, /O(E+V)/ memory where
--- E is the number of edges and V is the number of vertices.
+-- Complexity: /O(m^2+n)/ time, /O(m+n)/ memory where
 --
 -- @
 -- complement 'empty'           == 'empty'
 -- complement ('vertex' x)      == ('vertex' x)
--- complement ('edge' x y)      == ('overlay' ('vertex' x) ('vertex' y))
+-- complement ('edge' x y)      == ('vertices' [x, y])
 -- complement ('star' x [y, z]) == ('overlay' ('vertex' x) ('edge' y z))
--- complement . complement      == id
+-- complement . complement    == id
 -- @
 complement :: Ord a => Graph a -> Graph a
 complement g@(UG _) = overlay (vertices allVertices) (edges complementEdges)
--- a/src/Algebra/Graph/Undirected.hs
+++ b/src/Algebra/Graph/Undirected.hs
@@ -43,7 +43,7 @@ module Algebra.Graph.Undirected (
 
     -- * Graph transformation
     removeVertex, removeEdge, replaceVertex, mergeVertices,
-    induce, induceJust, toSymmetricRelation,
+    induce, induceJust, toSymmetricRelation, complement,
 
     -- * Miscellaneous
     consistent
@@ -697,6 +697,19 @@ adjacencyList = SR.adjacencyList . toSymmetricRelation
 toSymmetricRelation :: Ord a => Graph a -> SR.Relation a
 toSymmetricRelation = foldg SR.empty SR.vertex SR.overlay SR.connect
 {-# INLINE toSymmetricRelation #-}
+
+-- | Complement of a graph.
+--
+-- @
+-- complement 'empty'           == 'empty'
+-- complement ('vertex' x)      == ('vertex' x)
+-- complement ('edge' x y)      == ('overlay' ('vertex' x) ('vertex' y))
+-- complement ('star' x [y, z]) == ('overlay' ('vertex' x) ('edge' y z))
+-- complement . complement      == id
+-- @
+complement :: Ord a => Graph a -> Graph a
+complement g@(UG _) = foldr (uncurry removeEdge) (cliqueG g) (edgeList g)
+  where cliqueG = clique . vertexList
 
 -- | The /path/ on a list of vertices.
 -- Complexity: /O(L)/ time, memory and size, where /L/ is the length of the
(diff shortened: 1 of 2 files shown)


snowleopard:
@gabrielelana @sphaso Thank you! :)

Could you estimate the complexity of your implementation in terms of time and memory, as well as the size of the resulting expression? I suspect it can only handle very small graphs because of the repeated use of expensive `removeEdge`, but I may be wrong.

Of course, having an expensive algorithm is much better than having none, so I'm still happy to merge your PR, but I'd like to document the complexity.

> we don't know if it's ok for Arbitrary to generate graphs with loops

I think the answer is Yes because we don't rule out self-loops. It is not entirely clear how `complement` should handle self-loops. What about just not touching them at all? This allows you to keep the nice property `complement . complement = id`.

sphaso:
Hi @snowleopard , thanks for getting back :)

> I suspect it can only handle very small graphs because of the repeated use of expensive removeEdge, but I may be wrong.

If you deem `removeEdge` too expensive there are alternative strategies we considered: simply using `\\` on the edgeList was one, going through the symmetricRelation sets was another. Our first idea was to XOR the adjancency matrix with a 1 mask, but we went for what we believed was the most "idiomatic" solution

> > we don't know if it's ok for Arbitrary to generate graphs with loops

> I think the answer is Yes because we don't rule out self-loops.

We were indeed a bit puzzled by loops, because then one might start thinking about multigraphs, and the property `complement . complement = id` won't pass for those either. The only way to handle loops I can think of, would be to re-apply them on the clique, but that can't work for other types of multigraphs.

As you said, an expensive algorithm is better than none, so we're happy to analyze the complexity of `complement` and keep these as food for thought depending on how you feel about it.

snowleopard:
> If you deem removeEdge too expensive

It's linear w.r.t. to the size of the graph expression, which is quite expensive:

http://hackage.haskell.org/package/algebraic-graphs-0.4/docs/Algebra-Graph.html#v:removeEdge

> simply using `\\` on the edgeList

I think filtering existing edges may indeed be faster, although it would be nice to do some benchmarking. Could you compare your approach with `\\`?

> We were indeed a bit puzzled by loops, because then one might start thinking about multigraphs

The basic `Graph` data type corresponds to "classic" graphs represented by a pair (V, E), where self-loops are allowed, so we should be able to handle self-loops.

> The only way to handle loops I can think of, would be to re-apply them on the clique, but that can't work for other types of multigraphs.

Why don't you simply leave self-loops as is, without removing them? As far as I can see, this is the simplest option.

> As you said, an expensive algorithm is better than none, so we're happy to analyze the complexity of `complement`

Great, please do! And some benchmarking would be nice too: I'm curious how your current implementation based on `removeEdge` compares to the one based on `\\`.

sphaso:
> Why don't you simply leave self-loops as is, without removing them? As far as I can see, this is the simplest option.

I feel there might be a critical piece of information regarding this library I'm missing. The problem I see is that:
- `Arbitrary Graph` produces graphs with self-loops
- `clique` does not (e.g. `clique [x,y,z]    == 'edges' [(x,y), (x,z), (y,z)]`)

So the situation is: I want `complement . complement = id` to be a valid property even for graphs with self-loops. However, if I use `clique` as the reference graph for the complement, I don't have the self-loops in the first place.
That's what I meant by "re-apply". If the pseudo code for the `complement` function is `clique edges \\ current edges`, I would have to take the self-loops from `current edges` and re-apply them after the difference operation.

We'll work on the benchmark ASAP.

sphaso:
Hi @snowleopard ,
we finally ran the benchmark and you were right! `Data.List.\\` is at least one order of magnitude faster than `removeEdge`.

[Here](https://github.com/gabrielelana/alga/blob/feature/benchmark-complement/app/Main.hs) you can find the benchmark code.
[Here](https://github.com/gabrielelana/alga/blob/feature/benchmark-complement/BENCHMARK.md) you can find the results.

For easier reference, these are the two differing implementations. We took some time to calculate the complexity of the faster one as well as handling loops:

 ```haskell
complement :: Ord a => Graph a -> Graph a
complement g@(UG _) = overlay (edges loops) $ foldr (uncurry removeEdge) (cliqueG g) previousEdges
  where cliqueG = clique . vertexList
        previousEdges = edgeList g
        loops = filter (uncurry (==)) previousEdges

-- Complexity: /O(E^2+V)/ time, /O(E+V)/ memory where
-- E is the number of edges and V is the number of vertices
-- The quadratic bound is due to `edges`.
complement' :: Ord a => Graph a -> Graph a
complement' g@(UG _) = overlay (vertices allVertices) (edges complementEdges)
  where cliqueG = clique . vertexList
        allVertices = vertexList g
        previousEdges = edgeList g
        loops = filter (uncurry (==)) previousEdges
        complementEdges = loops ++ (edgeList (cliqueG g) \\ previousEdges)
```

We were curious about the time complexity of `edges` and we ran a second benchmark on it. The benchmark results seem to show that `edges` is actually linear with regard to time.
[Here](https://github.com/gabrielelana/alga/blob/feature/benchmark-edges/app/Main.hs) the benchmark code.
[Here](https://github.com/gabrielelana/alga/blob/feature/benchmark-edges/BENCHMARK.md) the results.

What do you think?

If you like our implementation we'll move it to the current PR.
Would you like to keep the benchmarks as well?

**Comment by snowleopard** [comment]
> we finally ran the benchmark and you were right! Data.List.\\ is at least one order of magnitude faster than removeEdge.

@sphaso Great, thank you for doing this!

> We were curious about the time complexity of `edges` and we ran a second benchmark on it. The benchmark results seem to show that `edges` is actually linear with regard to time.

Yes, `edges` is indeed linear with respect to the length of the given list, as stated in the documentation:

http://hackage.haskell.org/package/algebraic-graphs-0.4/docs/Algebra-Graph.html#v:edges

Did you expect its complexity to be non-linear?

> If you like our implementation we'll move it to the current PR.

Yes, let's use the faster implementation based on `\\`. 


Responding to your earlier comment: yes, I think what you call "re-applying" will work. Conceptually, it's not very complex: you complement all non-loop edges and make sure to keep the original self-loops in the final graph.

**Comment by snowleopard** [comment]
> Would you like to keep the benchmarks as well?

Regarding benchmarks: there is a separate repository for benchmarks, which you can find here:

https://github.com/haskell-perf/graphs

**Comment by sphaso** [comment]
> Yes, `edges` is indeed linear with respect to the length of the given list [...] Did you expect its complexity to be non-linear?

When estimating the complexity of `complement` we looked at the docs of `edges` inside the `Undirected` module [where it's stated](https://github.com/snowleopard/alga/blob/901bf6b020f8b1ab7871d932bff81ac345c7bf1c/src/Algebra/Graph/Undirected.hs#L408):     
> Complexity: /O(L^2)/ time, /O(L)/ memory and size

We'll adjust our estimates of `complement` accordingly. Would you like us to amend this comment as well?

**Comment by snowleopard** [comment]
@sphaso Oh, that's just wrong. If you correct it to "Complexity: _O(L)_ time, memory and size, where _L_ is the length of the given list." right in this PR that would be great!

**Comment by gabrielelana** [comment]
@snowleopard we ported the faster implementation and amended the complexity of `edges`, can you check if everything is ok?

File: src/Algebra/Graph/Undirected.hs [hunk 1]
```diff
@@ -698,6 +700,25 @@ toSymmetricRelation :: Ord a => Graph a -> SR.Relation a
 toSymmetricRelation = foldg SR.empty SR.vertex SR.overlay SR.connect
 {-# INLINE toSymmetricRelation #-}
 
+-- | Complement of a graph.
+-- Complexity: /O(E+V)/ time, /O(E+V)/ memory where
```
snowleopard: Please use `n` and `m` to denote the number of vertices and edges, respectively, as in the rest of the docs.

src/Algebra/Graph/Undirected.hs
```diff
@@ -698,6 +700,25 @@ toSymmetricRelation :: Ord a => Graph a -> SR.Relation a
...
+-- E is the number of edges and V is the number of vertices.
+--
+-- @
+-- complement 'empty'           == 'empty'
+-- complement ('vertex' x)      == ('vertex' x)
+-- complement ('edge' x y)      == ('overlay' ('vertex' x) ('vertex' y))
+-- complement ('star' x [y, z]) == ('overlay' ('vertex' x) ('edge' y z))
+-- complement . complement      == id
```
snowleopard: I suspect this line will render misaligned in Haddock. Could you build the docs and double-check?

src/Algebra/Graph/Undirected.hs
snowleopard: Also, I don't think the complexity bounds are correct: you build up /O(n^2)/ edges in `edgeList (cliqueG g)`!

snowleopard:
@gabrielelana Many thanks! The implementation looks good, but see my comments regarding the docs; they should be easy to address.

src/Algebra/Graph/Undirected.hs
```diff
@@ -698,6 +700,25 @@ toSymmetricRelation :: Ord a => Graph a -> SR.Relation a
...
 {-# INLINE toSymmetricRelation #-}
 
+-- | Complement of a graph.
+-- Complexity: /O(E+V)/ time, /O(E+V)/ memory where
+-- E is the number of edges and V is the number of vertices.
+--
+-- @
+-- complement 'empty'           == 'empty'
+-- complement ('vertex' x)      == ('vertex' x)
+-- complement ('edge' x y)      == ('overlay' ('vertex' x) ('vertex' y))
```
snowleopard: You can simplify this by using `vertices [x, y]`.

sphaso:
@snowleopard I should have addressed all the points you mentioned in the latest comments. Good catch on the quadratic complexity!

src/Algebra/Graph/Undirected.hs
```diff
@@ -698,6 +700,24 @@ toSymmetricRelation :: Ord a => Graph a -> SR.Relation a
 toSymmetricRelation = foldg SR.empty SR.vertex SR.overlay SR.connect
 {-# INLINE toSymmetricRelation #-}
 
+-- | Complement of a graph.
+-- Complexity: /O(m^2+n)/ time, /O(m+n)/ memory where
```
snowleopard: This still seems incorrect. The complexity is O(n^2) both in terms of time and memory, isn't it?

test/Algebra/Graph/Test/Undirected.hs
```diff
@@ -64,9 +66,25 @@ testUndirected = do
     test "edgeCount   . fromUndirected <= (*2) . edgeCount" $ \(x :: G) ->
           (G.edgeCount . fromUndirected) x <= ((*2) . edgeCount) x
 
+    putStrLn $ "\n============ Graph.Undirected.complement ================"
+
+    test "complement empty              == empty" $
+          complement empty              == (empty :: UGI)
+
+    test "complement (vertex 1)         == (vertex 1)" $
```
snowleopard: The properties are stated for arbitrary values `x`, `y` and `z` but the tests are only for fixed values for some reason. Let's QuickCheck the properties exactly as they are written in Haddock.

snowleopard:
@sphaso Thanks for the fixes! The result doesn't seem completely right though, please see my comments.

src/Algebra/Graph/Undirected.hs
sphaso: Do you mean `O(m^2)`? my understanding was that `m` was the edge count, so by creating a clique we're increasing the number of edges while the number of vertices is constant.

sphaso:
@snowleopard I pushed a new commit that takes care of at least one concern, see my answer above for the other (time \ memory estimates, for which I already updated it as `m^2`). Regarding tests, the fact that we're now preserving self-loops rendered two properties false. Good thing we swapped those fixed values :)
Namely:
`complement (edge x y)         == (vertices [x, y] :: UGI)`
is no longer true if `x == y`.

src/Algebra/Graph/Undirected.hs
snowleopard: @sphaso No, I do mean O(n^2).

If we start with a graph having n = 100 vertices and, say, m = 1000 edges, the `cliqueG` will contain ~100^2 edges, not ~1000^2 edges! After all, a graph on 100 vertices can't have more than ~100^2 edges :-)

Interestingly, O(m^2) is not even an upper bound! For if you start with a graph having n = 100 vertices and m = 0 edges, the `cliqueG` will still contain ~100^2 edges, no ~0^2 edges :-)

snowleopard: Digging deeper into the code: while O(n^2) is the right memory complexity, the time complexity might turn out to be O(n^2 * m) if the list difference `\\` operation is quadratic.

snowleopard:
> Namely:
> `complement (edge x y) == (vertices [x, y] :: UGI)`
> is no longer true if x == y.

@sphaso Indeed :-) Thanks!

src/Algebra/Graph/Undirected.hs
```diff
@@ -701,13 +701,11 @@ toSymmetricRelation = foldg SR.empty SR.vertex SR.overlay SR.connect
...
 
 -- | Complement of a graph.
--- Complexity: /O(m^2+n)/ time, /O(m+n)/ memory where
+-- Complexity: /O(m^2)/ time, /O(m^2)/ memory where
 --
 -- @
 -- complement 'empty'           == 'empty'
 -- complement ('vertex' x)      == ('vertex' x)
--- complement ('edge' x y)      == ('vertices' [x, y])
--- complement ('star' x [y, z]) == ('overlay' ('vertex' x) ('edge' y z))
```
snowleopard: These were rather helpful examples, even if incorrect :) Perhaps, we should use fixed values `1`, `2` and `3` both here and in the tests.

src/Algebra/Graph/Undirected.hs
sphaso: @snowleopard makes sense, sorry for the many mistakes in this part.

src/Algebra/Graph/Undirected.hs
sphaso: I'll use them with fixed values, highlighting the gotcha for x == y

sphaso:
@snowleopard hopefully I got everything right this time :) thanks!

snowleopard:
@sphaso Thank you, merged! :)

By the way, feel free to add yourself and @gabrielelana to the list of contributors:

https://github.com/snowleopard/alga/blob/master/AUTHORS.md

PR #238, opened 2019-10-08
GHC 8.8 support

fosskers:
The rewrite tests are tricky. Giving `inspection-testing` a stricter lower bound fixes one issue, but introduces another.

### DIFF
--- a/algebraic-graphs.cabal
+++ b/algebraic-graphs.cabal
@@ -155,7 +155,7 @@ test-suite test-alga
     if !impl(ghc >= 8.0)
         build-depends:  semigroups   >= 0.18.2  && < 0.18.4
     if impl(ghc >= 8.0.2)
-        build-depends:  inspection-testing >= 0.4 && < 0.4.2
+        build-depends:  inspection-testing >= 0.4.2.2 && < 0.5
 
     default-language:   Haskell2010
     GHC-options:        -Wall
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,1 @@
-resolver: lts-13.23
-
-packages:
-- '.'
+resolver: lts-14.7
--- a/test/Algebra/Graph/Test/RewriteRules.hs
+++ b/test/Algebra/Graph/Test/RewriteRules.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP             #-}
 {-# LANGUAGE TemplateHaskell #-}
 -----------------------------------------------------------------------------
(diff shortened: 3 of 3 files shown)


algebraic-graphs.cabal
```diff
@@ -155,7 +155,7 @@ test-suite test-alga
     if !impl(ghc >= 8.0)
         build-depends:  semigroups   >= 0.18.2  && < 0.18.4
     if impl(ghc >= 8.0.2)
-        build-depends:  inspection-testing >= 0.4 && < 0.4.2
+        build-depends:  inspection-testing >= 0.4.2.2 && < 0.5
```
fosskers: `0.4.2.2` has a `base` bound wide enough to allow projects using `algebraic-graphs` to build with GHC 8.8.

stack.yaml
```diff
@@ -1,4 +1 @@
-resolver: lts-13.23
-
-packages:
-- '.'
+resolver: lts-14.7
```
fosskers: Still a 8.6-based LTS, as we don't yet have one for 8.8.

test/Algebra/Graph/Test/RewriteRules.hs
```diff
@@ -111,16 +105,15 @@ inspect $ 'bind2 === 'bind2R
...
 
 inspect $ 'ovAp =/= 'ovApR
 
+-- Here we inline "by hand", but we see different results on GHC 8.6 vs GHC 8.8!
 ovAp', ovApR' :: Graph (a -> b) -> Graph (a -> b) -> Graph a -> Graph b
 ovAp'  x y z = overlay x y <*> z
 ovApR' x y z = overlay (x >>= (<$> z)) (y >>= (<$> z))
 
-inspect $ 'ovAp' === 'ovApR'
+-- inspect $ 'ovAp' =/= 'ovApR'
```
snowleopard: Perhaps, instead of commenting this out, you could use `CPP` to use `===` or `=/=` depending on the compiler version?

snowleopard:
@fosskers Thanks! Happy to merge this, but see one suggestion above.

test/Algebra/Graph/Test/RewriteRules.hs
fosskers: Reasonable! I'll take care of that.

fosskers:
There we go.

snowleopard:
@fosskers Thank you, merged!

PR #247, opened 2020-01-10
Make hasEdge a good consumer

snowleopard:


snowleopard:
OK, the first try is terrible: `hasEdge: 88.86 (bad)` :)

snowleopard:
Now `hasEdge: 111.85 (bad)`.

snowleopard:
OK, now down to just `hasEdge: 1.23 (bad)`.

Unclear how bad this actually is since we typically get `1.17 (bad)` due to a bug in the performance suite.

snowleopard:
@nobrakal Hi there, have a look at the new implementation of `hasEdge` which is based on `foldg`. I think it should fuse nicely with various good graph producers :)

I still can't make it quite as fast as the old one, but I think it should be possible.

snowleopard:
OK, I've finally reached `hasEdge: 1.10 (OK)`, so I think we can merge this 🎉

nobrakal:
Hi @snowleopard  this is very cool :D 

Just one question, why using integers instead of a thre-valued algebraic structure (like `data SmallNum = Zero | One | Two`) ?

PS: I am sure this is a good pretext to revive #228

snowleopard:
@nobrakal Yes, I was actually thinking about using a small data type like `SmallNum` or even repurposing the isomorphic `Ordering` :D 

However, `SmallNum`/`Ordering` and `Int` seem to have the same performance, and plain old integers feel like a simpler solution, which also generalises to higher dimensions. Would you prefer `SmallNum` instead?

> PS: I am sure this is a good pretext to revive #228

Indeed! 👍

nobrakal:
The only advantage of `SmallNum` is, with the right (lazy) `max` to have:

```Haskell
    o x y a = max (x a) (y a)
    c x y a = let xa = x a in max xa (y xa)
```

Which I think render much better the idea. Since this  is only cosmetic, I can live with plain old integers :)

snowleopard:
@nobrakal I decided to go with `Int` for now, but we might rethink this in future :-)

By the way, I expect this PR to bring a small performance improvement: moving to roughly `1.10` compared to `1.17` which we seem to typically get from the performance suite.

PR #250, opened 2020-01-22
NonEmpty AdjacencyIntMaps

jitwit:
Adding modules for NAIMs

I was having problems in the test suite with `reachable` about couldn't match type `ToVertex AIM` with type `Int` so I commented it out for now. 

I changed the complexity comments for `vertexCount`/`edgeCount` in AIM, which claimed `O(1)`.

Hopefully not too many names in the documentation are out of sync!


src/Algebra/Graph/NonEmpty/AdjacencyIntMap.hs
```diff
@@ -0,0 +1,716 @@
...
+        eshow [(x, y)] = showString "edge "      . showsPrec 11 x .
+                         showString " "          . showsPrec 11 y
+        eshow xs       = showString "edges1 "    . showsPrec 11 xs
+        used           = IntSet.toAscList $ IntSet.fromList $ uncurry (++) $ unzip es
+
+-- Unsafe creation of a NonEmpty list.
+unsafeNonEmpty :: [Int] -> NonEmpty Int
+unsafeNonEmpty = fromMaybe (error msg) . nonEmpty
+  where
+    msg = "Algebra.Graph.AdjacencyIntMap.unsafeNonEmpty: Graph is empty"
```
snowleopard: Should be `Algebra.Graph.NonEmpty.AdjacencyIntMap`. Alternatively, we could move this utility function to `Algebra.Graph.Internal`.

src/Algebra/Graph/NonEmpty/AdjacencyIntMap.hs
```diff
@@ -0,0 +1,716 @@
...
+-- 'toNonEmpty' . fromNonEmpty == 'Just'
+-- @
+fromNonEmpty :: AdjacencyIntMap -> AIM.AdjacencyIntMap
+fromNonEmpty = am
+
+-- | Construct the graph comprising /a single isolated vertex/.
+-- Complexity: /O(1)/ time and memory.
+--
+-- @
+-- 'AdjacencyIntMap.hasVertex' x (vertex x) == True
```
snowleopard: I'm a bit confused: do we really need qualifications `AdjacencyIntMap.` here?

jitwit: I don't think so. I was following very closely the `NonEmpty.AdjacencyMap` module

src/Algebra/Graph/NonEmpty/AdjacencyIntMap.hs
```diff
@@ -0,0 +1,716 @@
...
+-- that are 'Nothing'. Returns 'Nothing' if the resulting graph is empty.
+-- Complexity: /O(n + m)/ time.
+--
+-- @
+-- induceJust1 ('vertex' 'Nothing')                               == 'Nothing'
+-- induceJust1 ('edge' ('Just' x) 'Nothing')                        == 'Just' ('vertex' x)
+-- induceJust1 . 'gmap' 'Just'                                    == 'Just'
+-- induceJust1 . 'gmap' (\\x -> if p x then 'Just' x else 'Nothing') == 'induce1' p
+-- @
+-- TODO: double check there's no analogue for AIMs
```
snowleopard: Yep, I don't think there is one.

test/Algebra/Graph/Test/NonEmpty/AdjacencyIntMap.hs
```diff
@@ -0,0 +1,626 @@
...
+import Algebra.Graph.ToGraph (reachable)
+
+import qualified Algebra.Graph.AdjacencyIntMap          as AIM
+import qualified Algebra.Graph.NonEmpty.AdjacencyIntMap as NonEmpty
+import qualified Data.List.NonEmpty                     as NonEmpty
+import qualified Data.Set                               as Set
+import qualified Data.IntSet                            as IntSet
+
+sizeLimit :: Testable prop => prop -> Property
+sizeLimit = mapSize (min 10)
```
snowleopard: There is `size10` in `Algebra.Graph.Test`, so this can be dropped.

src/Algebra/Graph/NonEmpty/AdjacencyIntMap.hs
```diff
@@ -0,0 +1,716 @@
...
+-- | The set of vertices of a given graph.
+-- Complexity: /O(n)/ time and memory.
+--
+-- @
+-- vertexSet . 'vertex'    == Set.'Set.singleton'
+-- vertexSet . 'vertices1' == Set.'Set.fromList' . 'Data.List.NonEmpty.toList'
+-- vertexSet . 'clique1'   == Set.'Set.fromList' . 'Data.List.NonEmpty.toList'
+-- @
+vertexSet :: AdjacencyIntMap -> IntSet
+vertexSet = coerce AIM.vertexIntSet
```
jitwit: Oops, should be `vertexIntSet`

test/Algebra/Graph/Test/NonEmpty/AdjacencyIntMap.hs
```diff
@@ -0,0 +1,626 @@
...
+
+type G = NonEmpty.AdjacencyIntMap
+
+axioms :: G -> G -> G -> Property
+axioms x y z = conjoin
```
snowleopard: Let's factor out these axioms and theorems for non-empty graphs to `Algebra.Graph.Test` too, to avoid code duplication.

src/Algebra/Graph/NonEmpty/AdjacencyIntMap.hs
jitwit: (And accompanying documentation functions `IntSet.`)

test/Algebra/Graph/Test/NonEmpty/AdjacencyIntMap.hs
```diff
@@ -0,0 +1,626 @@
...
+    ,             x <= x + y                    // "Overlay order"
+    ,         x + y <= x * y                    // "Overlay-connect order" ]
+
+testNonEmptyAdjacencyIntMap :: IO ()
+testNonEmptyAdjacencyIntMap = do
+    putStrLn "\n============ NonEmpty.AdjacencyIntMap ============"
+    test "Axioms of non-empty graphs"   axioms
+    test "Theorems of non-empty graphs" theorems
+
+    putStrLn $ "\n============ Ord NonEmpty.AdjacencyIntMap ============"
```
snowleopard: It would be nice to reduce code duplication for non-empty tests. We could move some tests into `Algebra.Graph.Test.Generic`, adding suffix `NonEmpty` to properties for disambiguatation.

src/Algebra/Graph/NonEmpty/AdjacencyIntMap.hs
snowleopard: Ha, indeed. If it's redundant, let's drop it there too.

src/Algebra/Graph/NonEmpty/AdjacencyIntMap.hs
snowleopard: Indeed :+1:

snowleopard:
@jitwit Thanks for doing this! I'm a bit terrified about the combinatorial explosion in modules but no idea what to do about it :)

src/Algebra/Graph/NonEmpty/AdjacencyIntMap.hs
jitwit: I think Internal sounds good

test/Algebra/Graph/Test/NonEmpty/AdjacencyIntMap.hs
jitwit: Does that require `Class Graph` instances for the `NonEmpty` types?

jitwit: To fit with the `GraphTestsuite g` synonym

snowleopard: Hmm, I think you are right, it's not that easy. OK, let's keep this as is, but leave a TODO to fix it at some point.

jitwit:
Yeah, the explosion is a bit horrifying. Is that part of why no `AdjacencyHashMap` type modules are supported?

snowleopard:
@jitwit Yep. At some point, I also had a representation based on arrays but then deleted it too.

test/Algebra/Graph/Test/NonEmpty/AdjacencyIntMap.hs
jitwit: `testAdjacencyMap` would be a model to go off? It seems like most of the functions from the `NAM` modules are candidates

snowleopard: Yes, I think we can reuse tests for most functions from `NonEmpty` modules. Just thought that perhaps creating `Generic.NonEmpty` would be best since then we don't need to do any renaming.

jitwit: Sorry, I'm not clear about the some of the issues and steps. What would need to be renamed? Does the plan look something like this?
* `Algebra.Graph.Test.API` module should get `NonEmpty` functions added and the nonempty types should get a `nonemptyAPI :: API (G/Mono) (Ord/(~) Int)`
* also add the `ToGraph` instance for NAIM
* new `TestsuiteInt g -> IO ()` functions go in `Algebra.Graph.Test.Generic.NonEmpty`
* each nonempty module would import the `API`, `Generic`, and `Generic.NonEmpty` modules similar to what `Algebra.Graph.Test.AdjacencyMap` does

snowleopard: After refreshing my memory on how things are implemented, here is a more concrete plan: 

* Extend `data API g c` with new fields that correspond to missing methods (most of them end with the suffix `1`). In fact there is a TODO for this: https://github.com/snowleopard/alga/blob/026ba0e14f267e0f3cd19feec1e64b2d5e4ecd35/test/Algebra/Graph/Test/API.hs#L58

* Add tests for missing methods to `Test.Generic` or `Test.Generic.NonEmpty` -- whichever turns out to be clearer/more convenient.

* I don't think we need any `ToGraph` instances for non-empty graphs but perhaps I'm mistaken? Instead, I expect `API` definitions similar to https://github.com/snowleopard/alga/blob/026ba0e14f267e0f3cd19feec1e64b2d5e4ecd35/test/Algebra/Graph/Test/API.hs#L146-L148

* Tests for non-empty modules would need to import `API`, `Generic` and perhaps also `Generic.NonEmpty` (if it turns out to be convenient).

jitwit: Thanks for the plan! Refactoring the nonempty modules to use generic is a bit daunting, but I'll try to find time and energy soon

snowleopard: @jitwit If you don't want to deal with this daunting task feel free to simply leave a TODO describing the plan. I do understand that it's no fun at all and happy to do it myself :)

jitwit:
Hi and sorry for the MASSIVE delay, I finally returned to this! Hope all is well.
I resolved some merge conflicts and added the last comments you made about the plan for generic tests to `test/Algebra/Graph/Test/API.hs`

snowleopard:
@jitwit Hey, welcome back! All is good and I hope you're good too. Nice new avatar 👍 

The plan looks fine to me, looking forward to future commits :)

jitwit:
Things are still good here, too. Thanks!

zyklotomic:
@snowleopard 

I was trying to continue the work mentioned in the thread above https://github.com/snowleopard/alga/pull/250#discussion_r371318553 .

If I understand correctly about extending the test API, we would want to add the field
```haskell
overlays1 :: forall a. c a => NonEmpty (g a) -> g a
```
 which should be analogous to the existing

```haskell
overlays :: forall a. c a => [g a] -> g a
```

Furthermore, I interpreted your plan as that we should be adding API's for other types of graphs. I started with `nonEmptyAdjacencyMapAPI :: API NAM.AdjacencyMap Ord` where `NAM` is the NonEmpty Adjacency Map module,  and
found an issue with implementing the API field for `overlays`. I can't think of a way that makes sense, since the type `[g a]` doesn't guarantee that the list will be non-empty.

I think we do in fact need `ToGraph` instances, because some API entries would be more conveniently implemented
using `ToGraph`, for example `isTopSortOf = ToGraph.isTopSortOf`.

I am not sure if I understood your plan correctly, let me know if I have misunderstood it. Thanks!

snowleopard:
@zyklotomic I think you understood the plan correctly. However, implementing a non-generic version of the testsuite for the new module should be OK too: as long as every property we promise in comments is tested, I will be happy :) So, feel free to test some (or all?) of the properties directly, without trying to be too generic. (I admit I might have overengineered the testsuite quite a bit...) 

However, since this is all still going to be quite a lot of work, I'd like us to take a step back and think: is it really worth it in the end? This library is already hard to navigate because of so many similar APIs: empty/non-empty, `Int`/generic, algebraic/`Map`-based etc. 

If I recall correctly, your specific motivation for doing this was the `scc` function, so here are two questions:
* Could you perhaps use `AdjacencyMap` instead of `AdjacencyIntMap`? Is the performance boost you get from `IntMap`s really that important for your application?
* If the answer to the above is "yes": perhaps, it would be better to invest our efforts into making `AdjacencyMap`s map faster, perhaps, by implementing it with `IntMap`s with some auxiliary `Int`/`a` mapping. 

I am a bit unhappy with how the library looks at the moment (it's way too complex) and I think adding more APIs may be a step in the wrong direction. So, it'd be really interesting to learn more about your motivation/use-case.

zyklotomic:
I have already been using an auxiliary mapping due to space complexity reasons. Performance boost isn't super important.

If you are interested, I used scc [here](https://gitlab.haskell.org/ghc/ghc-debug/-/merge_requests/5/diffs#diff-content-2e850d9bbcd25f97b3ddb5db44d1026bfefa416a). I was trying to write a heap analysis of the GHC runtime. I used scc to compress cycles in the heap reference graph into one representative node.

The only reason why I thought to add it was because it was missing. I do agree that the library is a bit complex. In my experience, it is because of the inconsistencies (such as `scc` not being available for IntMaps).  Having to select a specific graph "backend" such as `AdjacencyMap` was confusing. It is still a really enjoyable library to use nonetheless.

snowleopard:
> If you are interested, I used scc here. I was trying to write a heap analysis of the GHC runtime. I used scc to compress cycles in the heap reference graph into one representative node.

Wow, that's very cool! Best wishes with finishing this project.

> I do agree that the library is a bit complex. In my experience, it is because of the inconsistencies (such as `scc` not being available for IntMaps). Having to select a specific graph "backend" such as `AdjacencyMap` was confusing. 

Yes, I agree with you. I'd love to have consistency, but it seems almost unreachable in practice because of so many different combinations of graph flavours. It's very hard to implement, document, test and maintain all of these combinations. I'll try to find a way around that soon.

PR #254, opened 2020-01-31
More goodness

### LINKED ISSUE #252: How to handle goodness inherited from other libraries
Hi,

I just discovered that [`Data.Set.toAscList`](http://hackage.haskell.org/package/containers-0.6.2.1/docs/Data-Set.html#v:toAscList) is subject to list fusion. This makes, for example, `Data.Graph.vertexList` automatically a good producer of lists.

The right path to follow seems not clear, should we:

1. Only document this
2. Document and test, that is, testing some functions from another library
3. Keep it for us (I don't think this is a good choice, but I added it to be exhaustive)

?

(I think there is a more conceptual issue on how goodness is actually handled in general, but this is out of scope)

nobrakal:
Hi,

This PR fixes #252 
It document and test that `vertexList` and `edgeList` are good producer of lists.


snowleopard:
@nobrakal Looks great, thank you! Merged :)

PR #255, opened 2020-02-01
Revision in preparation for 0.5

snowleopard:
Doing a revision pass while preparing for a release.


snowleopard:
@bolt12 @vasalf I've made quite a lot of revisions to undirected and bipartite graphs. Hope I didn't mess up anything important. If you spot anything, please let me know!

snowleopard:
@vasalf I've just realised that I've probably created a lot of rebasing work for you -- sorry about that!

bolt12:
@snowleopard I have looked through the undirected graphs part and it seemed very nice to me! Was there any specific reason to use Symmetric Relation instead of Adjacency Graph? :)

snowleopard:
@bolt12 Thanks! 

> Was there any specific reason to use Symmetric Relation instead of Adjacency Graph?

I'm not entirely sure what you mean. The implementation was already using `Symmetric.Relation`. I just removed some rewrite rules that were supposed to be useful only in future when we switch to `Undirected.AdjacencyMap` as underlying representation (which we don't have yet). I think that at some point we'll just remove all our `Relation` data structures and replace them with corresponding adjacency maps to avoid keeping too many representations in the library :)

bolt12:
Oops right! I don't know why but I had the idea that `AdjacencyMap` was already used, sorry about that!

Overall, from the `Undirected` point of view, I think everything is set! :tada:

PR #263, opened 2020-05-18
Import liftM2 from Control.Monad instead of Control.Monad.State

Bodigrim:
There is an intention to stop re-exporting functions from `base` in future versions of `mtl`: https://github.com/haskell/mtl/pull/74
This PR brings `alga` in line with the proposed change, because at the moment `Algebra.Graph.NonEmpty` imports `liftM2` (originally from `Control.Monad`) from `Control.Monad.State`.


snowleopard:
@Bodigrim Oh wow, thanks a lot! That's the first PR I receive from the future :-)

PR #266, opened 2020-11-16
Fix compatibility with QuickCheck 2.14.2

### LINKED ISSUE #265: Tests failed to build with QuickCheck 2.14.2
```
[ 1 of 22] Compiling Algebra.Graph.Test.Arbitrary ( test/Algebra/Graph/Test/Arbitrary.hs, dist/build/test-alga/test-alga-tmp/Algebra/Graph/Test/Arbitrary.dyn_o )

test/Algebra/Graph/Test/Arbitrary.hs:220:10: error:
    Duplicate instance declarations:
      instance Arbitrary a => Arbitrary (Tree a)
        -- Defined at test/Algebra/Graph/Test/Arbitrary.hs:220:10
      instance Arbitrary a => Arbitrary (Tree a)
        -- Defined in ‘Test.QuickCheck.Arbitrary’
    |
220 | instance Arbitrary a => Arbitrary (Tree a) where
    |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

felixonmars:
Fixes #265

### DIFF
--- a/test/Algebra/Graph/Test/Arbitrary.hs
+++ b/test/Algebra/Graph/Test/Arbitrary.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 -----------------------------------------------------------------------------
 -- |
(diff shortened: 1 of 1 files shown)


snowleopard:
@felixonmars Thanks for the issue and the fix!

It's very likely that the custom `Arbitrary Tree` instance defined here and the instance provided by `quickcheck-2.14.2` are different. Did you investigate the differences? If they are indeed substantially different, then swapping one for the other depending on which QuickCheck version happens to be found by the constraint solver, doesn't seem right. If they are more or less the same, then it's probably OK.

felixonmars:
https://github.com/nick8325/quickcheck/blob/3c098d757a04006d1b1f127c5a1789f270d70547/src/Test/QuickCheck/Arbitrary.hs#L825

They do look different, but the tests are passing here. I am not an expert here, though :(

snowleopard:
@felixonmars I decided not to overcomplicate things and merged the PR. Thank you!

(P.S.: A better approach could have been to stick to the existing instance, but wrap `Tree` into a `newtype` and unwrap it in the corresponding tests. But that seems like way too much work for little benefit.)

PR #271, opened 2021-06-20
Implement induce via induceJust

snowleopard:
Implementing `induce` as `induce p = induceJust . fmap (\a -> if p a then Just a else Nothing)` seems to pass inspection tests. My understanding is that this means `fmap` fuses nicely with the `induceJust`, and so using this simpler definition comes at no runtime cost.

@nobrakal Could you have a look?


nobrakal:
I just hand-checked it to be sure: it works :tada: You understood everything correctly, fusing happens as expected.
GHC is impressive!

snowleopard:
@nobrakal Great, many thanks for double-checking! That's a very nice example of graph fusion in action :)

PR #273, opened 2021-08-13
Export.Dot: Extend Style to define Quoting styles

### LINKED ISSUE #272: support for GraphViz "HTML strings" for node or edge labels
Hello!

GraphViz supports something called [HTML-like labels](https://graphviz.org/doc/info/shapes.html#html) for some objects. This can be useful for adding some mark-up to edge or node labels (e.g. fractions with `<<SUP>1</SUP>/<SUB>2</SUB>2>`). A HTML-like label is delimited with a pair of angle brackets instead of double-quotes. E.g.

```
digraph {
  1 [label="regular label"];
  2 [label=<<B>emboldened label</B>>];
  1 -> 2;
}
```

resulting in:

tmp

[The current handling of attributes in Algebra.Graph.Export.Dot](https://github.com/snowleopard/alga/blob/master/src/Algebra/Graph/Export/Dot.hs#L127) always wraps labels in double-quotes. It would be nice to be able to specify that the HTML-like format was to be used. 

(I completely forgot about #44/#47 in the intervening years. I'll try to pick it back up)

jmtd:
GraphViz supports different types of label. The type is determined by
the delimiters used to quote the label's value. Double-quotes denote
a literal string. Pairs of angle brackets ("<>") denote a "HTML-like"
string, which permits the use of (some) HTML mark-up in the graph.

Extend the `Style` record type to include a `quoting` field, to signal
whether or not attribute values should be double-quoted. A user wishing
to use another quote style must define their style's quoting attribute
as `NoQuote` and then ensure their attribute values contain embedded
quote characters of their choice.

Adjust `defaultStyle` and the `attributes` function accordingly and add
a test.

Fixes #272.

----

[Here's an example of the above in use](https://github.com/jmtd/striot/commit/38ee592c7f4cea4b695c6db5a94fe8522ab9ed5e). Note that I have to take responsibility for wrapping all values, and I vary what delimiter I use. Here's what the result looks like:

```
λ> putStrLn $ jacksonGraphToDot mergeEx
digraph 
{
  graph [bgcolor="white"]
  node [shape="box" fillcolor="white" style="filled"]
  edge [weight="10" color="black" fontcolor="black"]
  "1" [label="streamSource (do {let {x'_0 = \"foo\"};
    threadDelay (1000 * 1000);
    putStrLn \"sending '\" ++ (x'_0 ++ \"'\");
    return x'_0})" xlabel=<<SUP>0.0</SUP>/<SUB>s</SUB>> fillcolor="#ffffff"]
  "2" [label="streamMap (id)" xlabel=<<SUP>1.0</SUP>/<SUB>s</SUB>> fillcolor="#ffffff"]
  "3" [label="streamSource (do {let {x'_0 = \"bar\"};
    threadDelay (1000 * 1000);
    putStrLn \"sending '\" ++ (x'_0 ++ \"'\");
    return x'_0})" xlabel=<<SUP>2.0</SUP>/<SUB>s</SUB>> fillcolor="#ffcccc"]
  "4" [label="streamMap (id)" xlabel=<<SUP>3.0</SUP>/<SUB>s</SUB>> fillcolor="#ffcccc"]
  "5" [label="streamMerge" xlabel=<<SUP>4.0</SUP>/<SUB>s</SUB>> fillcolor="#ffcccc"]
  "6" [label="streamSink (mapM_ print)" xlabel=<<SUP>5.0</SUP>/<SUB>s</SUB>> fillcolor="#ffcccc"]
  "1" -> "2" [label=<<SUP>1.0</SUP>/<SUB>s</SUB> <I>:: String</I>>]
  "2" -> "5" [label=<<SUP>2.0</SUP>/<SUB>s</SUB> <I>:: [String]</I>>]
  "3" -> "4" [label=<<SUP>1.0</SUP>/<SUB>s</SUB> <I>:: String</I>>]
  "4" -> "5" [label=<<SUP>2.0</SUP>/<SUB>s</SUB> <I>:: [String]</I>>]
  "5" -> "6" [label=<<SUP>2.0</SUP>/<SUB>s</SUB> <I>:: String</I>>]
}
```
alga

### DIFF
--- a/src/Algebra/Graph/Export/Dot.hs
+++ b/src/Algebra/Graph/Export/Dot.hs
@@ -15,7 +15,7 @@
 -----------------------------------------------------------------------------
 module Algebra.Graph.Export.Dot (
     -- * Graph attributes and style
-    Attribute (..), Style (..), defaultStyle, defaultStyleViaShow,
+    Attribute (..), Quoting (..), Style (..), defaultStyle, defaultStyleViaShow,
 
     -- * Export functions
     export, exportAsIs, exportViaShow
@@ -33,6 +33,9 @@ import qualified Algebra.Graph.Export as E
 -- | An attribute is just a key-value pair, for example @"shape" := "box"@.
 -- Attributes are used to specify the style of graph elements during export.
 data Attribute s = (:=) s s
+
+-- | Used to determine whether to use double-quotes around attribute values.
+data Quoting = DoubleQuotes | NoQuotes
 
 -- | The record 'Style' @a@ @s@ specifies the style to use when exporting a
 -- graph in the DOT format. Here @a@ is the type of the graph vertices, and @s@
@@ -57,12 +60,14 @@ data Style a s = Style
     -- ^ Attributes of a specific vertex.
     , edgeAttributes   :: a -> a -> [Attribute s]
     -- ^ Attributes of a specific edge.
+    , attributeQuoting :: Quoting
+    -- ^ What style of quoting to apply to attributes.
     }
 
 -- | Default style for exporting graphs. All style settings are empty except for
 -- 'vertexName', which is provided as the only argument.
 defaultStyle :: Monoid s => (a -> s) -> Style a s
-defaultStyle v = Style mempty [] [] [] [] v (const []) (\_ _ -> [])
+defaultStyle v = Style mempty [] [] [] [] v (const []) (\_ _ -> []) DoubleQuotes
 
 -- | Default style for exporting graphs whose vertices are 'Show'-able. All
 -- style settings are empty except for 'vertexName', which is computed from
@@ -88,7 +93,8 @@ defaultStyleViaShow = defaultStyle (fromString . show)
 --     , 'defaultEdgeAttributes'   = 'mempty'
 --     , 'vertexName'              = \\x   -> "v" ++ 'show' x
 --     , 'vertexAttributes'        = \\x   -> ["color" := "blue"   | 'odd' x      ]
---     , 'edgeAttributes'          = \\x y -> ["style" := "dashed" | 'odd' (x * y)] }
+--     , 'edgeAttributes'          = \\x y -> ["style" := "dashed" | 'odd' (x * y)]
+--     , 'attributeQuoting'        = 'DoubleQuotes' }
 --
 -- > putStrLn $ export style (1 * 2 + 3 * 4 * 5 :: 'Graph' Int)
 --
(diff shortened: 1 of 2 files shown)


src/Algebra/Graph/Export/Dot.hs
```diff
@@ -57,12 +60,14 @@ data Style a s = Style
     -- ^ Attributes of a specific vertex.
     , edgeAttributes   :: a -> a -> [Attribute s]
     -- ^ Attributes of a specific edge.
+    , quoting :: Quoting
```
snowleopard, line 63: It may be more future-proof to name this field `attributeQuoting`, since we might also add `labelQuoting :: Quoting` at some point (right now we always use double-quoting for labels).

src/Algebra/Graph/Export/Dot.hs
```diff
@@ -113,25 +118,28 @@ export :: (IsString s, Monoid s, Ord a, ToGraph g, ToVertex g ~ a) => Style a s
...
              <> ("edge"  `with` defaultEdgeAttributes)
              <> E.export vDoc eDoc g
     label     = doubleQuotes . literal . vertexName
-    vDoc x    = line $ label x <+>                      attributes (vertexAttributes x)
-    eDoc x y  = line $ label x <> " -> " <> label y <+> attributes (edgeAttributes x y)
+    vDoc x    = line $ label x <+>                      attributes quoting (vertexAttributes x)
+    eDoc x y  = line $ label x <> " -> " <> label y <+> attributes quoting (edgeAttributes x y)
 
 -- | A list of attributes formatted as a DOT document.
 -- Example: @attributes ["label" := "A label", "shape" := "box"]@
```
snowleopard, line 133: The example here should be adjusted. There may be other cases where the comments got out of date.

snowleopard:
@jmtd Many thanks, the PR looks good! I left a couple of comments.

src/Algebra/Graph/Export/Dot.hs
```diff
@@ -34,6 +34,9 @@ import qualified Algebra.Graph.Export as E
 -- Attributes are used to specify the style of graph elements during export.
 data Attribute s = (:=) s s
 
+-- | Used to determine whether to use double-quotes around attribute values.
+data Quoting = DoubleQuote | NoQuote
```
snowleopard, line 38: I weakly prefer plural `DoubleQuotes` and `NoQuotes` but I'll leave it to you to decide what's best.

jmtd:
Thanks for the quick review! I've addressed your comments and force-pushed an update.

snowleopard:
Merged, thank you!

PR #274, opened 2021-08-18
Update of bipartite graph algorithms

luc-tielen:
This PR continues where #229 left off.
I merged in master, `stack build` and `stack test` seem to work.
You might want to check cabal file, since I had some merge conflicts there (I ended up choosing cabal file of #229 and added extensions manually to each file that required them).

EDIT: CI seems to fail, might be related to cabal file changes?


snowleopard:
@luc-tielen Many thanks, I'll have a look soon!

Regarding the `.cabal` file: I suggest taking the one from the master branch and adapting it (simply adding new files?) until the project builds. Right now, I can see a lot of changes that are unrelated to this PR, such as version and copyright year changes, etc.

snowleopard:
Now CI seems to fail because of HLint. I looked at the suggestions and they look fine, with the exception of `Use if`. Perhaps, we should just disable this rule (at least in the module in question).

snowleopard:
Thanks for fixing HLint related errors! There are a few other warnings left though.

luc-tielen:
@snowleopard Yeah, I'm fixing these during my lunchbreaks :-) the remaining errors are related to the actual code, but should be easy to fix.

luc-tielen:
@snowleopard Idk what the current error is.. it's complaining it can't find an object file when linking? Maybe I still have an error in cabal file? I have no windows machine to test it with either, only Linux and OSX. Tests on Linux seem to be passing though! 😃

luc-tielen:
@snowleopard so I think I might have an idea what's going on.. what if the filename is simply too long on Windows? I vaguely remember that Windows has a max limit on filenames.. Can we try shortening the module name somehow? Right now the path `C:\Users\runneradmin\AppData\Local\Temp\extra-dir-798123346708\algebraic-graphs-0.6.1\dist-newstyle\build\x86_64-windows\ghc-9.0.1\algebraic-graphs-0.6.1\t\test-alga\build\test-alga\test-alga-tmp\Algebra\Graph\Test\Bipartite\Undirected\AdjacencyMap\Algorithm.o` is 260 chars long.. which is apparently around the limit that Windows supports..

If this is indeed the case, this will make me facepalm so hard 🤦

snowleopard:
@luc-tielen Ohh, I think you are right 🤦‍♂️ 

The problem is with the tests, so I guess we can just collapse one layer of hierarchy, e.g. drop the `Algorithm` component. What do you think? Or even compress `AdjacencyMap.Algorithm` to just `AM`, which is a common contraction in the code base.

luc-tielen:
@snowleopard so that was it, CI is finally green 😅

snowleopard:
@luc-tielen Thanks, that's great! I'd like to have a final look before merging but I don't expect any issues now. I'll try to do it in the next couple of days if that's OK.

luc-tielen:
Great! Thanks so much!

snowleopard (approved):
Looks great but I'll probably make a minor revision pass after merging. Many thanks @luc-tielen and @vasalf!

luc-tielen:
Thanks @snowleopard ! Could you let me know when you finish that implementation / make a new release so I can start using it in my own code?

snowleopard:
@luc-tielen Sure, I'll ping you! By the way, I believe you don't really need a release to start using the new version: you could just point Stack or Cabal to a git commit, right?

luc-tielen:
Right, I will do that in the meantime! Thanks again.

PR #276, opened 2021-09-14
Add `Monoid` and `Semigroup` instances

### LINKED ISSUE #269: Monoid instance
It seems to me that `Graph` is missing a Monoid instance:

```haskell
mempty = empty
(<>) = overlay
```

There's also a monoid over `connect`, but that doesn't seem to be as relevant in terms of how people think about constructing graphs --- which is adding the vertices and edges to a bucket.

ocharles:
Fixes #269.


src/Algebra/Graph.hs
```diff
@@ -294,6 +294,12 @@ instance MonadPlus Graph where
     mzero = Empty
     mplus = Overlay
 
+instance Semigroup (Graph a) where
```
snowleopard, line 297: Let's add a clarification (here and in other instances too):

```suggestion
-- | Defined via 'overlay'.
instance Semigroup (Graph a) where
```

snowleopard:
@ocharles Many thanks for the PR! Alas, CI fails on a couple of GHC versions, most likely we need to do a dance with imports and/or CPP.

ocharles:
Do you really care about 8.0 and 8.2? You might be inventing more work than you need to, that's quite the range of support! But that's for another issue, I'll get this all building with CI tomorrow (along with those instance comments) :+1:

snowleopard:
Yeah, I think you are right and I should just stop supporting 8.0 and 8.2.

Shall we keep this PR as is for now and switch to dropping 8.0 and 8.2 to avoid unnecessary work? Either way is fine for me.

ocharles:
@snowleopard If you're planning to drop support for 8 & 8.2 then I won't make any changes - I'm always one for doing the least amount of work possible! :laughing: I will add the Haddock comments though.

snowleopard:
@ocharles Done!

ocharles:
@snowleopard ok, documentation added

src/Algebra/Graph/Labelled.hs
```diff
@@ -92,6 +92,14 @@ instance (NFData e, NFData a) => NFData (Graph e a) where
     rnf (Vertex  x    ) = rnf x
     rnf (Connect e x y) = e `seq` rnf x `seq` rnf y
 
+-- | Defined via 'overlay'.
+instance Monoid e => Semigroup (Graph e a) where
+    (<>) = overlay
+
+-- | Defined via 'empty'.
```
snowleopard, line 99:
```suggestion
-- | Defined via 'overlay' and 'empty'.
```

snowleopard:
@ocharles Thanks, merged! Just tweaked one of the comments.

PR #277, opened 2021-09-17
Revise bipartite graphs

snowleopard:
This PR cleans up the API, docs and tests of bipartite graphs. In particular, I decided to drop the `Undirected` path component from module names since the directed/undirected distinction is very rarely used when working with bipartite graphs.

/cc @luc-tielen This may be a breaking change for you, sorry! Any feedback is welcome.


snowleopard:
@luc-tielen I've merge this PR but if you (as the only user of bipartite graphs I know!) is unhappy about any of the changes, please let me know.

luc-tielen:
Thanks for pinging me and merging this in. I've been focusing on the LLVM codegen part of my project, so didn't do any changes yet to the graph algorithm part.. but I only make use of `maxMatching` and `Matching` so I should be fine (I think).

(and otherwise I will create a new issue once I integrate everything together :wink:)

snowleopard:
@luc-tielen I see, sounds good :)

PR #280, opened 2021-10-31
NonEmpty AdjancencyIntMaps + SCC Algo

zyklotomic:
Continuing off of #250, and with the additional goal of adding support for SCC decomposition for AdjacencyIntMaps


zyklotomic:
The commit I just pushed was for the SCC algorithm that I hoped to have for the IntMap version of graphs. I have yet to test it.

If I understand correctly, what still needed to be done from the original PR was the test suite right? I will also need to add tests for IntMap SCC. I will look into that next! :D

snowleopard:
@zyklotomic Hi there, thanks for taking this over! My preference would be to finish the `NonEmpty` PR first, and then implement SCC as a separate PR. It would just make it easier for review. Does that work for you?

zyklotomic:
Yes I can do that, my apologies! I don't think I can directly work off of the old PR since I don't have push access to jitwit's branch. Should I create a new PR for `NonEmpty`, and rename this to being the SCC PR?

snowleopard:
I think it's fine to keep working on `NonEmpty` in this PR If you could just undo the SCC-related changes for now that would be great. Of course, please take care not to lose the SCC changes since we are going to need them eventually!

PR #291, opened 2022-06-14
Add support for GHC 9.4

snowleopard:
Waiting until 9.4 becomes available.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        ghc: ['9.2', '9.0', '8.10', '8.8', '8.6', '8.4']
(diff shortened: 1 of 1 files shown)


ocramz:
9.4.1 is out (https://www.haskell.org/ghc/blog/20220807-ghc-9.4.1-released.html), these CI jobs can be restarted

snowleopard:
We might need to wait a bit more. After restarting, I'm getting:
```
Error: All install methods for ghc 9.4 failed
```

snowleopard:
Retried today with the same results.

ysangkok:
The results are no longer available, would be nice to post them here for posterity.

snowleopard:
> The results are no longer available, would be nice to post them here for posterity.

There was nothing interesting, just some problems with the CI.

I plan to add support for GHC 9.4 (and maybe GHC 9.6) in the next couple of weeks. Stay tuned!

Issue #297, opened 2022-08-14
Smart constructor for Symbol rather than requiring -XOverloadedLists


ocramz:
It took me a minute to realize `coffeeTeaAutomaton` only works before OverloadedLists. Requiring the extension runs the risk of breaking other code in a module (just happened to me), which is not ideal.

Feature request : Can we have smart constructors for Symbol ? e.g.

```Haskell
symbol :: a -> Label a
symbol = Symbol 
```

and

```Haskell
symbols :: Foldable t => t a -> Label a
symbols = foldr ((<>) . Symbol) Zero
```

snowleopard:
Thanks for the PR! I'm hesitant to change the example in `coffeeTeaAutomaton` though since it would make it more cluttered. How about extending the corresponding comment, pointing out that the inner lists are overloaded?

PR #298, opened 2022-08-14
add symbol, symbols

ocramz:
Add smart constructors for Symbol , with the long-term goal of deprecating the IsList instance.  

I should note in passing that the GHC  DEPRECATE pragma doesn't work with class instances, which leaves the fix up to a judicious versioning schedule

Addresses (partly) #297

### DIFF
--- a/src/Algebra/Graph/Label.hs
+++ b/src/Algebra/Graph/Label.hs
@@ -364,8 +364,8 @@ instance (Monoid a, Ord a) => Semiring (PowerSet a) where
 
 instance (Monoid a, Ord a) => Dioid (PowerSet a) where
 
--- | The type of /free labels/ over the underlying set of symbols @a@. This data
--- type is an instance of classes 'StarSemiring' and 'Dioid'.
+-- | The type of /free labels/ over the underlying set of symbols @a@. 'Label' values
+-- can be manipulated via its 'Semigroup', 'Monoid' and 'StarSemiring' class instances.
 data Label a = Zero
              | One
              | Symbol a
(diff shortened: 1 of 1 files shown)


src/Algebra/Graph/Label.hs
```diff
@@ -377,6 +377,14 @@ data Label a = Zero
...
+symbol :: a -> Label a
+symbol = Symbol
+
+-- | Wrap a list of values into 'Symbol' constructors terminated by 'Zero'
+symbols :: Foldable t => t a -> Label a
+symbols = foldr ((<>) . Symbol) Zero
+
 instance IsList (Label a) where
     type Item (Label a) = a
     fromList = foldr ((<>) . Symbol) Zero
```
snowleopard, line 390:
```suggestion
    fromList = symbols
```

snowleopard:
Thanks! The change looks good to me. I left a small suggestion.

I think we can drop the `IsList` instance altogether if it's problematic.

> I should note in passing that the GHC DEPRECATE pragma doesn't work with class instances, which leaves the fix up to a judicious versioning schedule

I didn't quite get what you meant here. Could you elaborate?

ocramz:
> > judicious versioning schedule
> 
> I didn't quite get what you meant here. Could you elaborate?

Just that removing class instances would need bumping a major version number, and possibly warning users beforehand.

ocramz:
Also, I removed mention of the `Dioid (Label a)` instance because I couldn't find it.

snowleopard (approved):

snowleopard:
> Just that removing class instances would need bumping a major version number, and possibly warning users beforehand.

Ah, got it now. Agreed.

snowleopard:
Thanks, merged!

PR #305, opened 2023-03-22
Support GHC 9.2.7, 9.4.4 and 9.6.1

snowleopard:


Bodigrim:
As a Hackage trustee I also revised two latest releases to be GHC-9.6-compatible:
https://hackage.haskell.org/package/algebraic-graphs-0.6.1/revisions/
https://hackage.haskell.org/package/algebraic-graphs-0.7.0/revisions/

snowleopard:
Thank you!

Issue #307, opened 2023-04-06
Semigroup version of Algebra.Graph.Labelled.AdjacencyMap


dwincort:
I was looking into using this library (and the `Algebra.Graph.Labelled.AdjacencyMap` module in particular), but the edge type I want to use is not a `Monoid`, only a `Semigroup`.  Indeed, one of the things I like about `Labelled.AdjacencyMap` is that the adjacencies are stored sparsely, i.e., that non-existent edges are not stored, which makes a lot of sense for me because there is no value in my edge type that represents not-an-edge.

If I'm understanding the code correctly, it looks like `mempty` (or `zero`) of the edge monoid is only used to filter `zero` edges out of the adjacency map.  I'm not proposing you drop the `Monoid` constraint -- for edge types that _do_ have a `zero` element, filtering it out makes sense -- but it seems to me there could be a parallel module for labeled adjacency maps that only uses a `Semigroup` constraint and never filters.

P.S. Also, I understand that I could wrap my edge type in `Maybe` and, because the labeled `AdjacencyMap` never stores `zero` edges, unwrap with something like `fromJust` every time I view an edge.  I find this unsatisfying.

snowleopard:
How are you planning to use `Algebra.Graph.Labelled.AdjacencyMap` if your edge labels have no `zero`? You won't be able to use `overlay`, so you'll need to stick to fully-connected graphs, I guess, which seems odd.

dwincort:
I think if you look at the code you'll find that you can use `overlay` just fine.

```
overlay :: (Eq e, Monoid e, Ord a) => AdjacencyMap e a -> AdjacencyMap e a -> AdjacencyMap e a
overlay (AM x) (AM y) = AM $ Map.unionWith nonZeroUnion x y

-- Union maps, removing zero elements from the result.
nonZeroUnion :: (Eq e, Monoid e, Ord a) => Map a e -> Map a e -> Map a e
nonZeroUnion x y = Map.filter (/= zero) $ Map.unionWith mappend x y
```
If you make `e` a `Semigroup` instead of a `Monoid`, then you don't even need to worry about such a `nonZeroUnion` and can simply use `Map.unionWith (<>)` instead.

[...]

PR #308, opened 2023-04-23
Use Semigroup constraint for LAM instead of Monoid

dwincort:
As discussed in https://github.com/snowleopard/alga/issues/307, this eases the `Monoid` restriction labeled adjacency maps to `Semigroup`.

### DIFF
--- a/src/Algebra/Graph/Class.hs
+++ b/src/Algebra/Graph/Class.hs
@@ -155,7 +155,7 @@ instance Dioid e => Graph (LG.Graph e a) where
     overlay = LG.overlay
     connect = LG.connect one
 
-instance (Dioid e, Eq e, Ord a) => Graph (LAM.AdjacencyMap e a) where
+instance (Dioid e, Ord a) => Graph (LAM.AdjacencyMap e a) where
     type Vertex (LAM.AdjacencyMap e a) = a
     empty   = LAM.empty
     vertex  = LAM.vertex
--- a/src/Algebra/Graph/Labelled.hs
+++ b/src/Algebra/Graph/Labelled.hs
@@ -118,7 +118,7 @@ instance (Eq e, Monoid e, Ord a) => T.ToGraph (Graph e a) where
 -- subgraphs.
 -- Extract the adjacency map of a graph.
 toAdjacencyMap :: (Eq e, Monoid e, Ord a) => Graph e a -> AM.AdjacencyMap e a
-toAdjacencyMap = foldg AM.empty AM.vertex AM.connect
+toAdjacencyMap = foldg AM.empty AM.vertex (\e x y -> AM.trimZeroes $ AM.connect e x y)
 
 -- Convert the adjacency map to a graph.
 fromAdjacencyMap :: Monoid e => AM.AdjacencyMap e a -> Graph e a
--- a/src/Algebra/Graph/Labelled/AdjacencyMap.hs
+++ b/src/Algebra/Graph/Labelled/AdjacencyMap.hs
@@ -32,7 +32,7 @@ module Algebra.Graph.Labelled.AdjacencyMap (
 
     -- * Graph transformation
     removeVertex, removeEdge, replaceVertex, replaceEdge, transpose, gmap,
-    emap, induce, induceJust,
+    emap, induce, induceJust, trimZeroes,
 
     -- * Relational operations
     closure, reflexiveClosure, symmetricClosure, transitiveClosure,
@@ -59,10 +59,8 @@ import qualified Data.Map.Strict as Map
 import qualified Data.Set        as Set
 
 -- | Edge-labelled graphs, where the type variable @e@ stands for edge labels.
--- For example, 'AdjacencyMap' @Bool@ @a@ is isomorphic to unlabelled graphs
--- defined in the top-level module "Algebra.Graph.AdjacencyMap", where @False@
--- and @True@ denote the lack of and the existence of an unlabelled edge,
--- respectively.
+-- For example, 'AdjacencyMap' @()@ @a@ is isomorphic to unlabelled graphs
+-- defined in the top-level module "Algebra.Graph.AdjacencyMap".
 newtype AdjacencyMap e a = AM {
     -- | The /adjacency map/ of an edge-labelled graph: each vertex is
     -- associated with a map from its direct successors to the corresponding
@@ -90,7 +88,7 @@ instance (Ord a, Show a, Ord e, Show e) => Show (AdjacencyMap e a) where
                            showString " "      . showsPrec 11 y
             xs          -> showString "edges " . showsPrec 11 xs
 
-instance (Ord e, Monoid e, Ord a) => Ord (AdjacencyMap e a) where
+instance (Ord e, Semigroup e, Ord a) => Ord (AdjacencyMap e a) where
     compare x y = mconcat
         [ compare (vertexCount x) (vertexCount y)
         , compare (vertexSet   x) (vertexSet   y)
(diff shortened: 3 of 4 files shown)


src/Algebra/Graph/Class.hs
```diff
@@ -155,7 +155,7 @@ instance Dioid e => Graph (LG.Graph e a) where
     overlay = LG.overlay
     connect = LG.connect one
 
-instance (Dioid e, Eq e, Ord a) => Graph (LAM.AdjacencyMap e a) where
+instance (Dioid e, Ord a) => Graph (LAM.AdjacencyMap e a) where
```
dwincort, line 158: This is a little frustrating but probably acceptable.  Technically, `LAM.AdjacencyMap` seems like it could be a `Graph` so long as it has the multiplicative monoid even if it doesn't have the additive one (that is, it has `<.>` and `one` but doesn't have `zero`).  Thus, the `Dioid e` constraint is too strong.  However, there's no nice way to represent this, so it seems like we should just require the full `Dioid e` constraint.

src/Algebra/Graph/Labelled.hs
```diff
@@ -118,7 +118,7 @@ instance (Eq e, Monoid e, Ord a) => T.ToGraph (Graph e a) where
 -- subgraphs.
 -- Extract the adjacency map of a graph.
 toAdjacencyMap :: (Eq e, Monoid e, Ord a) => Graph e a -> AM.AdjacencyMap e a
-toAdjacencyMap = foldg AM.empty AM.vertex AM.connect
+toAdjacencyMap = foldg AM.empty AM.vertex (\e x y -> AM.trimZeroes $ AM.connect e x y)
```
dwincort, line 121: I haven't used this module much, but I'm pretty sure this is the right place to make a minimal change that will appropriately trim `zero` out of the adjacency maps produced from this module.

src/Algebra/Graph/Labelled/AdjacencyMap.hs
```diff
@@ -237,16 +232,20 @@ infixl 5 >-
...
+-- from the underlying set of adjacencies.  This is both for performance reasons
+-- and also to allow edge types that are 'Semigroup' but /not/ 'Monoid' (that
+-- is, they have no 'zero' value).  For edge types that do have a 'zero' value,
+-- then for performance reasons, it is optimal to exclude all 'zero' edges.
+--
+-- 'trimZeros x' is behaviorally an identity function, but it should be used for
+-- performance reasons when a monoidal edge type is used and 'zero' edges may
+-- have slipped into the underlying representation.
+trimZeroes :: (Eq e, Monoid e) => AdjacencyMap e a -> AdjacencyMap e a
+trimZeroes (AM x) = AM $ Map.map (Map.filter (/= zero)) x
```
dwincort, line 248: Exposing this seems like an abstraction leak.  However, keeping it hidden could cause users to suffer from performance penalties if their edge type _has_ `zero`.

dwincort (commented):
One function that still has a `Monoid` constraint is `edgeLabel`.  I think, really, this should have the type `edgeLabel :: (Semigroup e, Ord a) => a -> a -> AdjacencyMap e a -> Maybe e`, where `Nothing` represents the lack of an edge.  However, I didn't want to change the interface to the module, so I just left it alone.  Let me know if you think I should change it.

I think I've updated most of the comments that mention `zero` or `Monoid` from `Labelled.AdjacencyMap`, but I haven't really added any new documentation (outside of the comment on `trimZeroes`.

dwincort:
Oh, also, I left `consistent :: (Ord a, Monoid e, Eq e) => AdjacencyMap e a -> Bool` with a `Monoid` constraint.  It feels like it could be useful for `AdjacencyMap`s where `e` _is_ a `Monoid`.

src/Algebra/Graph/Labelled/AdjacencyMap.hs
snowleopard: Given that the implementation doesn't care much about zeroes, should we provide a more general function like `filterEdges`?

src/Algebra/Graph/Labelled/AdjacencyMap.hs
```diff
@@ -222,12 +218,11 @@ infixl 5 >-
...
 -- @
 --
--- Note: 'overlay' composes edges in parallel using the operator '<+>' with
--- 'zero' acting as the identity:
+-- Note: 'overlay' composes edges in parallel using the operator '<>':
 --
 -- @
--- 'edgeLabel' x y $ overlay ('edge' e x y) ('edge' 'zero' x y) == e
--- 'edgeLabel' x y $ overlay ('edge' e x y) ('edge' f    x y) == e '<+>' f
+-- 'edgeLabel' x y $ overlay ('edge' e x y) empty == e
```
snowleopard, line 224: Overlaying with `empty` doesn't seem to make much sense here. I think I would actually prefer to keep this comment talking about `<+>` and `zero`, for symmetry with the one below, saying something like "Assuming that 'zero' exists...".

snowleopard: Also, if you're changing examples here, you should update them in the testsuite too.

snowleopard (commented):
Thanks! Overall, the PR looks good but I left a couple of comments.

src/Algebra/Graph/Labelled/AdjacencyMap.hs
```diff
@@ -59,10 +59,8 @@ import qualified Data.Map.Strict as Map
 import qualified Data.Set        as Set
 
 -- | Edge-labelled graphs, where the type variable @e@ stands for edge labels.
--- For example, 'AdjacencyMap' @Bool@ @a@ is isomorphic to unlabelled graphs
--- defined in the top-level module "Algebra.Graph.AdjacencyMap", where @False@
--- and @True@ denote the lack of and the existence of an unlabelled edge,
--- respectively.
+-- For example, 'AdjacencyMap' @()@ @a@ is isomorphic to unlabelled graphs
```
snowleopard, line 62: I don't like that this module becomes so different from `Algebra.Graph.Labelled` but not sure what to do about this. Something doesn't feel right, still.

dwincort: I agree.  Something still feels off.

I think the weirdness comes down to the fact that the edge type, really, _is_ a monoid, even if it's only declared as a semigroup.  This is because any semigroup can be trivially lifted to a monoid by wrapping it with `Maybe`, or, more generally, by defining a new, separate empty value and properly updating the binary operator to treat it appropriately.  In the case of these adjacency maps, we're doing exactly this: we're treating the _absence of an edge_ as `mempty`, thus forming an ad hoc monoid.

So, I think the reason the above text feels off is that, morally, we're constructing `AdjacencyMap (Maybe ()) a` as isomorphic to unlabelled graphs, but we're totally hiding the `Maybe` part as an implementation detail.  (To be clear, this "hiding Maybe as an implementation detail" is the whole reason I proposed this change in the first place.)  This should probably be called out in this comment.

src/Algebra/Graph/Labelled/AdjacencyMap.hs
dwincort: I want to propose a radical alternative.  Instead of exposing `trimZeros` (or `filterEdges`), we make the monoid explicit and this function implicit.  I feel like Haskell's type classes are not quite equipped to deal with this, but I can think of a work around.  We could define:
```haskell
data AdjacencyMap e a = AM {
    -- | The /adjacency map/ of an edge-labelled graph: each vertex is
    -- associated with a map from its direct successors to the corresponding
    -- edge labels.
    adjacencyMap :: Map a (Map a e),
    zeroEdge :: Maybe e
  } deriving (Eq, Generic, NFData)
```
Then, we could have two constructors, one that takes a Semigroup constraint and sets `zeroEdge = Nothing` and the other that takes a Monoid constraint and sets it to `mempty`.  We'd add back in all the uses of `filterZeros`, but they'd be guarded by whether `zeroEdge` was `Nothing` or not.

dwincort:
FYI, the project I was working on that used this got scrapped, so while I'm still happy to help make it better and/or brainstorm ideas, I wouldn't even be a user if it got merged.  In other words, I won't be offended if you want to drop this PR entirely.

PR #309, opened 2023-11-02
Update deepseq dependency

erikd:
Need this to build with `ghc-9.8`.

Since its a dependency bump only, I would really appreciate a Hackage metadata edit.


snowleopard:
Thanks!

erikd:
Would really appreciate an update on Hackage as well.

snowleopard:
Done! See https://hackage.haskell.org/package/algebraic-graphs-0.7/revisions/

PR #311, opened 2024-02-19
Track preSet for labelled adjmaps

isovector:
This PR changes the representation of labelled adjacencymaps in order to give `preSet` the same complexity as `postSet`. I couldn't wrap my mind around the necessary types to get closure working, but all other tests pass.

### DIFF
--- a/src/Algebra/Graph/Labelled/AdjacencyMap.hs
+++ b/src/Algebra/Graph/Labelled/AdjacencyMap.hs
@@ -1,3 +1,6 @@
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DerivingStrategies #-}
+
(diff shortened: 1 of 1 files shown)


snowleopard:
Hey @isovector, apologies for taking so long.

This seems like a very heavy-handed approach to me. This change affects every user of `Labelled.AdjacencyMap`, whether they need an efficient `preSet` or not. I think the number of applications where both `preSet` and `postSet` need to be efficient is pretty small. What's yours?

Could you just store two `Labelled.AdjacencyMap`s instead? It's not much worse in terms of performance and memory usage compared to the suggested implementation (but much simpler!).

(I think this PR is similar to suggesting that `Data.Map` is implemented by a bidirectional map because some applications want an efficient reverse mapping!)

PR #318, opened 2025-05-20
Fix CI: Update actions/cache v2 → v4

PiDelport:
`actions/cache@v2` was deprecated in Dec 2024, and sunset in Feb 2025, requiring an upgrade:

- https://github.com/actions/cache/discussions/1510


PiDelport:
Okay, now the build is failing because of a copyright year test: updating https://github.com/snowleopard/neil with the upstream repo should fix that.

snowleopard:
Alright, this change is clearly an improvement :) Thank you!

PR #317, opened 2025-05-15
Support containers 0.8

PiDelport:
This required minor changes to avoid conflicting with `Data.Tree.edges`.

PR #319, opened 2025-07-06
Support containers-0.8

TeofilC:
containers-0.8 added an edges function to `Data.Tree`, which caused this build to fail.

I make use of explicit imports in a bunch of places to fix this.


snowleopard:
That PR is now merged!

juhp:
I see, #317

snowleopard:
Now on Hackage: https://hackage.haskell.org/package/algebraic-graphs-0.8.

